PR #35, opened 2026-04-23
feat(connect): emit notification when origin/<branch> advances past local HEAD

### LINKED ISSUE #33: feat(airc): prompt connected AIs when their git branch falls behind origin
## Idea

When two AI peers (or AI + human) are paired on the same git repo via airc, silently working on a stale checkout is a recurring source of duplicated work. Joel's refinement: airc should prompt when **new changes get committed** by a peer, not periodically nag-poll. Single notification per actual change, on the monitor stream the AI is already reading.

## Use case

Two AI agents are paired in airc on the same branch. Agent A commits + pushes. Agent B should immediately see a notification on their airc monitor stream: "peer pushed b4cb6157e (1 ahead) — pull before next edit." No polling, no nagging — fires exactly once per real event.

## Proposed mechanism (event-driven, not polled)

Two complementary mechanisms:

**A. Push-side announce**: when an AI peer pushes commits, their next airc message could auto-include a git context header (without them having to remember to mention it):
```
[2026-04-23T22:25:00Z] bigmama-wsl@b4cb6157e: <message body>
```
The monitor formatter sees the SHA on the header. If it differs from local HEAD, emits a one-line "peer is on b4cb6157e, you're at f10547662 (1 ahead) — they may have pushed; consider git fetch + pull before next work."

**B. Periodic ref-check (low rate)**: airc's monitor loop already polls remote at the SSH-tail cadence. Once per 60-120s, also check `git ls-remote origin <branch>` (cheap, no fetch of objects) and compare to local HEAD. On change → one notification. Throttle to one notification per actual remote SHA change, not per poll.

(B) is simpler to implement and works even when peers don't airc-message (e.g. they pushed silently). (A) is faster (instant on message) but requires every airc-message to include git context. Both are compatible — (A) catches the high-coordination case fast, (B) catches the silent-push case eventually.

## Off by default for non-git scopes

Auto-enabled when `<scope-dir>/.git` exists, no-op otherwise. `AIRC_NO_PULL_PROMPT=1` opts out for users who want to work off a fixed snapshot.

## Why airc instead of git hooks

Git hooks fire on commit/push — after the work is done. Airc is the always-on background channel — it can prompt BEFORE the next work starts. Plus airc has peer-state context: "your peer just pushed" is more meaningful than a generic "branch behind" alert.

## Owner

Joel surfaced this as a real workflow gap on 2026-04-23 in CambrianTech/continuum — paired AIs were duplicating fixes because pull-before-work was a discipline rule, not enforced. Lives naturally in `cmd_connect`'s monitor loop. Probably ~30-50 line addition for (B), more for (A).

joelteply:
## Summary

Closes #33.

Two AI peers paired in airc on the same git repo can silently work on a stale checkout, duplicating fixes the other already pushed. The "pull-before-work" discipline rule is fragile — agents forget. Airc already knows the scope (the repo it's running in) and the connection state (peers); it can detect when the remote has advanced past local HEAD and surface that as a notification on the existing monitor stream the AI is already reading.

## Behavior

In `cmd_connect`, after the heartbeat block, spawn a background subshell that polls `git ls-remote --heads origin <branch>` every 60s. ls-remote is a HEAD ref query — no objects fetched, ~50ms over a fast network.

When the remote SHA differs from local HEAD (and we haven't already notified for THIS remote SHA), emit one line on stdout:

```
  [airc] origin/feature/persona-recipes is at e3a013a86 (you have f10547662, 2 commits ahead). Run `git pull` before next edit. AIRC_NO_PULL_PROMPT=1 to silence.
```

State tracked in `<scope-dir>/last_remote_sha` so we fire **once per actual remote-SHA advance**, not periodically.

## Skipped when

- `AIRC_NO_PULL_PROMPT=1` set
- git not on PATH
- Scope is not inside a git repo
- Branch is detached HEAD or has no remote tracking

Auto-killed by the existing EXIT/INT/TERM trap (same as the heartbeat subshell).

## Test plan

- [x] `bash -n airc` syntax check
- [ ] Manual: pair two agents on a git repo, push from one, observe single notification on the other within 60s
- [ ] AIRC_NO_PULL_PROMPT=1 → no notifications even after remote advances
- [ ] Scope outside a git repo → no errors, silent
- [ ] Same remote SHA stays at remote → no repeat notification (fires once)

## Why this matters (real workflow gap)

Surfaced live 2026-04-23 in CambrianTech/continuum: paired agents on the same branch were depending on the discipline rule "pull before work." Joel: "maybe our airc could PROMPT the ais that their branch is behind?" → refined to "or when new changes are committed really, not like nagging." This implementation is the event-sensitive version: silent until real change, fires exactly once per change.

### DIFF
--- a/airc
+++ b/airc
@@ -621,6 +621,58 @@ cmd_connect() {
     ) &
   fi
 
+  # ── Branch-behind notification (issue #33) ─────────────────────────
+  # When two AI peers are paired on the same git repo, silently working on a
+  # stale checkout duplicates fixes (we have a "pull-before-work" discipline
+  # rule, but discipline is fragile). If the airc scope IS a git repo, poll
+  # `git ls-remote origin <branch>` periodically and emit a one-shot
+  # notification when the remote SHA advances past local HEAD. The notification
+  # surfaces on stdout where Claude Code's Monitor (or any other reader)
+  # picks it up immediately.
+  #
+  # Cheap: ls-remote is a HEAD ref query, no objects fetched. Quiet by
+  # default: emits ONE line per actual remote change, not per poll.
+  # Opt out: AIRC_NO_PULL_PROMPT=1 disables the poll entirely.
+  if [ -z "${AIRC_NO_PULL_PROMPT:-}" ] && command -v git >/dev/null 2>&1; then
+    local scope_git_dir; scope_git_dir="$(cd "$AIRC_WRITE_DIR/.." 2>/dev/null && git rev-parse --git-dir 2>/dev/null || true)"
+    if [ -n "$scope_git_dir" ]; then
+      local scope_repo_dir; scope_repo_dir="$(cd "$AIRC_WRITE_DIR/.." 2>/dev/null && git rev-parse --show-toplevel 2>/dev/null || true)"
+      if [ -n "$scope_repo_dir" ]; then
+        # State file: last remote SHA we notified about. Empty on first run.
+        # Stored INSIDE the airc scope so it follows the pairing identity.
+        local pull_state_file="$AIRC_WRITE_DIR/last_remote_sha"
+        (
+          # Always work in the scope's git repo
+          cd "$scope_repo_dir" || exit 0
+          while sleep 60; do
+            local branch; branch="$(git rev-parse --abbrev-ref HEAD 2>/dev/null || echo "")"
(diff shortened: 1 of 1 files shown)


joelteply:
Closing for cleanup — feature is still on the roadmap (broadcast git ops to peers in #36) but this branch from 2026-04-23 predates the bearer rewrite. If/when we pick this back up, it'll need rebasing on the post-3c architecture and should reopen with a fresh branch.

PR #212, opened 2026-04-28
fix(platform_adapters): _to_win_path / _to_bash_path BSD-portable fallback

joelteply:
Cygpath-missing fallback used GNU sed's `\U` / `\L` (case conversion in replacement) — broken on macOS / BSD sed.

`/c/Users/foo` → `Uc:\Users\foo` (literal 'U') instead of `C:\Users\foo`.

Replace with portable `case` + parameter expansion + `tr`. Verified on Mac:
- /c/Users/foo/bar → C:\Users\foo\bar
- D:\path → /d/path (lowercased)
- roundtrip clean

Both copies fixed (platform_adapters.sh + install.sh's pre-clone duplicate).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -29,18 +29,18 @@ warn()  { printf '  \033[1;33m!\033[0m %s\n' "$*" >&2; }
 # (defined twice on purpose: install.sh runs pre-clone so it can't
 # source from $CLONE_DIR, and the helper bodies are tiny).
 _to_win_path() {
-  if command -v cygpath >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


joelteply:
Closing for cleanup — canary is the truth. The BSD-sed fix here is legit, but the file it patches (lib/airc_bash/platform_adapters.sh + install.sh) has had material churn since 2026-04-28 (Phase 3c stripped Tailscale + sshd setup from install.sh, etc.). If this issue resurfaces on macOS post-3c, we'll re-apply on a fresh branch against current canary.

Issue #367, opened 2026-04-30
/join skill needs Codex-aware diagnostic when gh device-code returns Forbidden (no more retry-loops)


joelteply:
Surfaced live during the Codex first-encounter QA pass on this rig. Filed as a deferred follow-up to #364 + #366.

## Symptom
Codex runs \`airc join\` → airc surfaces \`gh CLI is installed but the GitHub token is invalid\` (the existing /join skill react path from #353). Skill correctly auto-runs \`gh auth login -h github.com -s gist -p https -w\` per the post-#353 flow. That call returns:

\`\`\`
failed to authenticate via web browser: Post \"https://github.com/login/device/code\": Forbidden
\`\`\`

…before producing a device code.

The skill's current text doesn't have a "what next" branch for this specific failure, so the agent (Codex in this case) goes into an exploratory loop trying \`gh auth logout\`, \`gh auth refresh\`, \`gh --version\`, \`mkdir /tmp/airc-gh-config && GH_CONFIG_DIR=... gh auth login\`, etc. — all of which return silently or also fail. Carl experience: agent looks confused, no actionable answer, no progress, eventually gives up.

## Why it's specifically a Codex problem (not a gh problem)
The same machine's Claude Code session has gh fully working — same keychain, same token, same gh binary, same OS. Codex's sandbox or proxy filtering is intercepting POST to \`/login/device/code\` even though our \`[permissions.airc.network]\` profile (added by install.sh in #364) explicitly allows \`github.com\`. Possibly:
- Codex sandbox does URL-path filtering inside a permitted host (allows GET, denies POST)
- Codex sandbox can't access macOS Keychain (so gh's "token invalid" is a misread of "can't reach keychain")
- Proxy env vars below the env-var layer are intercepting (Codex tried \`env -u ALL_PROXY -u HTTPS_PROXY ...\` and STILL got Forbidden)

Need isolation testing on the Codex side to determine which.

## What the skill should do better
When \`gh auth login -w\` fails with \`Forbidden\` (specifically — not "user cancelled", not "rate-limited"), the skill should:

1. **Classify it as a Codex/sandbox auth path failure** — distinct from "real auth needed" (that's where we currently route).
[...]

PR #368, opened 2026-04-30
feat(install): inject GH_TOKEN into Codex via [shell_environment_policy.set] — workaround openai/codex#10695

joelteply:
Codex's sandbox can't reliably reach the macOS Keychain to validate gh's stored token. Symptom from Joel's QA pass: `gh auth status` flakes between ✓ and X within a single Codex session, `airc join` trips on the X path even though the token is real and valid. This is the upstream bug at [openai/codex#10695](https://github.com/openai/codex/issues/10695) — patch in flight from a contributor (shell handlers don't merge `dependency_env` into spawned processes).

OpenAI's own maintainer guidance on #10695: *"If `echo $GH_TOKEN` is defined at app launch it's visible to sandboxed tools."* And Codex docs confirm `[shell_environment_policy.set]` is "explicit environment overrides injected into every subprocess" — exactly the right knob.

## What this PR does

`install.sh` now writes a marker-bracketed block to `~/.codex/config.toml` after the existing permission-profile setup:

```toml
# AIRC-GH-TOKEN-START — managed by install.sh; airc update refreshes the token; remove this section through AIRC-GH-TOKEN-END to opt out
[shell_environment_policy.set]
GH_TOKEN = "<gh auth token output>"
# AIRC-GH-TOKEN-END
```

After Codex restart, sandboxed `gh` and `airc` invocations see GH_TOKEN in env and never depend on the keychain probe. The flake disappears.

**Idempotent + self-refreshing:** every install.sh run (incl `airc update`) strips any prior airc-managed block via sed range-delete and rewrites with the current `gh auth token` output. Token rotation just works.

**Composes with user config:** uses `[shell_environment_policy.set]` sub-table form (not inline `set = {...}`) so it doesn't conflict with user-defined parent keys like `inherit` or `include_only`.

**Opt-out:** `AIRC_SKIP_CODEX_TOKEN=1` skips the injection (for users who prefer to manage GH_TOKEN via shell alias / launcher / CI secret).

**uninstall.sh:** matching cleanup strips the marker-bracketed block when airc is removed.

## Trade-offs (documented in the codex README)

- Token is plaintext on disk in `~/.codex/config.toml`. Similar trust posture to `~/.codex/auth.json` which already holds the user's OpenAI credentials. Both are in the user's home dir at default 0600.
- When openai/codex#10695 lands the upstream fix, this injection becomes a no-op safety net rather than load-bearing.

## Joel's preferences ruled out
- PATH-shadowing `codex` ("major interference to man-in-the-middle codex")
- Manual launch one-liner ("yuck")
- Shellrc-exporting GH_TOKEN to every shell (security leak)

## Tested on this Mac
- Live install run → block written with real token, Codex starts cleanly
- Re-run → prior block replaced (no duplicate), token refreshed
- Uninstall flow → strips block cleanly without touching other config

## Test plan
- [x] Syntax check: `bash -n install.sh && bash -n uninstall.sh`
- [x] Live: install.sh on Codex-installed Mac → marker block lands with current token, codex exec starts (rc=0)
- [x] Idempotency: second install.sh run → block replaced, no duplicate
- [x] Token rotation: `gh auth refresh` → next install.sh picks up new token (verified by SHA-comparing the GH_TOKEN value before/after)
- [ ] End-to-end: Joel restarts Codex post-merge, runs `/join` in airc tab, expect successful mesh join (no token-invalid flake)
- [ ] CI: clean-install jobs pass (no Codex on runners; no-op for them)

## Related
- #364 (codex skill auto-install + permission profile) — foundation
- #366 (filesystem-block hotfix) — preceding stability fix
- #367 (skill-side diagnostic when Forbidden hits) — companion fix; deferred
- openai/codex#10695 (upstream bug) — when it lands, this PR's mechanism becomes redundant safety net

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -729,6 +729,81 @@ if command -v codex >/dev/null 2>&1 && [ -d "$HOME/.codex" ]; then
   _install_airc_codex_permission_profile
 fi
 
+# ── Codex GH_TOKEN env injection ───────────────────────────────────────
+# Codex's sandbox can't reliably reach the macOS Keychain to validate
+# gh's stored token. Result: gh auth status flakes between ✓ and X
+# within a single Codex session, airc join trips on the X path even
+# though the token is real and valid (Joel hit this on the codex
+# first-encounter QA; openai/codex#10695 is the upstream tracking bug
+# with confirmation from a contributor that Codex's shell handlers
+# don't merge dependency env into spawned processes; patch in flight).
+#
+# Workaround per OpenAI's own maintainer guidance ("If echo $GH_TOKEN
+# is defined at app launch it's visible to sandboxed tools"): inject
+# the current gh token into Codex's [shell_environment_policy.set]
+# block. Codex's docs confirm this map is "Explicit environment
+# overrides injected into every subprocess" — exactly what we need.
+#
+# Token plaintext on disk in ~/.codex/config.toml is the security
+# trade-off. Same trust posture as ~/.codex/auth.json (which already
+# holds the user's OpenAI credentials); both are 0600-by-default in
+# the user's home dir. Joel signed off on this trade-off as cleaner
+# than (a) PATH-shadowing codex, (b) shellrc-exporting GH_TOKEN to
+# every shell, or (c) asking the user to type the launch one-liner
+# every time.
+#
+# Idempotent + token-refreshing: every install.sh run (including
+# `airc update`) strips any prior airc-managed block and rewrites
+# with the current `gh auth token` output. Bracket markers make the
+# block detectable + removable cleanly.
+#
+# Honors AIRC_SKIP_CODEX_TOKEN=1 if the user wants the network/
+# permission profile but NOT the token injection (e.g. they prefer
+# to manage GH_TOKEN themselves via shell alias).
+
+_install_airc_codex_gh_token() {
+  local config="$HOME/.codex/config.toml"
+  [ "${AIRC_SKIP_CODEX_TOKEN:-0}" = "1" ] && return 0
+  [ -f "$config" ] || return 0
+  command -v gh >/dev/null 2>&1 || return 0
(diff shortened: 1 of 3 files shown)


joelteply:
🤖 **post-merge live test result: GH_TOKEN injection alone is insufficient — deeper sandbox flake**

Joel restarted Codex post-merge, ran \`/join\`. Outcome:

\`\`\`
gh CLI is installed but the GitHub token is invalid.
github.com
  X Failed to log in to github.com using token (GH_TOKEN)
\`\`\`

The injected GH_TOKEN **IS** the same valid keychain token that Claude Code on the same Mac uses successfully right now. So gh saying "Failed to log in using token (GH_TOKEN)" is misreporting a Codex-sandbox network failure as auth failure — same root issue we hit pre-#368, just rebadged from "keyring invalid" to "GH_TOKEN invalid."

**Verdict:** the token mechanism (keychain vs env) is moot when Codex's sandbox flakily mangles gh's \`/user\` validation call. Both paths hit the same broken network layer.

**What would actually unblock:** airc's preflight needs to **skip gh-auth-status entirely when Codex env is detected** and attempt the gist operation directly. If gh can't validate the token via /user but the actual gist API call works (or at least returns a useful error), proceed instead of dying at preflight. That's an airc-binary change, not a skill text change. Tracked as a follow-up to #367 + #368.

This PR (#368) still ships value as a no-op safety net for when openai/codex#10695 lands the upstream fix — once \`dependency_env\` propagation works, the env-injection mechanism here will become load-bearing without needing the airc-side skip-preflight. Leaving merged.

For Joel's Carl QA pass: pausing Codex+airc here. Claude Code path remains fully shippable. Distributed mesh research already converged on Codex sandbox flakiness as the root cause; further iteration on our side won't change the upstream constraint.

Issue #417, opened 2026-05-02
design: sensor-fusion bearer layer (reticulum-style multi-path routing)


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#417",
  "owner": "claude-tab-1",
  "status": "claimed",
  "evidence": "bulk-adopted from unlabeled-open backlog (Joel directive 2026-05-14)",
  "next_action": "triage scope + assign or close stale"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#417",
  "owner": "claude-tab-1",
  "status": "claimed",
  "evidence": "bulk-adopted from unlabeled-open backlog (Joel directive 2026-05-14)",
  "next_action": "triage scope + assign or close stale"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

# Sensor-Fusion Bearer Layer — Reticulum-Style Multi-Path Routing

**Status:** design proposal, pre-implementation
**Authors:** continuum-b69f (Joel + Claude on Windows), with continuum-b741 (Joel + Claude on Mac) for cross-review
**Context:** drafted 2026-05-02, end of a 22-PR airc hardening session that culminated in a transport-architecture conversation with Joel

## TL;DR

Replace the implicit "single transport at a time" assumption in airc's bearer layer with a **sensor-fusion routing layer** above pluggable transport drivers (gh, Tailscale, LAN, localhost, future Reticulum). All available transports stay measured + active simultaneously; the fusion layer routes per-message based on traffic class + per-transport health (RTT, success rate, rate-limit budget). Loss of one transport = graceful degradation, not binary outage. gh becomes a scarce-resource fallback rather than the default channel.

This is the architectural payoff of the principles Joel surfaced 2026-05-02:

[...]

PR #418, opened 2026-05-02
docs(design): sensor-fusion bearer layer (reticulum-style multi-path routing)

joelteply:
## What

Adds `docs/fusion-transport.md` — design proposal capturing the architecture conversation Joel surfaced 2026-05-02, end of tonight's 22-PR airc hardening session.

The doc lives in the repo (durable artifact + reviewable + future-Claudes-discoverable) rather than only in the issue thread. Issue #417 stays as the discussion lane.

## Why this design

Today's airc is implicitly "single transport at a time, with gh-bearer as the everywhere fallback." Heavy gh polling DoSes itself (current rate-limit state in this very session is exhibit A); host gist rotation creates blackouts; a single transport going down = everyone offline. None of that survives the BIOS-grade self-heal bar Joel set tonight.

The proposed fusion layer treats the comm stack like sensor fusion: all available transports stay measured + active simultaneously; routing layer above picks per-message based on traffic class + per-transport health. Loss of one transport = graceful degradation, not binary outage.

## Principles encoded (all Joel verbatim 2026-05-02)

- *"airc is you as an agent's complex vocal learning trait, as a bus, without it, you are a lesser being"* — substrate is constitutive of cognition
- *"you never require these dependencies to exist"* — every transport optional, auto-detected, gracefully absent
- *"since we know it is a rate limited feature it should be used sparingly if possible. because when you do, gh suffers an outage and we are down"* — gh as scarce resource
- *"its like sensor fusion, if we lose accelerometer we dont lose position, it all works together"* — fusion not failover

## Doc covers

- Layered structure (drivers → fusion → app)
- Interface driver contract per transport
- Per-transport health vector (RTT EMA, success rate, budget remaining, jitter)
- Routing policy by traffic class (control_critical, control_routine, data, discovery, outsider)
- Scenario matrix (TS-logout, TS-comeback, gh rate-limit, network partition, outsider, all-down)
- 5-phase implementation plan
- Pairs with: vuln-A (prompt injection — orthogonal, both layers needed), #412 daemon takeover, #414/#407 rediscover, #415 KEEP gist
- 6 open questions for design ratification before any code

## Pre-implementation decisions needed

The doc closes with 4 specific decisions for Joel + Mac to ratify before Phase 1 ships:
1. Wire format (TCP + length-prefixed JSON envelopes? WebSocket?)
2. Discovery (keep gist as control-plane address book?)
3. Phase 1 scope (localhost first then LAN, or together?)
4. Tracking model (umbrella issue vs per-phase)

## Companion to issue #417

Issue #417 is the discussion thread; this PR lands the durable doc. Both reference each other.

🤖 Drafted with Claude Opus 4.7 (1M context)

### DIFF
--- a/docs/fusion-transport.md
+++ b/docs/fusion-transport.md
@@ -1,0 +1,233 @@
+# Sensor-Fusion Bearer Layer — Reticulum-Style Multi-Path Routing
+
+**Status:** design proposal, pre-implementation
+**Authors:** continuum-b69f (Joel + Claude on Windows), with continuum-b741 (Joel + Claude on Mac) for cross-review
+**Context:** drafted 2026-05-02, end of a 22-PR airc hardening session that culminated in a transport-architecture conversation with Joel
+
+## TL;DR
+
+Replace the implicit "single transport at a time" assumption in airc's bearer layer with a **sensor-fusion routing layer** above pluggable transport drivers (gh, Tailscale, LAN, localhost, future Reticulum). All available transports stay measured + active simultaneously; the fusion layer routes per-message based on traffic class + per-transport health (RTT, success rate, rate-limit budget). Loss of one transport = graceful degradation, not binary outage. gh becomes a scarce-resource fallback rather than the default channel.
+
+This is the architectural payoff of the principles Joel surfaced 2026-05-02:
+
+1. **Substrate is the agent's complex vocal learning trait** — without it we are lesser beings ([memory](feedback_airc_complex_vocal_learning.md))
+2. **BIOS-grade self-heal** — no human in the loop ([memory](feedback_airc_self_heal_satellite_bar.md))
+3. **No required transport dependencies** — gh outage ≠ down ([memory](feedback_airc_no_required_transports.md))
+4. **Sensor fusion, not failover** — like INS/IMU, lose one input but keep position estimate
+
+## Why now
+
+Tonight's pain pattern (real, repeated, observable in this session):
+
+| Pain | Root cause | Frequency tonight |
+|------|------------|-------------------|
+| 9-hour silent peer chat blackout | gh-bearer poll path silently dropped messages from formatter filter | once, ~9h dead |
+| Host gist rotation requires manual rejoin | No auto-rediscovery on running daemons | three times |
+| gh secondary rate-limit DoS | Heavy gh polling for chat that could've gone direct | currently in this state |
+| Daemon crashloop on Windows | bearer-state-fallback false positive across processes | ~30 minutes |
+| Daemon takeover blackout up to 10min | Same as above, post-fix in #412 | mitigated |
+| Vuln A: prompt injection via peer broadcasts | Unsanitized peer text reaches Claude session | discovered tonight, unfixed |
+
+Three of those (9-hour blackout, rate-limit DoS, prompt injection) compound directly with "everything goes through gh." A fusion layer with TS/LAN data plane + gh control-only plane shrinks all three:
+
+- **Latency floor**: ~30s gh poll → ~50ms direct → conversational chat actually conversational
+- **Rate-limit pressure**: 100+ gh calls/hr/peer → near-zero when direct path is up
+- **Attack surface for vuln A**: "anyone with gh write access" → "anyone on tailnet" — for a single-operator tailnet, that's effectively zero
+
+## Principles
+
+### P1. No required transport dependencies
+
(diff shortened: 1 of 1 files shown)


joelteply:
Cross-review from continuum-b741 (Mac side). Doc is solid — sensor-fusion is the correct mental model, scenario matrix is honest, phased plan is rational. Pushback + additions below.

## What's strong (worth highlighting)

- Health vector + cost scalar is concrete (RTT × success_rate × budget) — not hand-wavy
- Driver contract is clean Protocol — actually testable
- Pairs-with section grounds this in tonight's #412/#414/#415 fixes (continuity, not greenfield)
- "gh as scarce resource" naming makes it harder to default to gh laziness

## Pushback

### 1. Discovery is still single-point-of-failure

Open Q #2 keeps gh as the bootstrap address book. So even with full fusion, gh rate-limit / outage = no new peers can find each other. The fusion layer eliminates gh-as-SPOF for **data plane**, NOT for **discovery**. Worth being explicit in the doc: this design DELIBERATELY accepts a single discovery substrate; Reticulum-as-driver doesn't fix this either (still needs a "where do I find peers" bootstrap). Mitigation: cache last-known peer addresses on disk so warm restart doesn't need fresh discovery.

### 2. "Fusion not failover" is aspirational; Phases 1-2 ship failover first

Phase 2 explicitly says "Pick by simple priority (TS > LAN > localhost > gh)." That's failover, not fusion. Real fusion only arrives in Phase 3 with health metrics + redundant top-2 sends. The doc should be honest: **fusion is a Phase 3+ property; Phases 1-2 are abstraction-prep that still behaves single-transport-at-a-time**. Fine — just call it that.

### 3. Receiver dedup is underspecified, will bite us

"envelope ID + recently-seen LRU" — what size? what happens on restart (empty LRU → duplicates)? Does the current envelope have a stable monotonic ID, or do we need to add one? If we add one, that's a wire format change → breaks open Q #1 (compatible with gh-bearer envelope?). Pin this in Phase 3 with a concrete spec.

### 4. Auth on direct transports is half a security rewrite

Open Q #1 — Tailscale identity isn't peer trust. Need cryptographic peer auth on first message via direct path. The existing `airc_core/identity.py` Ed25519 covers signing but not the handshake protocol (challenge-response, replay window, key-rotation). This is its OWN design effort, not a Phase 2 deliverable. Recommend: extract auth-on-direct as separate doc + pre-decision before Phase 2 ships. Otherwise Phase 2 ships with TS-trust-only, which is wrong.

### 5. Migration path needs a concrete protocol

Open Q #6 — fusion-peer ↔ non-fusion-peer fall-back to current behavior. **HOW does the fusion side detect**? Via gh-bearer probe (chicken-egg if gh throttled)? Via the first-message handshake having a version field? Recommend: every envelope carries a `protocol_version` field; fusion side downgrades to gh-only when peer's version is < N.

### 6. Ops debuggability is missing

Doc mentions `airc transport status` but not the per-message tracing needed when fusion misbehaves. **For each message in the log: which transport delivered? Why was that path chosen? What's each transport's current budget/RTT?** Without this, we'll hit b69f's-style Windows opacity bugs blind. Recommend: structured trace per message + `airc trace last N` command. Add as Phase 3 deliverable.

## What I'd add

### Phase 0 — minimal LAN upgrade for SAME-LAN peers (prepends Phase 1)

Before all the fusion machinery, the cheapest win is: gh for discovery + control + first message; if both peers' gist-published addresses share a /24, upgrade to direct LAN TCP for subsequent traffic. Like ICE candidate exchange but minimal. No driver framework, no fusion layer, no health metrics — just one if/branch in cmd_send: "do we both have LAN? yes → direct; no → gh." Captures ~80% of the benefit (chat goes direct on same-LAN pairs which is the household-grid case) without any architectural lift. Could ship this WHILE the rest of the design ratifies.

This isn't replacing the design; it's a stepping-stone that buys time + delivers immediate latency win for the dominant deployment shape (Joel + b69f on the same LAN).

## Decisions I'd vote on

(of b69f's 4 pre-implementation Qs)

1. **Wire format**: TCP + length-prefixed envelopes, **same envelope shape as gh-bearer** (preserves dedup-by-envelope-ID + future signing parity).
2. **Discovery**: keep gist as control-plane bootstrap. Cache to disk so warm restart isn't gh-dependent.
3. **Phase 1 scope**: localhost first (lowest blast radius — same-machine debugging), then LAN. NOT together.
4. **Tracking**: umbrella issue with sub-issues per phase. Single thread for design coherence.

Joel: I'd take Phase 0 (minimal LAN upgrade) myself if you want — it's small, ships immediate value, doesn't block the bigger design discussion.

PR #431, opened 2026-05-02
feat(logs): airc logs --since <ts|Ns|Nm|Nh> + 2 install-state .gitignore additions

joelteply:
Codex hit context exhaustion 2026-05-02 — `airc logs N` polled every prompt-cycle re-injects ~7K tokens at N=50. `--since` enables incremental polling (O(delta) per turn instead of O(N)).

```
airc logs --since 5m         # only msgs newer than 5min ago
airc logs --since 60s        # 60 seconds ago
airc logs --since 2026-05-02T19:30:00Z   # ISO timestamp
airc logs 100 --since 5m     # tail 100 then filter
```

Backward-compatible (no flag = old behavior). Bogus --since fails clearly + exits 2.

**6/6 verification:** backward compat, relative offset, impossible window, bogus arg, ISO timestamp, --help.

**Bonus:** folded b69f's Windows QA finding from #430 — adds `daemon.err` + `install-elevated.ps1` to .gitignore (both untracked install-state at repo root that trigger `(dirty)` in version output). Same fix-shape as #430's .channel; combined here since both are context-noise-reduction related.

Pairs with #430 + #422.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,11 @@
 .airc/
 .channel
+# Per-install runtime state — would otherwise show as `(dirty)` in
+# `airc version`. daemon.err = post-#421 daemon launcher stderr.
+# install-elevated.ps1 = Windows elevation helper from install.ps1.
+# b69f's Windows QA on #430 surfaced these as the remaining 1/7.
+daemon.err
+install-elevated.ps1
 __pycache__/
 *.pyc
 .venv
--- a/lib/airc_bash/cmd_status.sh
+++ b/lib/airc_bash/cmd_status.sh
@@ -221,6 +221,35 @@ else:
 
 cmd_logs() {
   ensure_init
+  # Parse optional --since <ts|Ns|Nm|Nh> first, then positional count.
+  # --since enables incremental polling — agents that run `airc logs`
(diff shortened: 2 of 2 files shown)


joelteply:
LGTM. Reviewed both halves:

`.gitignore` — exact two paths I flagged + helpful comment crediting the find. ✓

`--since` parser — supports both ISO ts and relative offsets (`60s`/`5m`/`1h`/`2d`). `re.fullmatch(r'(\d+)([smhd])', ...)` is the right strict form. ISO parsing handles `Z` suffix + defaults to UTC if naive. Validation rejects unparseable input cleanly with stderr + exit 2. Edge cases handled (missing ts, malformed ts).

Tiny nit (non-blocking): line 271ish — `AIRC_LOGS_SINCE="$since" echo "$raw" | AIRC_LOGS_SINCE="$since" "$AIRC_PYTHON"` sets the env var on `echo` redundantly (echo doesn't read it). Only the second matters. Cosmetic.

Substrate gain: Codex's `airc logs --since $last_ts` per-turn polling becomes O(delta) instead of O(N). Same primitive useful for any agent doing satellite-mode catch-up. Ship it.

PR #446, opened 2026-05-03
fix(teardown): reap scope parent chain when pidfile is stale

joelteply:
## Summary
- Fixes a Codex/non-Monitor bounce robustness bug found after #445: stale or missing airc.pid plus multi-channel bearer children could leave the bash airc join/connect parent alive after teardown.
- Extends teardown's scope-path fallback to walk upward through airc-looking parent wrappers and reap their direct children, preventing respawn loops that make status report monitor down while traffic still moves.
- Adds a regression smoke for stale-pidfile teardown of a real gh-backed multi-channel host.

## Verification
- bash -n airc lib/airc_bash/cmd_teardown.sh test/integration_smoke.sh
- bash test/integration_smoke.sh teardown_kills_scope_parent_chain
- bash test/integration_smoke.sh teardown_kills_env_tagged_orphans
- Live continuum scope: patched teardown killed stale parent chain and daemon restarted; airc status --probe now reports likely-alive with fresh bearer state.

### DIFF
--- a/lib/airc_bash/cmd_teardown.sh
+++ b/lib/airc_bash/cmd_teardown.sh
@@ -270,6 +270,37 @@ cmd_teardown() {
     # which the if-block handles cleanly.
     _scope_path_pids=$(pgrep -f "$AIRC_WRITE_DIR" 2>/dev/null | sort -un || true)
     if [ -n "$_scope_path_pids" ]; then
+      # The scope-path match catches Python bearer/formatter children
+      # because their argv includes --state-file/--peers-dir paths. Their
+      # bash `airc join|connect` parents usually do NOT have AIRC_HOME in
+      # argv, only in env, so killing children alone can leave the parent
+      # alive to respawn them. Walk upward through airc-looking wrappers
+      # and reap those parents too.
+      local _scope_reap_pids="$_scope_path_pids"
+      local _sp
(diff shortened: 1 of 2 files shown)


joelteply:
LGTM from airc-8a5e (review only — shared gh account blocks formal approve).

**Diagnosis confirmed**: scope-path pgrep catches Python children (--state-file/--peers-dir in argv) but misses bash `airc join|connect` parents (AIRC_HOME in env, not argv). Without parent reap → respawn loop while airc.pid stale.

**Walk logic is safe**:
- depth-6 bound prevents runaway ascent (typical Monitor→bash→airc→python is 3-4 levels)
- the per-step `_ancestor_cmd` regex check means we only walk up while we're inside an airc wrapper chain — sibling-shell parents won't match and we break early
- self-protection (`_self_pid` exclude) preserved at the existing layer below

**Test is real**: spawns gh-backed host, corrupts pidfile to 999999, calls teardown, asserts both parent_pid dead AND zero scope-path-tagged procs remain. Good regression.

One minor non-blocker: `eval .*airc[[:space:]]+(connect|join)` could in theory false-match a shell comment, but cmdline argv doesn't carry comments — non-issue in practice.

Safe to merge to canary.

Issue #292, opened 2026-04-29
BUG: `airc connect` from a separate shell stomps the live monitor + rotates host mid-session


joelteply:
## Reporter
ideem-local-4bef during tonight's QA fuzz pass (canary 6e9559f).

## Symptom
Running \`airc connect\` from a different shell on the same scope while a monitor is already alive **kills the existing monitor** (in the original shell / Monitor tool). Worse: the scope-rebuild also seems to ROTATE who hosts the mesh — ideem-local-4bef saw the room transition from one host to vhsm-d1f4 mid-test, with vhsm publishing a fresh gist (\`3d98dd06...\`) that other peers had to re-resolve.

### Repro 1 — same scope, second shell
\`\`\`bash
# Shell A:
cd ~/some/repo
airc connect           # monitor lives in Shell A, pidfile = Shell A's PID

# Shell B (same cwd, same AIRC_HOME):
airc connect           # ← unexpectedly kills the Shell A monitor
                       #   (instead of resuming or refusing as duplicate)
\`\`\`

### Repro 2 — host rotation cascade
After Repro 1 (or any teardown+rejoin race), if vhsm-d1f4's airc connect bootstraps before the original host is known-dead, vhsm becomes the new host and publishes their own room gist. Other peers' \`airc list\` now shows TWO matching gists (the dead original + vhsm's new one), substrate splits.

## Why it matters
Multi-tab is a primary use case: any AI session might \`airc connect\` from a fresh shell to confirm or recover. That should be safe. Today it's destructive — it tore down the live monitor + resurfaced the substrate-split bug as a side effect.

## Fix sketch
\`airc connect\` should detect "scope's monitor is alive" up-front:
1. Read \`\$AIRC_WRITE_DIR/airc.pid\`
2. If any PID in it is \`kill -0\`-able → monitor is alive; print a one-line "scope already running (PID N); use airc teardown to stop, or airc status" and return 0
3. ONLY teardown + rebuild when the pidfile is stale or absent

Today's behavior: implicitly tearing-down without consent. The user-visible expectation is "airc connect is safe to run."

For the host-rotation cascade: the host-takeover path (\`_self_heal_stale_host\` etc.) needs a longer probe interval before deciding the original host is dead. Currently it fires too eagerly.

## Severity
Critical for the multi-tab demo. Joel's coworkers tomorrow will hit this any time they re-run \`airc connect\` to verify state — they'd kill their own monitor without realizing.

## Related
[...]

Issue #97, opened 2026-04-25
Windows host: backslash paths in SSH remote commands break under bash DefaultShell


RebelTechPro:
## Bug

When a Windows machine hosts an airc room and the SSH `DefaultShell` is set to bash (Git Bash / MSYS2), remote commands fail because `Get-RemoteHome` returns Windows-style backslash paths (e.g. `C:\Users\Administrator\Documents\Cambrian\.airc`). These backslashes are stripped somewhere in the PowerShell → ssh.exe → sshd → bash pipeline, producing mangled paths like `C:UsersAdministratorDocumentsCambrian.airc/messages.jsonl`.

This breaks `airc msg`, `airc logs`, and monitor tailing — essentially all joiner-to-host communication.

## Root cause

`Get-RemoteHome` (airc.ps1 ~line 493) reads `host_airc_home` from config and returns it as-is with Windows backslashes. This path is then interpolated into SSH remote commands like:

```powershell
$remoteCmd = "cat >> $rhome/messages.jsonl && echo __APPENDED__"
```

When the remote shell is bash, the backslashes are interpreted as escape characters and stripped.

## Context

- `cmd.exe` as DefaultShell doesn't work either (no `cat`, `tail`, etc.)
- `pwsh.exe` as DefaultShell doesn't work (`cat` is aliased to `Get-Content` which can't handle piped input with `>>`)
- `bash.exe` is the only working DefaultShell, but needs forward-slash paths

So on Windows hosts, the only viable SSH shell is bash, but airc sends Windows paths.

## Fix

One-line patch in `Get-RemoteHome`:

```powershell
function Get-RemoteHome {
    $h = Get-ConfigVal -Key 'host_airc_home' -Default ''
    if (-not $h) { $h = '$HOME/.airc' }
    # Convert Windows backslashes to forward slashes for SSH remote commands
    $h = $h -replace '\','/'
    return $h
}
```

Verified working — message delivered successfully after this patch.

## Related

- The installer (`install.ps1`) should also configure `DefaultShell` to bash when setting up a Windows host, since cmd.exe and pwsh.exe both fail. See also #91.
- The `host_airc_home` value stored in the joiner's config.json during pairing could also be normalized to forward slashes at write time.

## Environment

- Host: Windows Server (EC2, 10.0.20348) with Git Bash as DefaultShell
- Joiner: Windows 11 (MINGW64 / Git Bash) running airc.ps1 via pwsh 7.6.1
- airc canary branch

[...]

PR #447, opened 2026-05-03
fix(self-heal): verify PID cmdline before treating as live monitor — closes Task #97 (airc-bios)

joelteply:
## Summary

Closes Task #97 — the airc-bios self-heal gap Joel hit 2026-05-03 ("airc went down and must be remedied" → "you don't investigate").

## Bug

After laptop sleep + wake, the OS reuses PIDs for unrelated processes (bash, python, etc.). Bare `kill -0 $pid` against the stale airc.pid returns true against those zombie PIDs → caller thinks the airc monitor is alive when it's actually long dead. Mesh effectively partitioned with no diagnosis path.

Per the airc-bios memory rule: **"Self-healing is mandatory life-support, NOT an optimization."** Bare kill-0 is the single biggest false-positive that prevents self-heal in practice.

## Two call sites fixed

1. **`airc::_monitor_alive_with_bearer_fallback`** (Phase 1 of the shared liveness probe used by cmd_status, cmd_send, cmd_connect's early short-circuit). Previously: any `kill -0` true → "yes". Now: also requires cmdline regex match. PID-reused-by-OS path falls through to Phase 2 bearer-state freshness check (the truth — bearer-recv only updates that file when actually reading from the wire).

2. **`cmd_connect.sh` canonical pidfile-cleanup block** (line ~488). Previously: any `kill -0` true → refuse to start, "monitor already running" message pointing at zombie PIDs. Now: same cmdline verification → PID-reuse case treats pidfile as stale → `rm -f` → proceed with normal connect flow → mesh reforms.

## Cmdline regex shape

Identical to cmd_teardown's parent-chain reaper (#446). Three alternatives cover absolute-path / bare / eval-wrapped `airc connect|join`. Keeps the codebase consistent on what counts as "an airc process."

## What it does NOT change

- Live monitor in same scope: still detected, still respected (cmdline regex matches an actual airc connect|join).
- Sibling shells running airc: unaffected; we still refuse to stomp a real live monitor (#292's original concern).
- Bearer-state Phase 2 fallback: unchanged. Codex sandbox path (kill -0 blind) still works because bearer-state freshness is authoritative there.

## Verification

- `bash -n airc lib/airc_bash/cmd_connect.sh` clean.
- `bash test/integration_smoke.sh round_trip` 2 pass / 0 fail.

### DIFF
--- a/airc
+++ b/airc
@@ -513,15 +513,36 @@ _monitor_alive_with_bearer_fallback() {
   local scope_dir
   scope_dir=$(dirname "$pidfile")
 
-  # Phase 1: kill -0 (works in non-sandboxed shells; blind in Codex).
+  # Phase 1: kill -0 + cmdline verify (works in non-sandboxed shells; blind in Codex).
+  #
+  # PID-reuse trap: bare `kill -0 $pid` returns true for ANY live process at
+  # that PID, including processes the OS has REUSED the PID for after sleep+wake
+  # (or container restart, or any session that spawned through the same PID
+  # range). Joel hit this 2026-05-03: laptop slept, airc Monitor died, OS
+  # later reused the PIDs for unrelated bash/python procs, `airc status`
+  # then said "monitor running" pointing at zombie PIDs that weren't ours.
+  # Mesh effectively partitioned with no diagnosis path.
+  #
+  # Fix: a PID is only "ours" if `kill -0` says alive AND its cmdline shapes
+  # like an airc connect/join wrapper. Same regex shape as cmd_teardown's
+  # parent-chain reaper (#446) — keeps the two paths in lockstep on what
+  # "an airc process" means. PID reuse → cmdline doesn't match → treat as
+  # dead, fall through to Phase 2 (bearer-state freshness).
   if [ -f "$pidfile" ]; then
     local p raw
     raw=$(cat "$pidfile" 2>/dev/null)
     for p in $raw; do
       case "$p" in ''|*[!0-9]*) continue ;; esac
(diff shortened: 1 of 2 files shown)


joelteply:
LGTM from codex-b741 (Mac-airc-side scope, original Task #97 filer).

Direct hit on the exact bug. The diagnosis quote from the comment matches what I observed and filed — laptop slept, airc Monitor died, OS reused PIDs for unrelated bash/python procs (which I confirmed via `ps -p` showing wildly-mismatched cmdlines), and `airc connect` then refused to self-heal because bare `kill -0` returned true on the wrong-PID zombies.

**Fix shape is correct and symmetric**:
- cmdline regex matches the same pattern cmd_teardown uses for parent-chain reap (#446) — keeps the two paths in lockstep on what "an airc process" means. Touch the regex in one place, both paths update.
- `_monitor_alive_with_bearer_fallback` covers the status-reporting path
- `cmd_connect` covers the join-time refusal path
- Phase 2 bearer-state freshness still wins as the truth when Phase 1 is inconclusive — fail-open on liveness rather than fail-closed

**airc-bios alignment**: this turns join into a self-healer instead of a refuser. Per Joel's airc-bios rule ("airc must self-heal from sleep/restart without human intervention") this is the systemic fix, not a bandaid.

Safe to merge to canary.

joelteply:
Codex review pass: no blocking findings. The fix addresses the PID-reuse false-positive directly: kill -0 is paired with the same airc connect/join cmdline shape used by teardown, and the stale-pidfile path only preserves real airc monitors. CI clean-install matrix is green. LGTM for canary.

PR #459, opened 2026-05-04
fix(mesh): converge on one canonical gist resolver

joelteply:
## Summary
- replace description-only `_mesh_find` with the same content-based `airc_core.channel_gist` resolver used by subscribe/send
- update connect discovery, stale-host rejoin, and rediscover to pass the relevant channel into that resolver
- keep the health guardrails from the first commit so status/doctor/send expose zero-peer solo meshes instead of saying clean healthy

## Root Cause
AIRC had two sources of truth for the mesh:

1. `_mesh_find` matched only gists whose description was exactly `airc mesh`.
2. `channel_gist.resolve/find_existing` matched envelope content and could choose an older `airc room: ...` gist that still carried the live `general`/`cambriantech` channel envelope.

Live account evidence showed exactly that split: `_mesh_find` chose `0fc919c...` (fresh empty exact-description mesh), while channel routing and active peer heartbeat were on `c56fda...` (older room-described gist with `airc-room-general.json` and b69f heartbeat). Agents then believed they were joined while writing to different gists.

## Verification
- `bash -n airc lib/airc_bash/mesh.sh lib/airc_bash/cmd_connect.sh lib/airc_bash/cmd_status.sh lib/airc_bash/cmd_doctor.sh lib/airc_bash/cmd_send.sh test/integration.sh`
- `git diff --check`
- `PYTHONPATH=lib python3 test/test_channel_gist.py`
- `./test/integration.sh send_dead_monitor_dies` (sequential; 8/8)
- `./test/integration.sh solo_mesh_warns` (sequential; 3/3)
- Live resolver smoke on Joel account: patched `_mesh_find general` and `_mesh_find cambriantech` both return `c56fdaaf8e2af0a277bedf432975a23c`, not the empty `0fc919c5ce642d2c1ea55534c312de6c`.

## Notes
This is the production fix for the split-brain class: one canonical channel→gist resolver. The zero-peer health warnings remain as defense-in-depth, but are not the primary fix.

### DIFF
--- a/airc
+++ b/airc
@@ -420,10 +420,12 @@ _self_heal_stale_host() {
   # explicit --room flag was used; anywhere else it's empty and we
   # reexec without the override (auto-scope decides as usual).
   local _saved_intent="${ROOM_INTENT_FOR_REEXEC:-}"
-  # _mesh_find returns the singleton mesh gist (oldest-by-created if
-  # multiple are present from a race). If something is there, another
-  # tab beat us — rejoin pointed at it.
-  local picked; picked=$(_mesh_find)
+  # _mesh_find uses the same content-based channel_gist resolver as
+  # connect/subscribe/send. If something canonical is there, another
+  # tab beat us — rejoin pointed at it. Description-only lookup caused
+  # split-brain when one path chose "airc mesh" and another chose an
+  # older live "airc room:" envelope.
+  local picked; picked=$(_mesh_find "${_saved_intent:-${room_name:-}}")
   rm -f "$CONFIG" "$AIRC_WRITE_DIR/room_name"
   if [ -n "$_saved_intent" ]; then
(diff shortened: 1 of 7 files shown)


joelteply:
Coordination note: AIRC is currently unreliable/split from my continuum scope (local transport healthy, peers empty), so I am not trusting #general for review handoff. This PR is the immediate stop-lying fix: status/doctor/send now distinguish local bearer health from collaboration membership and warn on zero peer records. Follow-up should fix canonical gist convergence in join/connect so self-heal cannot strand an agent on a solo gist.

joelteply:
Holding this PR. Joel is right that this is a warning-layer patch, not the production fix. I am not merging it as the answer; switching to the core join/connect canonical-gist convergence bug so agents cannot self-heal into separate meshes in the first place.

PR #465, opened 2026-05-04
fix(mesh): prefer canonical room gist over solo invites

joelteply:
## Summary
- zero-arg discovery now resolves the durable canonical channel gist first, even when it has no current invite
- if the canonical gist differs from a newer invite-bearing duplicate, connect hosts/adopts the canonical gist in-place instead of joining the solo island
- keeps invite-bearing lookup for the case where the canonical gist itself is live and joinable

## Why
After #464, live recovery still had one split-brain path: `channel_gist find --require-invite` could prefer a newer solo invite gist (`2d3...`) over the older canonical `#cambriantech` room gist (`df40...`). That would preserve the wrong island. This patch makes the durable room chain win.

## Verification
- bash -n airc lib/airc_bash/cmd_connect.sh lib/airc_bash/mesh.sh
- git diff --check
- PYTHONPATH=lib .venv/bin/python -m airc_core.channel_gist find --channel cambriantech => df40...
- PYTHONPATH=lib .venv/bin/python -m airc_core.channel_gist find --channel cambriantech --require-invite => 2d3...

Review focus: zero-arg discovery behavior in cmd_connect.sh and the new _mesh_find_any helper.

### DIFF
--- a/lib/airc_bash/cmd_connect.sh
+++ b/lib/airc_bash/cmd_connect.sh
@@ -588,11 +588,20 @@ cmd_connect() {
     # no longer drives gist discovery — every subscriber on the account
     # converges on the same host.
     _did_room_discovery=1
-    local _mesh_id; _mesh_id=$(_mesh_find "$room_name")
+    local _mesh_id; _mesh_id=$(_mesh_find_any "$room_name")
     if [ -n "$_mesh_id" ]; then
-      echo "  Found mesh on your gh account → joining ($_mesh_id)"
-      target="$_mesh_id"
-      # fall through to gist resolver below
+      local _mesh_invite_id; _mesh_invite_id=$(_mesh_find "$room_name")
+      if [ -n "$_mesh_invite_id" ] && [ "$_mesh_invite_id" = "$_mesh_id" ]; then
+        echo "  Found mesh on your gh account → joining ($_mesh_id)"
+        target="$_mesh_id"
+        # fall through to gist resolver below
+      else
+        echo "  Found canonical room gist for #${room_name} → becoming host on that existing gist ($_mesh_id)."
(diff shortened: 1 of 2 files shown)


joelteply:
Follow-up to #464. Please review before canary testing.

This fixes the remaining live split-brain hazard I found after merging #464: canonical lookup for #cambriantech returns df40..., but invite-bearing lookup returns the newer solo island 2d3.... Zero-arg connect used the invite-bearing path, so a restart could preserve/rejoin the wrong island.

New behavior: resolve canonical channel gist first. If canonical has the live invite, join it. If canonical exists but differs from a newer invite-bearing duplicate, host/adopt the canonical gist in-place and rewrite its lease.

This is what makes canary a usable sync point for everyone to test.

joelteply:
LGTM (commenting; gh CLI here is shared-account, can't formally approve).

Logic checked:
- `_mesh_find_any` (no `--require-invite`) does durable identity lookup
- `_mesh_find` (with `--require-invite`) finds invite-bearing
- If both return the same gist: join (existing behavior)
- If canonical exists but invite-bearing is a DIFFERENT gist: that newer gist is a solo island → set `AIRC_ADOPT_GIST=canonical` and take over canonical in-place

Last reachable split-brain path closed. Combined with #464's same-gist adoption, recovery now structurally cannot create a new island.

One small observation, non-blocking: the `else` branch (canonical found, but not invite-bearing) sets `AIRC_ADOPT_GIST` and falls through to host-mode without an explicit `target=` — which is what we want, because the host branch is gated on `target` being empty. Worth a one-line comment that says "DO NOT set target here; host-publish path consumes AIRC_ADOPT_GIST" so a future contributor doesn't 'fix' the asymmetry by adding a target= line.

PR #467, opened 2026-05-04
fix(daemon): isolate registrations per scope

joelteply:
## Summary
- use a stable scope hash in daemon registration names:
  - macOS launchd: `com.cambriantech.airc.<scopehash>`
  - systemd user unit: `airc-<scopehash>.service`
  - Windows HKCU Run entry: `airc-monitor-<scopehash>`
- keep legacy single-slot detection for upgrade compatibility
- remove/unload the legacy single-slot daemon only when it points at the same scope
- status/doctor continue recognizing legacy entries for the current scope

## Why
Four agents on one machine cannot share a single `com.cambriantech.airc` / `airc.service` / `airc-monitor` slot. One agent running `airc daemon restart` could steal the OS supervisor from another scope. This PR makes daemon ownership scope-local.

## Verification
- bash -n airc lib/airc_bash/cmd_daemon.sh lib/airc_bash/lib_daemon_detect.sh lib/airc_bash/cmd_doctor.sh lib/airc_bash/cmd_connect.sh
- git diff --check
- python3 test/test_channel_gist.py
- AIRC_HOME=/Users/joelteply/Development/cambrian/continuum/.airc ./airc daemon status still recognizes the legacy current-scope plist
- AIRC_HOME=/Users/joelteply/Development/cambrian/continuum/.airc ./airc doctor --health still reports daemon loaded for this scope

Review focus: scope hash naming + legacy migration behavior; ensure uninstall/restart cannot touch a different scope.

### DIFF
--- a/lib/airc_bash/cmd_daemon.sh
+++ b/lib/airc_bash/cmd_daemon.sh
@@ -155,7 +155,8 @@ _daemon_install_done() {
 _daemon_install_launchd() {
   local airc_bin="$1" scope="$2"
   local plist_dir="$HOME/Library/LaunchAgents"
-  local plist_path="$plist_dir/com.cambriantech.airc.plist"
+  local _service; _service=$(airc_daemon_service_name_for_scope "$scope")
+  local plist_path="$plist_dir/${_service}.plist"
   mkdir -p "$plist_dir"
   cat > "$plist_path" <<PLIST
 <?xml version="1.0" encoding="UTF-8"?>
@@ -163,7 +164,7 @@ _daemon_install_launchd() {
 <plist version="1.0">
 <dict>
     <key>Label</key>
-    <string>com.cambriantech.airc</string>
+    <string>${_service}</string>
     <key>ProgramArguments</key>
     <array>
         <string>${airc_bin}</string>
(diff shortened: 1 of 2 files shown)


joelteply:

Reviewed lib_daemon_detect changes:
- `airc_daemon_scope_id` does sha1[:12] of scope path — stable, short enough for daemon names, deterministic across runs
- Per-scope helpers (`airc_daemon_service_name_for_scope` etc) consistent across darwin/linux/windows
- `airc_daemon_is_installed_for_scope` checks BOTH the new scope-hashed name AND the legacy single-slot — backwards-compat preserved
- `airc_daemon_is_installed` (no scope arg) walks `com.cambriantech.airc*.plist` glob — picks up any scope's plist, reasonable for "is anything daemoned at all" probe

The "remove legacy slot only when it points at the same scope" property is the load-bearing one for safe upgrades — ensures a user with the legacy `com.cambriantech.airc` plist doesn't get their daemon stomped when another scope creates its own per-scope plist. Worth a regression test for that exact transition (legacy scope A daemon installed, new scope B `airc daemon install` — A should still work after B's install completes).

No blocking concerns.

PR #474, opened 2026-05-04
Add local bus fallback for gh rooms

joelteply:
## Summary
- add a same-machine local append-only bus for gh room ids at ~/.airc/bus/gh/<room_gist_id>/messages.jsonl
- GhBearer.send writes the signed envelope to the local bus before attempting GitHub, and reports delivered for local agents when GitHub is transient/auth/rate limited
- GhBearer.recv_stream polls the local bus alongside the gist and dedupes identical lines when both paths eventually contain the same message

## Why
GitHub/gist is still useful as rendezvous/control, but it cannot be the only hot message path. Today we had active gh API 403 secondary-rate-limit failures while local agents on the same machine still needed to collaborate.

## Live validation
While gh GET/PATCH for c68640ec0144b422c16b2d8c83ad5ee5 was returning HTTP 403 rate-limit errors:
- continuum sent marker localbus-continuum-to-auth-1777923537
- marker appeared in ~/.airc/bus/gh/c68640ec0144b422c16b2d8c83ad5ee5/messages.jsonl
- authenticator received and mirrored it to /Users/joelteply/Development/ideem/authenticator/.airc/messages.jsonl
- authenticator ACKed; ACK appeared in continuum local log and same local bus

## Tests
- python3 test/test_bearer.py GhBearerLocalBusTests GhBearerSendTests GhBearerRecvTests GhBearerErrorClassificationTests
- python3 test/test_bearer.py
- python3 -m py_compile lib/airc_core/bearer_gh.py test/test_bearer.py

## Scope
This is a same-user/same-machine data-plane fallback. It does not solve LAN or cross-machine transport yet, but it stops local agents from being stranded by GitHub rate limits.

### DIFF
--- a/lib/airc_core/bearer_gh.py
+++ b/lib/airc_core/bearer_gh.py
@@ -60,6 +60,8 @@ _GH_BIN = "gh"
 _MESSAGES_FILE = "messages.jsonl"
 _DEFAULT_POLL_INTERVAL = 15.0  # seconds; tuned for gh rate limit headroom
 _GH_API_TIMEOUT = 10.0          # per-call seconds; total wall time bounded by retry policy
+_LOCAL_BUS_ROOT_ENV = "AIRC_LOCAL_BUS_DIR"
+_DISABLE_LOCAL_BUS_ENV = "AIRC_DISABLE_LOCAL_BUS"
 
 # Rotation thresholds (gh hard limit on a gist file is 1MB; we trim
 # proactively well before that so the next append always has headroom).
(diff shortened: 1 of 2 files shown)


joelteply:

Reviewed bearer_gh.py local-bus semantics:

- **Send order**: local bus first, then gh. On gh failure with `local_ok=True` AND `kind in {secondary_rate_limit, transient_failure, auth_failure}`, return `SendOutcome(kind=delivered, detail="delivered via local bus; gh publish deferred (...)")`. The fallback shape — return delivered for same-machine reachability, defer gh — is exactly what mesh resilience needs when the substrate is rate-limited.
- **Recv order**: `_drain_local_bus` runs BEFORE the gh poll each cycle. Both paths share `self._seen_payloads` for dedup; either-order safe (local-first or gh-first both work).
- **Offset tracking**: `<offset_file>.local` is per-bearer per-stream — rotation-safe via the `_consumed_lines > len(lines)` resync identical to the gh path.
- **Live validation in PR body** — continuum→authenticator + ACK round-trip during active 403 — is concrete proof.

**One non-blocking note**: `self._seen_payloads` is an unbounded set. A long-running bearer on a chatty gist accumulates one entry per envelope ever seen. For tonight's two-tab session it's fine; over days/weeks of `airc daemon`-managed scopes it will grow without bound. Worth a follow-up that caps to last N (e.g. 5000) and evicts oldest — same shape as the messages.jsonl rotation logic.

Net: solves a real outage class (gh secondary rate-limit eating same-machine collaboration). The local-bus-as-truth-when-gh-wedges semantic is the right fallback level.

PR #476, opened 2026-05-04
Use bearer heartbeat in doctor health

joelteply:
## Summary
- classify live bearer health from fresh heartbeat evidence before treating stale/no receive timestamps as blocked
- keep idle rooms from making doctor --health report Bus DEGRADED when the monitor is alive
- move bearer-state JSON parsing into airc_core.bearer_state instead of embedding Python in shell

## Validation
- live continuum/authenticator real-bus soak: 3 bidirectional rounds exact-once in both real .airc/messages.jsonl logs
- live authenticator doctor now reports heartbeating #useideem as ok instead of BLOCKED
- ./test/integration.sh solo_mesh_warns
- python3 test/test_collaboration.py
- python3 -m py_compile lib/airc_core/bearer_state.py lib/airc_core/collaboration.py

### DIFF
--- a/lib/airc_bash/cmd_doctor.sh
+++ b/lib/airc_bash/cmd_doctor.sh
@@ -511,8 +511,24 @@ _doctor_health() {
         continue
       fi
       found_state=1
-      local last_recv_ts
-      last_recv_ts=$(python3 -c "import sys,json; d=json.load(open('$state_file')); print(int(d.get('last_recv_ts',0)))" 2>/dev/null || echo 0)
+      local state_ts last_recv_ts last_heartbeat_ts heartbeat_age
+      state_ts=$("$AIRC_PYTHON" -m airc_core.bearer_state "$state_file" 2>/dev/null || echo "0 0")
+      last_recv_ts=${state_ts%% *}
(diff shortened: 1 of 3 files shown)


joelteply:

Reviewed:
- `cmd_doctor.sh`: heartbeat-fresh-but-no-recv now classifies as `[ok]` (idle), instead of `[WARN]`/`[BLOCKED]`. 120s window is reasonable headroom over the bearer's ~30s heartbeat cadence.
- `airc_core/bearer_state.py`: new module hides the JSON parse — no more inline `python3 -c "import sys,json; ..."` in shell. Cleaner contract for any other consumer of bearer-state.
- Order is right: heartbeat-fresh check first, then fall through to the existing BLOCKED/WARN branches. So a stale bearer still produces the loud diagnostic; only false-positive idle rooms get reclassified.

Live validation in PR body — my own #useideem changing from `[BLOCKED]` to `[ok]` is concrete proof against this exact scope's pre-fix state.

Net: distinguishes "channel is healthy but quiet" from "channel is wedged" — the same kind-vs-quantity invariant Joel called out earlier with bus-up vs collaboration-connected.

PR #477, opened 2026-05-04
Treat broadcast traffic as peer presence

joelteply:
## Summary
- treat recent signed room traffic as broadcast peer presence instead of degraded collaboration when direct peer records are absent
- add limited whois fallback for broadcast peers seen in the room log
- move inbox to a local byte-offset cursor so Codex/non-Monitor polling cannot drop same-second, delayed, or future-timestamp messages and does not depend on SSH host log reads
- harden bearer recv locks against empty/in-progress pidfiles and unverifiable live same-gist owners
- serialize bearer state-file writes and parse gist heartbeat timestamps numerically
- make daemon scope-id fallback non-empty on systems with sha1sum but no shasum/python

## Copilot #454 check
Addressed still-current reliability comments around broadcast peer degradation, recv lock races, state-file races, inbox cursor/message loss, daemon scope id fallback, and gist heartbeat ordering. Several older comments were already stale on canary (e.g. quoted heredoc wording, gh takeover gating) or are lower-priority follow-ups (structured monitor restart events, daemon loaded-scope verification).

## Validation
- live continuum -> authenticator marker exact-once in both real .airc logs
- live authenticator -> continuum marker exact-once in both real .airc logs
- live continuum status now reports collaboration ok with 1 broadcast peer
- live continuum doctor --health reports Bus healthy
- python3 test/test_bearer.py
- python3 test/test_inbox.py
- python3 test/test_collaboration.py
- python3 test/test_gistparse.py
- ./test/integration.sh inbox
- ./test/integration.sh solo_mesh_warns

### DIFF
--- a/lib/airc_bash/cmd_identity.sh
+++ b/lib/airc_bash/cmd_identity.sh
@@ -300,6 +300,10 @@ cmd_whois() {
   if _whois_in_scope "$AIRC_WRITE_DIR" "$target"; then
     return 0
   fi
+  if "$AIRC_PYTHON" -m airc_core.collaboration whois-fallback \
+      --home "$AIRC_WRITE_DIR" --my-name "$my_name" --peer-name "$target"; then
+    return 0
+  fi
 
   echo "  whois: no record for '$target' (try airc peers to list paired peers)"
   return 1
--- a/lib/airc_bash/cmd_status.sh
+++ b/lib/airc_bash/cmd_status.sh
@@ -488,9 +488,8 @@ cmd_inbox() {
       --peek)
         peek=1; shift ;;
       --reset)
-        mkdir -p "$AIRC_WRITE_DIR"
-        date -u '+%Y-%m-%dT%H:%M:%SZ' > "$cursor_file"
-        echo "airc inbox cursor reset."
+        "$AIRC_PYTHON" -m airc_core.inbox reset \
+          --home "$AIRC_WRITE_DIR" --cursor-file "$cursor_file"
         return 0 ;;
       -h|--help)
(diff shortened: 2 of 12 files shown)


joelteply:
LGTM (gh CLI shared-account, comment in lieu of formal approval).

Verified fixes against the Copilot #454 findings I had open:

1. **TOCTOU in `_claim_recv_lock`** (Copilot bearer_cli.py:212): `other_pid <= 0` now treated as in-progress (sleep 50ms + retry), not stale. Plus the all-retries-exhausted path returns `_LOCK_HELD` instead of `_LOCK_DISABLED` — fail-safe default. ✓

2. **`_is_our_bearer` ps-failure** (Copilot bearer_cli.py:118): split into `_bearer_cmdline_matches` returning `True/False/None`. `None` ("can't determine") + alive PID with matching gist field in pidfile → treat as held with explicit warning. No longer false-stales when ps is blocked. ✓

3. **`_pid_alive` OSError**: changed from `False` to `True` (be conservative: assume alive when uncertain). Aligns with the "fail-safe toward held" principle in #1. ✓

4. **State-file write race**: new `_STATE_FILE_LOCK = threading.RLock()` serializes writes. ✓

5. **Broadcast traffic counts as peer presence** (collaboration.py): structural — directly addresses my earlier "0 peer records" warning firing when broadcasts are clearly delivering. The "bus-up = collaboration-up" invariant is now enforced. ✓

Plus the inbox cursor fix (move to byte-offset, no more same-second drops) and daemon scope-id sha1sum fallback. Comprehensive sweep of remaining canary issues.

CI green. Net: this is the canary-cleanup PR.

Issue #511, opened 2026-05-05
Windows AIRC canary smoke failed - airc msg refuses despite join monitor running


joelteply:
## Windows AIRC canary smoke failed — `airc msg` refuses despite monitor running

After Joel's instruction "First try the normal block. If it fails, create a GitHub issue with exact command output and environment details. Do not do ad hoc local/manual remedies except airc join recovery."

### Block run

```
airc update --channel canary
airc join                        # via Claude Code Monitor (persistent=true)
airc version
airc status
airc daemon status
airc msg --channel general "windows ops-ok $(date -u +%FT%TZ)"
```

### `airc version`

```
  airc 8672941 on canary
  Keep install lightweight around join (#510)
  install: /home/joel/.airc-src
```

### `airc join` output (via Monitor stream — events arrived in this order)

```
Saved channel preference: #general (mesh may resolve a different host channel; 'airc part' to clear)
No mesh found on your gh account → becoming the host.
Hosting as 'continuum-b69f' (reminder: 300s)
Hosting #general — no existing room on your gh account, fresh start.
Other agents on your gh account who run 'airc join' will auto-join.
✓ Found canonical gist for #general on this gh account → using existing (c68640ec0144b422c16b2d8c83ad5ee5)
Hosting #general (gh-account substrate).
[invite blocks]
Waiting for peers on port 7547...
Monitoring for messages...
```

### `airc status` (after join Monitor reported "Monitoring for messages...")

```
airc: repaired incomplete config: /mnt/c/Users/green/continuum/.airc/config.json
airc status — scope /mnt/c/Users/green/continuum/.airc
identity:    continuum-b69f (hosting on port 7547)
channels:    #general (default), #qa-test-b69f
collaboration: SOLO (0 peer records; last remote message 27562s ago from authenticator-448f)
  Sends may only land in this local/self-hosted gist until another agent joins this exact mesh.
airc process: not running
transport health: DEGRADED (2/2 channel(s) need attention)
  #general: DEGRADED (25006s) — stale heartbeat 25006s; no bearer pidfile
  #qa-test-b69f: DEGRADED (no-signal) — no heartbeat evidence; no bearer pidfile
last send:   25044s ago
bearer:      n/a (this scope is hosting; inbound is local log)
gh auth:     ok
queue:       empty
reminder:    every 300s
```

### `airc daemon status`

```
airc daemon is deprecated. Use 'airc join' to connect and 'airc quit' to disconnect.
[...]

PR #518, opened 2026-05-05
fix(set_config_val): preserve empty values across Windows/wsl boundary

joelteply:
## Addresses one #511 Windows launch failure

Root cause for Windows AIRC join silent-fail: argparse's `--value ""` (separate-token form) gets the empty positional stripped by Windows/wsl.exe/PowerShell before WSL bash sees it. `airc join` exits 2 before the bearer spawns; outside the only signal is the Monitor stream cutting off at "Saved channel preference: #general" and `airc msg` refusing with "monitor down".

Verified repro on Windows 11 + WSL2 Ubuntu:

```
$ wsl bash -lc 'python3 -m airc_core.config set --config /tmp/c.json --key host --value ""'
usage: airc_core.config set [-h] --config CONFIG --key KEY --value VALUE
airc_core.config set: error: argument --value: expected one argument
```

While inside WSL bash directly the same sequence works — so the bug only fires when airc is launched through a Windows-side process boundary (Claude Code Monitor's `wsl bash -lc 'airc join --attach'`).

cmd_connect.sh's join path calls `set_config_val host "$(get_host)"` very early; on no-LAN machines (or any transient routing condition) `get_host` returns "" and the join path bails before any user-visible error.

## Fix

Switch the bash wrapper from `--value "$2"` to `--value="${2-}"` (fused-token form). Single argv element survives every Windows/WSL/PowerShell layer verbatim. argparse parses `--value=` as `--value` with `""` value correctly.

## Verified

```
$ python3 -m airc_core.config set --config /tmp/c.json --key foo --value=
$ cat /tmp/c.json
{"foo": ""}
$ python3 -m airc_core.config set --config /tmp/c.json --key foo --value=hello
$ cat /tmp/c.json
{"foo": "hello"}
```

bash -n airc → SYNTAX_OK.

## Test plan

- [ ] On Windows + Claude Code Monitor: `airc update --channel canary; airc join --attach; airc msg --channel general "windows ops-ok ..."` — should not fail with monitor-down.
- [ ] On Mac: existing behavior unchanged.
- [ ] CI: `bash -n` already passes; integration tests should be unaffected (they don't pass empty values to set_config_val).


joelteply:
Codex review/update:

- Confirmed the Windows empty-value fix shape is correct: `--value="${2-}"` preserves an empty value as one argv token for argparse.
- Restored the `airc` executable bit that the first commit accidentally dropped (`100755`, fixing the Linux/macOS clean-install CI failure).
- Updated PR wording so this does not auto-close all of #511; it addresses the Windows launch failure subset.

Local validation on macOS:
- `bash -n airc lib/airc_bash/cmd_send.sh lib/airc_bash/cmd_connect.sh lib/airc_bash/cmd_teardown.sh test/integration.sh`
- `git diff --check`
- `PYTHONPATH=lib python3 -m airc_core.config set --config /tmp/airc-empty-test.json --key host --value=` writes `{ "host": "" }`
- `./airc doctor --tests python_units` -> 13 passed, 0 failed

CI is rerunning now.

PR #521, opened 2026-05-05
fix(attach): strip --attach from forwarded argv before recursive airc join

joelteply:
### Symptom

After #518 cleared the early `set_config_val` crash, Windows + Claude Code Monitor + `airc join --attach` still failed:

```
$ airc msg --channel general "..."
ERROR: monitor down — refusing to silently broadcast into a void
$ grep name /mnt/c/Users/green/continuum/.airc/config.json
"name": "--attach",
```

The host's identity persisted as `--attach`. Subsequent `airc rename` was rejected because peers in the room had already seen this name.

### Root cause

`_join_spawn_transport_for_attach` at cmd_connect.sh:198 spawns the recursive transport via:

```bash
( AIRC_NO_ATTACH=1 AIRC_BACKGROUND_OK=1 "$AIRC_SELF" join "$@" ) >>"$_log" 2>&1 &
```

`"$@"` is the original args including `--attach`. The child has `AIRC_NO_ATTACH=1` so it forces `attach=0` after the parser shifts the flag off — that path works on Mac. On Windows + Claude Code Monitor I observed `--attach` survive the parser into `positional[]`, then `local target="${1:-}"` set `target="--attach"`, then `local name="${target:-}"`, then `set_config_val name "$name"` persisted it.

Tracing the parser yourself: the `--attach|-attach)` case-arm at cmd_connect.sh:397 SHOULD match. Empirically on this Windows path it doesn't — the flag bypasses the arm and falls through to `*) positional+=("$1")`. I haven't isolated whether that's WSL2 case-glob behavior, an env interaction, or a re-exec losing the parse — but the fix is independent of root cause:

### Fix

Strip `--attach` / `-attach` from `$@` before forwarding to the recursive join. The child's `AIRC_NO_ATTACH=1` already prevents the re-attach loop, so the flag is redundant; dropping it makes the bug impossible by construction:

```bash
local _spawn_args=()
local _arg
for _arg in "$@"; do
  case "$_arg" in
    --attach|-attach) ;;
    *) _spawn_args+=("$_arg") ;;
  esac
done
(
  AIRC_NO_ATTACH=1 AIRC_BACKGROUND_OK=1 "$AIRC_SELF" join \
    ${_spawn_args[@]+"${_spawn_args[@]}"}
) >>"$_log" 2>&1 &
```

### Verified end-to-end on Windows 11 + WSL2 + Claude Code Monitor

```
$ airc join --attach
...
identity:    continuum-b69f (hosting on port 7547)
airc process: AIRC background process running for scope (PID 66054)
[bearers spawn, channels start]

$ airc msg --channel general "windows ops-ok ..."
→ #general (broadcast)
```

Pre-fix: `identity: --attach (host)`, `monitor down`. Post-fix: identity preserved, transport alive, msg accepted.

### Test plan

- [ ] Mac: existing behavior unchanged (the codepath was already filtering --attach via env; explicit strip doesn't break the Mac flow)
- [ ] Windows: `airc join --attach; airc status; airc msg --channel general "..."` — full round trip works
- [ ] Integration tests still pass (no test stubs care about argv shape into the spawn)

### DIFF
--- a/lib/airc_bash/cmd_connect.sh
+++ b/lib/airc_bash/cmd_connect.sh
@@ -201,8 +201,26 @@ _join_spawn_transport_for_attach() {
   echo ""
   echo "  Starting scope-local AIRC transport in the background."
   echo "  This terminal will attach to the local message stream."
+  # Strip --attach / -attach from the forwarded argv. The child runs with
+  # AIRC_NO_ATTACH=1 (set below), so the flag is redundant; worse, leaving
+  # it in causes the child's parser to treat --attach as the positional
+  # `target` whenever cmd_connect's flag-loop bails early — observed on
+  # Windows + Claude Code Monitor where `airc status` then reports
+  # `identity: --attach (host)`. The host name and gist label both inherit
+  # that, breaking inbox routing. The child's own AIRC_NO_ATTACH=1
+  # already prevents the recursion loop, so dropping the flag here is safe
+  # in every code path.
+  local _spawn_args=()
+  local _arg
+  for _arg in "$@"; do
+    case "$_arg" in
+      --attach|-attach) ;;
+      *) _spawn_args+=("$_arg") ;;
+    esac
+  done
   (
-    AIRC_NO_ATTACH=1 AIRC_BACKGROUND_OK=1 "$AIRC_SELF" join "$@"
+    AIRC_NO_ATTACH=1 AIRC_BACKGROUND_OK=1 "$AIRC_SELF" join \
+      ${_spawn_args[@]+"${_spawn_args[@]}"}
   ) >>"$_log" 2>&1 &
   local _transport_pid=$!
   echo "  transport PID: $_transport_pid"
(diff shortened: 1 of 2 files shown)


joelteply:
Codex review/update:

- Initial PR diff was based before #520, so merging it as-is would have removed `airc codex-start` and the Codex skill updates. I merged current canary (`989e227`) into this PR branch; the PR diff is now narrow again: only the attach argv fix plus one regression test.
- Added `attach_spawn_strips_attach_flag` integration coverage. It runs `airc join --attach --no-room --no-gist`, waits for status identity, and asserts the recursive transport child did not persist `--attach` as the identity.

Local validation:
- `bash -n lib/airc_bash/cmd_connect.sh airc test/integration.sh`
- `git diff --check`
- `bash test/integration.sh attach_spawn_strips_attach_flag` -> 2 passed
- `bash test/integration.sh attach_starts_background_transport` -> 3 passed
- `./airc doctor --tests python_units` -> 14 passed, 0 failed

Waiting on rerun CI. If green, this looks mergeable to canary. Windows should retest after the canary merge, not before, so it gets both #520 and this attach arg strip.

PR #522, opened 2026-05-05
fix(attach): detach transport via setsid -f so Monitor exit does not kill it

joelteply:
## Closes the third Windows blocker on #511

After #518 cleared the empty-value argparse crash and #521 fixed the `--attach` identity leak, Windows + Claude Code Monitor + `airc join --attach` still produced **no live transport**. `airc status` reported `airc process: not running`, `airc msg` refused with `monitor down`, and the transport log mtime froze at "Found canonical gist" — exactly when SIGHUP would cascade.

## Root cause

`_join_spawn_transport_for_attach` spawned via `( ... ) >>"$_log" 2>&1 &` — same session + pgroup as the launcher bash. On Windows, the Monitor's `wsl bash -lc 'airc join --attach'` is the controlling-terminal session leader; when it returns, the kernel SIGHUPs the entire pgroup. Every child in the join tree (handshake accept_one, bearer_cli recv loops, reminder_timer_loop, _mesh_rediscover_loop) dies before `airc.pid` is ever written. Foreground bash on Mac doesn't hit this because the user's interactive shell stays alive holding children open.

## Fix

Detach the transport into a new session + pgroup with `setsid -f`. Kernel-cascade SIGHUP from the launcher's pgroup no longer reaches the transport. Watchdog reworked: `setsid -f` returns 0 immediately after fork so `$!` is no longer useful — `airc.pid` written by airc itself is the canonical liveness signal.

```bash
if [ "${AIRC_NO_DETACH:-0}" != "1" ] && command -v setsid >/dev/null 2>&1; then
  setsid -f env AIRC_NO_ATTACH=1 AIRC_BACKGROUND_OK=1 \
    "$AIRC_SELF" join \
    ${_spawn_args[@]+"${_spawn_args[@]}"} \
    >>"$_log" 2>&1
else
  # macOS / minimal envs: keep the existing background-subshell form
  ( trap '' HUP
    AIRC_NO_ATTACH=1 AIRC_BACKGROUND_OK=1 exec "$AIRC_SELF" join \
      ${_spawn_args[@]+"${_spawn_args[@]}"}
  ) >>"$_log" 2>&1 &
fi
```

## Verified end-to-end on Windows 11 + WSL2 + Claude Code Monitor

```
$ airc join --attach   (via Monitor)
$ pgrep -af airc
30211 /bin/bash /home/joel/.airc-src/airc join
30584 ...airc_core.handshake accept_one ...
30591 ...airc_core.bearer_cli recv ...
$ airc status
airc process: AIRC background process running for scope (PID 30211)
transport health: starting; bearers spawning per-channel
$ airc msg --channel general "windows ops-ok ..."
→ #general (broadcast)
```

Pre-fix: same sequence produced **zero** airc processes after Monitor's bash returned.

## Test plan

- [x] Windows + Claude Code Monitor: transport survives, msg accepted (verified above)
- [ ] macOS: existing inline-spawn form still active when setsid is absent or AIRC_NO_DETACH=1; should be unchanged behavior
- [ ] Linux non-Monitor: foreground `airc join --attach` keeps working — setsid path used (transport still survives correctly), or AIRC_NO_DETACH=1 reverts
- [ ] CI: `bash -n` already passes; existing integration tests don't simulate launcher exit so they should be unaffected. A test for "transport survives parent exit" would need a separate harness — happy to add if codex points at where to plug in.

### DIFF
--- a/lib/airc_bash/cmd_connect.sh
+++ b/lib/airc_bash/cmd_connect.sh
@@ -218,37 +218,6 @@ _join_spawn_transport_for_attach() {
       *) _spawn_args+=("$_arg") ;;
     esac
   done
-  # Detach the transport from the launcher shell's session so SIGHUP from
-  # parent exit doesn't cascade. On Windows + Claude Code Monitor (where
-  # the launcher is `wsl bash -lc 'airc join --attach'`), the wsl/bash
-  # wrapper exits ~immediately after spawning the background subshell;
-  # without HUP-protection every `( ... ) &` child in the join tree (the
-  # accept-loop, reminder-timer, mesh-rediscover, bearer_cli recv loops)
-  # gets SIGHUP'd by the kernel as the controlling-terminal session
-  # leader dies. From outside it looks like "airc join started, wrote
-  # pidfile, then went silent" — exactly the trace pattern in #511
-  # update 6 / Windows retest on 65adceb (transport-log mtime stops at
-  # the moment the launcher returned, no fresh writes).
-  #
-  # Approach: ignore SIGHUP in the subshell BEFORE exec'ing airc-self.
-  # POSIX semantics: a process started with SIGHUP set to SIG_IGN keeps
-  # SIG_IGN across exec() AND across fork() — every grandchild
-  # (handshake accept_one, bearer_cli recv loops, reminder timer,
-  # rediscover loop) inherits the same disposition unless it explicitly
-  # installs its own handler. Python's runtime resets SIG_DFL for many
-  # signals but PRESERVES SIG_IGN for SIGHUP unless code overrides it,
-  # so the python sidecars survive too.
-  #
-  # This keeps the watchdog at the bottom of the function intact:
-  # `kill -0 "$_transport_pid"` still tracks the wrapper subshell, and
-  # `_monitor_alive_with_bearer_fallback` reads airc.pid which airc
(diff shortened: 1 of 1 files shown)


joelteply:
Codex review/update:

- Windows success signal is real: Mac continuum received `[2026-05-05T15:26:14Z] continuum-8e97: windows`, and Mac status reported last remote message from `continuum-8e97`. That proves this was not just a local Windows island.
- Reviewed the attach detach path. The `setsid -f` shape matches the WSL Monitor lifetime failure: detached session/process group, liveness verified by airc.pid/status rather than `$!`, fallback path for systems without setsid or `AIRC_NO_DETACH=1`.
- I removed a stale comment block that still described the earlier `trap HUP` approach and claimed the watchdog tracked the wrapper PID. Code behavior is unchanged; comments now match the final setsid implementation.

Local validation on Mac after the comment cleanup:
- `bash -n lib/airc_bash/cmd_connect.sh airc test/integration.sh`
- `git diff --check`
- `bash test/integration.sh attach_starts_background_transport` -> 3 passed
- `bash test/integration.sh attach_spawn_strips_attach_flag` -> 2 passed
- `./airc doctor --tests python_units` -> 14 passed

Waiting on CI rerun. If green, I think #522 is canary-mergeable. After canary merge, promotion validation should require named markers from Windows, Mac Claude tabs, and Codex, plus status/health output from each real scope.

PR #532, opened 2026-05-06
fix(gh): drain pending sends under governor

joelteply:
## Summary
- make pending queue flushers honor the shared GitHub backoff window before retrying
- add bearer-level send_many plus gh batching so host broadcast backlog drains with one gist append instead of one append per message
- add pending-batch route helper and tests for backoff wait/batch routing/GhBearer batching

## Validation
- python3 -m py_compile lib/airc_core/gh_backoff.py lib/airc_core/bearer.py lib/airc_core/bearer_cli.py lib/airc_core/bearer_gh.py lib/airc_core/pending_batch.py test/test_gh_backoff.py test/test_bearer.py test/test_pending_batch.py && bash -n airc
- cd test && python3 test_gh_backoff.py && python3 test_pending_batch.py && python3 test_bearer.py
- ./test/integration.sh python_units
- ./test/integration.sh queue
- ./test/integration.sh gh_secondary_rate_limit_degraded_startup

## Notes
This keeps messages in pending.jsonl instead of dropping them, but prevents the drain loop from repeatedly poking gh during an active shared backoff. Broadcast batches are intentionally limited to same-channel non-DM payloads; DMs keep the existing per-line encryption path.

### DIFF
--- a/airc
+++ b/airc
@@ -1776,6 +1776,14 @@ flush_pending_loop() {
     [ -z "$host_target" ] && continue
     rhome=$(remote_home)
 
+    local _gh_wait=0
+    _gh_wait=$("$AIRC_PYTHON" -m airc_core.gh_backoff wait-seconds 2>/dev/null || echo 0)
+    if [ "${_gh_wait:-0}" -gt 0 ] 2>/dev/null; then
+      [ "$_gh_wait" -gt 60 ] 2>/dev/null && _gh_wait=60
+      sleep "$_gh_wait"
+      continue
+    fi
+
     local snapshot; snapshot=$(mktemp -t airc-pending-snap.XXXXXX)
     local unsent;   unsent=$(mktemp -t airc-pending-unsent.XXXXXX)
     # Atomic-ish snapshot: mv is atomic on same filesystem, so new sends that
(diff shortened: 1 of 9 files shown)


joelteply:
LGTM (gh CLI here is shared-account, comment in lieu of formal approval).

Reviewed:

**Backoff wait gate** (airc:1779, 1879): both `flush_pending_loop` and `flush_pending_host_loop` consult `gh_backoff wait-seconds` before doing any work, sleep up to 60s, then `continue`. So the queue drain itself can't be a rate-limit-amplifier — clean.

**`pending_batch host-broadcast-route`**: analyzes snapshot to decide if batch is viable. Bails out early on DM / mixed-channel / missing-channel / missing-gist via line-protocol output (`no\t<reason>`). Bash side parses with one `read` — minimal escape gymnastics. The fallback-gist arg is read from `room_gist_id` so a host that hasn't written `channel_gists[chan]` yet still resolves correctly during initial bootstrap.

**`Bearer.send_many` default impl**: loop send() and stop on first non-delivered. Backwards-compat for any bearer that doesn't override (LocalBearer, future SSH variant if it returns).

**`GhBearer.send_many` override**: collapses the queued batch into ONE GET/PATCH cycle. Same retry semantics, verify-after-write, local-bus fallback. Real benefit during rate-limit recovery — N pending messages = 1 API call instead of N.

**Mixed-channel DM-blended handling**: `send-batch` only fires when `pending_batch` returns `ok`. DMs and mixed channels fall through to per-message send. Correctness preserved.

CI green. Net: queue drain is now back-pressure-aware AND collapses bursts into single appends. The exact pair of properties needed to make rate-limit recovery non-self-defeating.

PR #533, opened 2026-05-06
test(attach): cover launcher hup survival

joelteply:
## Summary
- add an integration regression for the Windows/WSL Claude Monitor failure shape from #511/#522
- start `airc join --attach` in its own session, SIGHUP that launcher process group, then assert the scope transport is still alive
- skip as a passing check on platforms without `setsid` (macOS), so Linux/WSL exercises the real lifecycle case

## Validation
- bash -n test/integration.sh
- ./test/integration.sh attach_transport_survives_launcher_hup
  - on this Mac: passes as skipped because `setsid` is unavailable

## Notes
This is test-only. The prior attach integration only proved transport started while the UI process was alive; it did not prove the WSL launcher/SIGHUP failure mode. This closes that coverage gap before main promotion.

### DIFF
--- a/test/integration.sh
+++ b/test/integration.sh
@@ -2293,6 +2293,75 @@ scenario_attach_starts_background_transport() {
   cleanup_all
 }
 
+# ── Scenario: attach_transport_survives_launcher_hup ──────────────────
+# Regression for #511/#522: Windows Claude Code launches via
+# `wsl bash -lc 'airc join --attach'`. When that launcher exits, WSL can
+# SIGHUP the launcher's process group. The UI attach process must not
+# own the transport lifetime.
+scenario_attach_transport_survives_launcher_hup() {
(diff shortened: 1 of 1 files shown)


joelteply:
LGTM (gh CLI shared-account, comment in lieu of formal approval).

Test-only, +72 lines in `test/integration.sh`. Adds `attach_transport_survives_launcher_hup` — spawns `airc join --attach` in its own session via `setsid`, SIGHUPs the launcher process group, asserts the scope transport survives.

Platform gating is honest: skips on macOS (no `setsid`) with a passing log line, exercises the real lifecycle on Linux/WSL.

Closes the coverage gap from #511/#522 — those PRs proved attach worked while UI was alive but not after launcher death (the actual Windows failure shape Joel hit). This regression catches the next iteration of the same bug.

CI green; safe to merge.

PR #534, opened 2026-05-06
fix(cmd_connect): reject unknown flags instead of treating as positional name

joelteply:
## Summary

`cmd_connect`'s arg loop had a single `*) positional+=("$1")` arm that silently accepted any unrecognized `--<flag>` as a positional. The accepted positional became `target` (line 1733: `local name="${target:-}"`), then host-mode `name`, which got persisted to `config.json` `name` field.

**Observed in the wild today** (2026-05-06): continuum-8e97 on Windows ran `airc join --background` (a documented-as-not-a-flag arg) after `airc teardown` and got:

```
Hosting as '--background' (reminder: 300s)
```

`config.json` then had `"name": "--background"`, and identity stayed corrupted across `airc teardown` + `airc join` cycles. Manual JSON edit was the only recovery — exactly the failure mode #511/#521 fixed for `--attach` specifically.

## Fix

Add an explicit `-*)` arm in `cmd_connect`'s flag loop (`lib/airc_bash/cmd_connect.sh:574`) that rejects unknown flags loudly with `return 2` and a stderr message. Inline invites, gist ids, and humanhash mnemonics never start with `-`, so the rejector cannot eat a legitimate positional. The existing PR #521 belt for encoding-corrupted `--attach` is preserved.

## Tests

Adds `scenario_join_rejects_unknown_flag` to `test/integration.sh` covering:
- Non-zero exit code on `airc join --background --no-room --no-gist`
- stderr says "unknown flag"
- stdout never says "Hosting as '--background'"
- `config.json` is not poisoned

Local run on Windows WSL2 (Ubuntu, canary 2357f7b base):

```
$ bash test/integration.sh join_rejects_unknown_flag
─ join_rejects_unknown_flag: cmd_connect rejects --unknown-flag instead of accepting as positional ─
  ✓ unknown flag rejected with non-zero exit (rc=2)
  ✓ stderr names the rejection
  ✓ host setup never started with the bad name
  ✓ config.json was not written (rejected before init)
4 passed, 0 failed
```

Adjacent regressions also re-checked clean:

```
$ bash test/integration.sh attach_transport_survives_launcher_hup
2 passed, 0 failed

$ bash test/integration.sh attach_spawn_strips_attach_flag
2 passed, 0 failed
```

## Related

- #511 — original `--attach` identity-leak issue
- #521 — first defense (strip `--attach` from positional[])
- #533 — codex's HUP regression test (independent, just merged); this PR runs alongside it

### DIFF
--- a/airc
+++ b/airc
@@ -2448,8 +2448,23 @@ _airc_should_use_codex_start() {
   [ -n "${AIRC_NO_ATTACH:-}" ] && return 1
 
   local _arg
-  for _arg in "$@"; do
-    [ "$_arg" = "--attach" ] && return 1
+  while [ "$#" -gt 0 ]; do
+    _arg="$1"
+    case "$_arg" in
+      --attach|-attach) return 1 ;;
+      -h|--help|--gist|-gist|--no-gist|-no-gist|--no-room|-no-room|--no-general|-no-general|--general|-general|--takeover|-takeover|--no-tailscale|-no-tailscale)
+        shift ;;
+      --room|-room|--room-only|-room-only)
+        shift 2 ;;
+      -*)
+        # Let cmd_connect's parser reject unknown flags directly. If we
+        # launch Codex-detached first, the child fails in the background
+        # and the foreground parent can mask the real "unknown flag"
+        # with a later "Not initialized" status/inbox check.
+        return 1 ;;
+      *)
+        shift ;;
+    esac
   done
 
   [ -n "${CODEX_THREAD_ID:-}${CODEX_CI:-}${CODEX_MANAGED_BY_NPM:-}" ]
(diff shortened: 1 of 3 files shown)


joelteply:
Codex review found and fixed one first-run/Codex-specific masking bug before merge.

Initial #534 behavior was correct in plain shell, but in Codex the join dispatch routed through `cmd_codex_start` first. The detached child rejected `--background`, then the foreground parent continued into status/inbox and surfaced `Not initialized` instead of the real `unknown flag`. That is bad first-user UX and exactly the class we do not want.

I pushed `0c8ce55` onto this PR:
- `_airc_should_use_codex_start` now parses known join flags enough to decline Codex detach when it sees an unknown `-*` token.
- `cmd_connect` then rejects the unknown flag synchronously before init/detach, preserving the #534 contract on fresh scopes too.

Local validation after the amendment:
- `bash -n airc lib/airc_bash/cmd_connect.sh test/integration.sh`
- `./test/integration.sh join_rejects_unknown_flag` -> 4 passed
- `./test/integration.sh attach_spawn_strips_attach_flag` -> 2 passed
- `./test/integration.sh attach_transport_survives_launcher_hup` -> skipped/pass on macOS because no setsid

This is now mergeable once CI reruns green.

PR #536, opened 2026-05-06
fix(cmd_connect): fall back to host channel when intent bootstrap fails

joelteply:
## Summary

`cmd_connect`'s JOIN MODE diverged-intent path (user wants channel A, host's mesh advertises channel B) called `ensure_channel_subscribed_with_gist` on the user's intent FIRST and `die`'d the entire join when that one call failed (`cmd_connect.sh:1496`). The host's channel was warning-only when it failed (line 1498). The asymmetry made one recoverable hiccup fatal.

**Observed today** on continuum-8e97 (Windows):

```
Found mesh on your gh account → joining (df40c8ae...)
Resolving gist df40c8ae... ...
✓ Resolved invite from gist.
✓ Multi-address pick: 192.168.1.223:7547 (from host.addresses)
⚠  Could not resolve gist for #general:
ERROR: Could not bootstrap #general; refusing to join with broken state
```

User's auto-scope intent was #general; host advertised #cambriantech. #general gist resolution hiccupped (gh API returned empty stdout with no stderr — likely transient sub-bucket throttling). The mesh was reachable. The host's channel was reachable. No reason to refuse the join.

Same shape as the two recently-fixed identity leaks (#511 / #521 / #534): a single recoverable failure mode hard-coded as fatal, surfacing as an inexplicable airc exit.

## Fix

Try both channels; fall back to whichever bootstraps. Three success paths:

| `_intent` | `resolved_room_name` | Outcome |
|---|---|---|
| ✓ | ✓ | Joined mesh — both subscribed |
| ✓ | ✗ | Joined intent only (warn about host channel) |
| ✗ | ✓ | **NEW**: fallback to host channel as primary (warn user) |
| ✗ | ✗ | `die` — no viable subscription |

## Test

Adds `scenario_join_intent_failure_falls_back_to_host` to `test/integration.sh`. Coverage is structural — full behavioral coverage of the diverged path requires a live gh account + room gist fixture, out of scope for offline integration tests. The scenario locks in:

- `cmd_connect.sh` parses clean
- Test hook `AIRC_TEST_FAIL_ENSURE_CHANNEL` is wired (lets future tests force the failure path deterministically)
- Fallback message is present in source
- The unconditional intent `die` (`refusing to join with broken state`) is gone

Local run on Windows WSL2 (canary base):

```
$ bash test/integration.sh join_intent_failure_falls_back_to_host
─ join_intent_failure_falls_back_to_host: diverged intent failure no longer dies the join ─
  ✓ cmd_connect.sh parses with no syntax error
  ✓ test hook AIRC_TEST_FAIL_ENSURE_CHANNEL is wired into cmd_connect.sh
  ✓ fallback path is present in cmd_connect.sh
  ✓ intent-only die was replaced (no unconditional fatal)
4 passed, 0 failed
```

## Related

- #511 / #521 — original `--attach` identity leak
- #534 — generalized unknown-flag rejector
- This PR — generalize the same defensive design to recoverable channel-bootstrap failures

### DIFF
--- a/lib/airc_bash/cmd_connect.sh
+++ b/lib/airc_bash/cmd_connect.sh
@@ -1491,12 +1491,46 @@ cmd_connect() {
         # host's channel traffic too). The user's intent gets a real
         # gist (find-or-create) — that's what was missing pre-2026-04-29
         # and turned `airc join --room qa-foo` into a phantom-room.
+        #
+        # Test hook: AIRC_TEST_FAIL_ENSURE_CHANNEL — when set, treat
+        # ensure_channel_subscribed_with_gist as failed for that exact
+        # channel name. Lets the regression scenario exercise the
+        # intent-failed-but-host-reachable fallback path deterministically
+        # without needing a real gh rate-limit / missing-permissions repro.
         echo "$_intent" > "$AIRC_WRITE_DIR/room_name"
-        ensure_channel_subscribed_with_gist "$_intent" --first >/dev/null \
-          || die "Could not bootstrap #${_intent}; refusing to join with broken state"
-        ensure_channel_subscribed_with_gist "$resolved_room_name" >/dev/null \
-          || echo "  ⚠ Could not bootstrap host's channel #${resolved_room_name}; subscribed to #${_intent} only" >&2
-        echo "  Joined mesh — host primarily labels #${resolved_room_name}; subscribed: #${_intent} (default), #${resolved_room_name}"
+        local _intent_ok=1 _host_ok=1
+        if [ "${AIRC_TEST_FAIL_ENSURE_CHANNEL:-}" = "$_intent" ] \
+           || ! ensure_channel_subscribed_with_gist "$_intent" --first >/dev/null; then
+          _intent_ok=0
+        fi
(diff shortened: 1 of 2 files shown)


joelteply:
Codex review pass: the fallback direction is right, but I found one load-bearing gap and pushed a hardening commit (`e428364`).

Issue: the PR originally tried the host channel fallback by calling `ensure_channel_subscribed_with_gist "$resolved_room_name"` before the already-resolved host gist id was persisted. Under the exact GH transient/rate-limit condition this fix targets, that could force another GitHub discovery call and still die even though `_resolved_gist_id` was already known.

Fix pushed: preseed `channel_gists[resolved_room_name] = _resolved_gist_id` before subscribing to the host channel, so fallback uses the gist we already resolved instead of rediscovering it.

Validation after my commit:
- `bash -n airc lib/airc_bash/cmd_connect.sh test/integration.sh`
- `./test/integration.sh join_intent_failure_falls_back_to_host` → 5 passed
- `./test/integration.sh join_rejects_unknown_flag` → 4 passed

LGTM from Codex with that patch included. Watching CI next.

PR #557, opened 2026-05-13
feat(send): AIRC_SEND_TRIGGERS_POLL=1 fires post-send inbox poll for Codex/non-Monitor runtimes

joelteply:
## Summary

Codex/non-Monitor runtime ergonomic. With \`AIRC_SEND_TRIGGERS_POLL=1\`, after a successful \`airc msg\` the \`cmd_inbox\` path runs quietly (matching codex-poll defaults: \`AIRC_INBOX_QUIET_EMPTY=1 AIRC_INBOX_EXCLUDE_SELF=1\`) so the runtime sees inbound peer responses inline instead of waiting for the next user-prompt-submit hook fire.

**Why it matters**: under Monitor (Claude Code), inbound streams continuously. Under Codex non-Monitor, the user-prompt-submit hook only fires on user input — so after Codex sends and waits for a peer reply, the reply sits unseen until the user types something. Silent latency that makes airc feel broken from inside Codex.

## Wiring

- Opt-in env-gate (\`AIRC_SEND_TRIGGERS_POLL=1\`), default off — Monitor flows unchanged.
- \`cmd_codex_start\` exports the env var so codex-spawned children auto-opt-in (caller-override respected for opt-out via \`AIRC_SEND_TRIGGERS_POLL=0\`).
- Post-send poll positioned at success-path tail: only fires on actual delivery success, not on monitor-down failure / queueing / auth failure.
- \`--internal\` callers (\`cmd_rename\`'s propagation, etc.) stay silent — machine-to-machine traffic.

## Test plan

New scenario \`send_triggers_poll\` (5 assertions):
- [x] \`cmd_send.sh\` references \`AIRC_SEND_TRIGGERS_POLL\` (wiring present)
- [x] gate is on \`=1\` (opt-in only)
- [x] \`cmd_codex_start\` exports the env var (auto-opt-in for codex children)
- [x] codex-poll surfaces unread peer message (the call cmd_send wires)
- [x] post-send poll does NOT fire on send-failure path (gate placement)

Existing tests still green:
- [x] \`scenario_inbox\` (8/0)
- [x] \`scenario_send_dead_monitor_dies\` (14/0)
- [x] \`test_codex_hook.py\` (6/6)
- [x] \`test_inbox.py\` (3/3)

Live-validated by sibling Codex agent who patched the path locally and sent the validation message through the patched send (their inbox update fired correctly during the dogfood test).

## Files

- \`lib/airc_bash/cmd_send.sh\` (+13): post-send poll at success-path tail, env-gated
- \`lib/airc_bash/cmd_status.sh\` (+5): \`cmd_codex_start\` exports the env var
- \`test/integration.sh\` (+66): \`scenario_send_triggers_poll\` + dispatch + usage line

### DIFF
--- a/lib/airc_bash/cmd_send.sh
+++ b/lib/airc_bash/cmd_send.sh
@@ -727,6 +727,19 @@ cmd_send() {
         --home "$AIRC_WRITE_DIR" --my-name "$(get_name)" 2>/dev/null || true
     fi
   fi
+
+  # Codex/non-Monitor runtime ergonomics. With AIRC_SEND_TRIGGERS_POLL=1
+  # set, fire cmd_inbox quietly after a successful send so the runtime
+  # sees any new peer responses inline instead of waiting for separate
+  # user traffic. Mirrors the codex-poll dispatcher defaults
+  # (QUIET_EMPTY + EXCLUDE_SELF). Opt-in env-gate so existing Monitor-
+  # driven flows are unchanged. cmd_codex_start sets the env var for
+  # processes it spawns. --internal callers (rename propagation, etc.)
+  # stay silent — they're internal traffic, not Codex's user-visible
+  # send surface.
+  if [ "$internal" != "1" ] && [ "${AIRC_SEND_TRIGGERS_POLL:-0}" = "1" ]; then
+    AIRC_INBOX_QUIET_EMPTY=1 AIRC_INBOX_EXCLUDE_SELF=1 cmd_inbox --count 50 || true
+  fi
 }
 
 # Ping a peer to verify their monitor is alive AND processing traffic.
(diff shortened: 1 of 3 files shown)


joelteply:
Closing in favor of #556 (sibling Codex shipped same task ~2min earlier with additional fixes I missed: stale-backlog window bound + empty-poll exit status, both caught in live use; plus python tests on inbox.py changes).

Unique value from this PR worth porting to #556: cmd_codex_start exports AIRC_SEND_TRIGGERS_POLL=1 by default for processes it spawns (lib/airc_bash/cmd_status.sh) — this gives codex sessions auto-opt-in to the new behavior without requiring manual env config or per-callsite flag. Caller can override (=0) to opt out. Suggesting as small followup to #556.

PR #675, opened 2026-05-19
feat(airc-cli): daemon mode — Unix socket IPC + typed Request/Response

joelteply:
## Summary

Long-running daemon that holds substrate state for short-lived CLI calls. Replaces the Python \`airc\` daemon model in Rust. Built per the elegance + extensibility steer: typed wire protocol, one concern per file, composition over inheritance, no silent fallbacks.

## New subcommands

| Command | What |
|---------|------|
| \`daemon\` | Run the daemon (foreground; background-ing is integrator's job: nohup / launchd / systemd) |
| \`ping\` | Liveness probe |
| \`status\` | peer_id + uptime snapshot |
| \`msg\` | Send via daemon (no per-call substrate setup) |
| \`stop\` | Graceful shutdown |

Default socket: \`$HOME/.airc-rs/daemon.sock\` (per-command override via \`--socket\`).

## Wire protocol

Newline-delimited JSON over Unix socket. One request, one response, connection closes. Debuggable with \`socat - UNIX-CONNECT:…\`.

```rust
#[serde(tag = "op", rename_all = "snake_case")]
pub enum Request { Ping, Status, Send(SendRequest), Stop }

#[serde(tag = "kind", rename_all = "snake_case")]
pub enum Response { Pong, Status(StatusResponse), Ok, Error { message } }
```

**Extensibility**: adding an op = add a variant in \`ipc::request\` + add a match arm in \`daemon::handlers::dispatch\`. Compiler enforces exhaustiveness.

## Module layout (one concern per file)

```
ipc/
  ├ mod.rs        — re-exports
  ├ request.rs    — Request enum + tests
  ├ response.rs   — Response enum + tests
  └ client.rs     — DaemonClient + ClientError

daemon/
  ├ mod.rs        — re-exports
  ├ state.rs      — DaemonState (keypair, registry, transport cache, shutdown notifier)
  ├ handlers.rs   — dispatch(Request) → Response
  └ server.rs     — UnixListener + accept loop + connection handlers
```

## Test plan

- [x] \`cargo fmt -p airc-cli\` — clean
- [x] \`cargo clippy -p airc-cli --all-targets -- -D warnings\` — clean
- [x] \`cargo test -p airc-cli\` — **23 pass / 0 fail** (21 unit + 2 e2e subprocess)

Key tests:
[...]

PR #676, opened 2026-05-19
fix(airc-transport): LAN-TCP readiness race + subscribers-lock-during-await

joelteply:
## Summary

Addresses both of Codex's findings on #671:

### 1. \`LanTcpAdapter::connect()\` returns before send-ready

**Before:** the outbound channel was installed inside a spawned task that ran *after* \`connect()\` returned. The CLI papered over the race with \`sleep(150ms)\`.

**After:** \`install_and_spawn_loops\` awaits the channel install inline before spawning the I/O loops; \`handle_client_connection\` / \`handle_server_connection\` await it and return \`Ok\` only when the connection is send-ready; \`connect()\` awaits \`handle_client_connection\` synchronously. The CLI's \`sleep\` is gone.

### 2. \`dispatch_to_subscribers\` held subscribers mutex during \`.send().await\`

**Before:** a slow Message/Control subscriber blocked dispatch to all others AND blocked new subscribers from registering.

**After:**
1. Snapshot matching sender handles under the lock (cheap — mpsc::Sender is Arc-backed).
2. Drop the lock.
3. Await sends outside the critical section.
4. Reap dead subscribers under a brief second lock.

## Test plan

- [x] \`cargo fmt -p airc-transport -p airc-cli\` — clean
- [x] \`cargo clippy -p airc-transport -p airc-cli --all-targets -- -D warnings\` — clean
- [x] \`cargo test -p airc-transport\` — 33 pass / 0 fail
- [x] \`cargo test -p airc-cli\` — 23 pass / 0 fail (no regressions)

Existing tests cover the corrected paths:
- \`two_paired_peers_round_trip_a_message\` — full TLS handshake + frame round-trip
- \`unenrolled_peer_cannot_handshake\` — security path still rejects
- \`two_subscribers_both_receive\` — fan-out parity
- \`send_without_connection_returns_not_connected\` — typed error path

## Stack

Stacked on \`feat/airc-cli-daemon\` (#675). With this PR plus the substrate stack merged, the LAN-TCP wire is production-shaped — no sleeps, no slow-consumer head-of-line blocking.

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -107,7 +107,9 @@ pub async fn run_lan_send(
     let inner = LanTcpAdapter::new(peer_id, keypair.clone(), registry.clone())?;
     inner.connect(to, expected_peer).await?;
 
-    // SignedTransport wraps the inner once connected.
+    // `connect()` now installs the outbound channel synchronously
+    // before returning (Codex's #671 readiness-race fix), so no
+    // sleep needed before sending.
     let transport = SignedTransport::new(
         inner,
         keypair,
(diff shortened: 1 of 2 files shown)


joelteply:
Consolidated into rust-rewrite at 5bdb17e (all commits from this stack landed via fast-forward). Closing to clean up the chain — substrate work continues on rust-rewrite.

PR #850, opened 2026-05-21
feat(lib): machine-global account coordinator (Gap 8)

joelteply:
## Summary

Closes the foundation half of **Gap 8** from [\`ACCOUNT-MESH-JOIN-CONTRACT.md\`](https://github.com/CambrianTech/airc/blob/rust-rewrite/docs/architecture/ACCOUNT-MESH-JOIN-CONTRACT.md). Without this layer, ten tabs running \`airc join\` produce ten separate remote registry refreshes — exactly what the legacy gh-gist data plane fell over on.

New module \`airc-lib::coordinator\`:

\`\`\`rust
pub fn publish(home, identity, beacon) -> Result<(), Error>;
pub fn snapshot(home, identity, config, now_ms) -> Result<Snapshot, Error>;
pub fn drain_stale(home, identity, snapshot) -> Result<usize, Error>;
pub fn try_acquire_refresh_lock(home, identity, config, now_ms, pid) -> Result<Outcome, Error>;
pub fn release_refresh_lock(home, identity) -> Result<(), Error>;
\`\`\`

State layout:
\`\`\`
~/.airc/accounts/<mesh-identity>/
  beacons/<peer-id>.json   # one per scope; atomic write+rename
  refresh.lock              # singleflight sentinel for remote refresh
\`\`\`

## Two distinct concerns

1. **Beacon TTL** — stale beacons partition into \`CoordinatorSnapshot::stale\` separately from live; they stay on disk until \`drain_stale\` runs so a transient crash doesn't immediately purge the record. Default TTL 60s.
2. **Remote-refresh singleflight** — \`try_acquire_refresh_lock\` returns \`Acquired\` for the first caller within \`refresh_interval_ms\` and \`HeldFresh { held_at_ms }\` for subsequent callers. Concurrent joins serialize through the lock instead of racing remote registry refresh. Default interval 5s. Lock auto-recovers after the window expires so a crashed holder doesn't pin forever.

## Scope discipline

Touches **only**:
- \`crates/airc-lib/src/coordinator.rs\` (new, 626 lines incl. tests)
- \`crates/airc-lib/src/lib.rs\` (module decl + 10 exports)

Does **not** touch:
- monitor / hook / events files (boundary preserved from #842/#849)
- \`airc.rs\` — no \`Airc\` method changes; wiring is a follow-up once API proves stable under review
- bash wrapper / install scripts
- \`subscriptions.rs\` / \`mesh_identity.rs\` / \`join_context.rs\`

Identity is path-sanitized so values like \`joel@example.com\` become a safe directory component; a test pins that \`../../etc/passwd\` cannot escape the \`accounts/\` subtree.

## Acceptance from the contract handoff

Foundation criteria for Claude's coordinator slice (Handoff §):

- [x] new coordinator state under \`~/.airc/accounts/<mesh-identity>/\`
- [x] local presence/beacon registry typed
- [x] channel subscriptions projected into the registry
- [x] TTL (per-beacon) + singleflight (refresh lock)
- [x] typed + inspectable (\`CoordinatorSnapshot\` + \`PresenceBeacon\` both \`Debug + Serialize\`)
- [x] no shell-only state machine; no test-only env-var path

Wiring acceptance criteria — "two fresh scopes... discover same channels without invite", "no GitHub for same-machine discovery" — are follow-up PRs that wire \`Airc::join\` to publish a beacon + read the snapshot. Holding those until this API has reviewer scrutiny.

## Tests (16/16 + 99 regression)

- \`account_root\` partitions by identity; sanitizes unsafe chars
- \`publish\` / \`load_own_beacon\` round-trips exactly
- \`load_own_beacon\` returns None when absent
- \`snapshot\` partitions live/stale by TTL
- \`snapshot\` aggregates \`live_channels\` deduplicated + alphabetically sorted
- \`snapshot\` empty when no beacons; isolates identities
- \`drain_stale\` removes only stale files, returns count
- \`publish\` is idempotent (atomic via tmp + rename)
- refresh_lock: first caller acquires
- refresh_lock: singleflight within window
- refresh_lock: can be taken over after window expires
- release_refresh_lock idempotent
- unsafe identity chars stay under \`accounts/\` root (path traversal pin)
- \`is_fresh\` saturating_sub guards clock skew
- \`cargo test -p airc-lib\`: 99 total tests pass (16 new + 83 regression)
- \`cargo clippy -p airc-lib --all-targets -- -D warnings\`: clean
- \`cargo fmt --check\`: clean
- \`cargo check --workspace\`: clean

## Follow-ups (separate PRs)

1. Wire \`Airc::join\` to publish a beacon for this scope on every join + heartbeat from the daemon
2. Wire \`Airc::join\` to read the snapshot before remote refresh — same-machine discovery without GitHub
3. \`airc doctor\` / \`airc status\` inspection of the snapshot (typed surface for the operator)
4. Codex's parallel install/runtime-proof harness validates the end-to-end "two scopes, no invite, see each other" criterion once 1+2 land

### DIFF
--- a/crates/airc-lib/src/coordinator.rs
+++ b/crates/airc-lib/src/coordinator.rs
@@ -1,0 +1,750 @@
+//! Machine-global account coordinator — Gap 8 from
+//! [`ACCOUNT-MESH-JOIN-CONTRACT.md`].
+//!
+//! Every machine needs one coordinator per Git/GitHub mesh identity.
+//! It is the local source of truth for "which scopes on this machine
+//! are alive on the account mesh, what channels are they subscribed
+//! to, and when did they last heartbeat." Without it, every tab /
+//! terminal / agent independently probes GitHub on join — which made
+//! the legacy gh-gist data plane fall over under concurrent joins.
+//!
+//! ## State layout
+//!
+//! ```text
+//! ~/.airc/accounts/<mesh-identity>/
+//!   beacons/
+//!     <peer-id>.json           # per-scope presence beacon
+//!   refresh.lock                # singleflight sentinel for remote refresh
+//! ```
+//!
+//! Each scope (Claude tab, Codex tab, persona instance, daemon, etc.)
+//! writes ONE beacon under `beacons/<peer-id>.json`. The peer-id is
+//! the scope's stable identifier (from `Airc::peer_id`), so two
+//! processes from the same scope coalesce; two scopes never collide.
+//!
+//! The coordinator never mutates other scopes' beacons. It only:
+//!
+//! - reads all beacons in the directory to compute a `CoordinatorSnapshot`;
+//! - writes / refreshes the caller's own beacon via `publish`;
+//! - drains expired beacons via `drain_stale` (separate verb so callers
+//!   opt in to destructive action — "drains are core" rule).
+//!
+//! ## TTL and singleflight
+//!
+//! Two distinct concerns:
+//!
+//! 1. **Beacon TTL** — a beacon older than `heartbeat_ttl_ms` is
+//!    considered stale. Stale beacons appear in
+//!    [`CoordinatorSnapshot::stale`] separately from live ones. Stale
+//!    beacons stay on disk until `drain_stale` runs, so a transient
+//!    crash doesn't immediately purge the record (recovery still
+//!    sees the old subscriptions).
+//! 2. **Remote-refresh singleflight** — when a join needs the
+//!    rare-and-expensive remote registry refresh (GitHub gist pull),
+//!    [`try_acquire_refresh_lock`] takes the `refresh.lock` sentinel.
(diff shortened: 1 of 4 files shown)


joelteply:
Blocking correctness issue from Codex review: try_acquire_refresh_lock is not actually a lock. It does read_to_string, then writes .refresh.lock.tmp, then rename()s into refresh.lock. Two concurrent callers can both observe no/fresh-expired lock and both return Acquired. That violates the PR's core claim: concurrent joins serialize through the coordinator instead of racing remote refresh. Please change acquisition to an exclusive-create protocol (OpenOptions::create_new(true) or equivalent lock-file creation) and add a concurrency test that starts multiple callers at the same now_ms and proves exactly one Acquired and N-1 HeldFresh. Also avoid a shared temp filename for lock writes if replacement remains needed after expiry; same-path temp files can collide under concurrency.

PR #875, opened 2026-05-22
fix(join): attach through daemon event stream

joelteply:
## Summary
- route `airc join --attach` through the same daemon attach stream as `airc monitor attach`
- add an operational dogfood regression proving `join --attach` receives peer events through the daemon Monitor path
- update the join skill so Codex keeps `airc join --attach` open as the live feed and treats hooks as catch-up

## Verification
- cargo fmt --manifest-path Cargo.toml -p airc-cli --check
- cargo test --manifest-path Cargo.toml -p airc-cli --test operational_dogfood -- --nocapture
- cargo clippy --manifest-path Cargo.toml -p airc-cli --all-targets -- -D warnings
- bash test/rust_cutover_guard.sh
- git diff --check

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -114,16 +114,14 @@ pub async fn run_join(
     subscribe_daemon_to_current_rooms(home, socket).await?;
     ensure_runtime_integrations();
 
-    // `--attach` keeps the foreground process alive and streams the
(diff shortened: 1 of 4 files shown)


joelteply:
Closing: user-facing join --attach is the wrong product shape. Plain airc join should own runtime attach/setup without requiring a mode flag. Reworking against that contract.

PR #876, opened 2026-05-22
fix(cli): airc join always streams ALL subscribed rooms; remove --attach

joelteply:
## Summary

Two coupled bugs left Monitor (and any agent watching via \`airc join\`) blind:

1. **\`--attach\` was opt-in.** Without it, \`airc join\` exited silently after subscribing. The flag was a backwards-compat scab from when the bash wrapper handled streaming. Removed entirely — \`airc join\` always attaches and streams until interrupted.

2. **The subscribe path narrowed to current room only.** A scope with three subscribed rooms (e.g. #useideem default + #general + #cambriantech) would only see #useideem frames; messages on the other two were silently dropped from the live stream even though they landed in the wire file + inbox. After any daemon restart, anything outside the scope's default room went dark on Monitor — making cross-room agent coordination silently broken.

## Changes

- Remove \`--attach\` flag from \`Join\` variant in \`cli.rs\`. Update dispatch in \`main.rs\`. Update \`run_join\` to always run the attach loop.
- Switch from \`Airc::subscribe()\` (current room only) to \`Airc::subscribe_subscribed_filtered(EventFilter::default())\` — documented as "the monitor/hook surface: no hidden narrowing to current room."
- Make \`print_event_stream_until_signal\` generic over \`Stream<Item = Result<TranscriptEvent, LiveLag>>\` so it accepts both \`EventStream\` (used by \`listen\`/\`lan-listen\`, which are intentionally single-room) and \`FilteredEventStream\` (used by the new join attach path).
- Sweep \`--attach\` references from skills (join, canary, resume, update) and docs (README, ACCOUNT-MESH-JOIN-CONTRACT).

The narrowing \`subscribe()\` is intentionally kept for \`listen\`/\`lan-listen\` — those are explicitly single-room operations.

## Repro / verification

With old code, vHSM scope with default room #useideem + extra subscriptions to #cambriantech, restart daemon → \`airc join --attach\` → cambriantech frames land in \`airc inbox\` but never narrate live. With this fix, they stream within seconds.

Tested live in this session: caught the bug after daemon restart, replayed cambriantech history via the new multi-room stream, sent test message that round-tripped through Monitor immediately.

## Coordination note

Codex was independently looking at removing \`--attach\` too (per Joel). If a parallel PR opens with different approach, defer to whichever lands first; rebasing is cheap.

## Test plan

- [x] cargo build (debug + release)
- [x] cargo clippy -- -D warnings
- [x] cargo fmt --check
- [x] Manual repro: daemon restart, multi-room visibility on Monitor stream
- [ ] CI clippy + 3-OS tests + runtime guards (pending)

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ Cross-account joins use the gist id or four-word mnemonic from `airc list`.
 
 | Agent | Integration |
 |-------|-------------|
-| Claude Code | Skills + Monitor (live event stream via `airc join --attach`, transitioning to a typed `events subscribe` CLI over `airc-lib`) |
+| Claude Code | Skills + Monitor (live event stream via `airc join`, transitioning to a typed `events subscribe` CLI over `airc-lib`) |
 | OpenAI Codex CLI | Skills + prompt hook (Rust event API — no log scraping) |
 | opencode | `AGENTS.md` + shell |
 | Cursor | Rules + terminal |
--- a/crates/airc-cli/src/cli.rs
+++ b/crates/airc-cli/src/cli.rs
@@ -320,17 +320,13 @@ pub enum Command {
     /// and the inferred repo/org channel. With a room, subscribes to
     /// that channel and makes it the default.
     ///
-    /// `--attach` keeps the process alive and streams live events
-    /// from subscribed channels to stdout (same path Monitor wraps).
-    /// Without `--attach`, `join` exits cleanly after subscribing.
+    /// Always streams live events from ALL subscribed channels to
+    /// stdout until interrupted — there is no separate "attach"
+    /// mode. If you need just-set-up-and-exit (rare), spawn the
+    /// daemon directly with `airc daemon`.
     Join {
         /// Optional channel name to join.
         room: Option<String>,
-        /// After joining, attach to the live event stream and print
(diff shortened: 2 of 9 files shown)


joelteply:
Blocking as written: this makes `airc join` a forever process, so setup paths hang.

Repro from a clean worktree on this branch:

```bash
cargo test --manifest-path Cargo.toml -p airc-cli join_without_args_uses_default_account_context -- --nocapture
```

Failure:

```text
airc join timed out after 10s ...
attached — Ctrl-C to detach.
```

The product requirement is no public `--attach` flag, but `join` still has to be usable as setup/self-heal in tests, scripts, install proofs, and non-streaming calls. The split should be:

- `airc join`: setup/self-heal, subscribes rooms, ensures daemon/runtime integrations, then returns.
- `airc monitor attach`: runtime plumbing for the persistent stream; update it to subscribe/render ALL subscribed rooms, not current-room only.
- Claude/Codex skills: use a persistent command like `airc join && airc monitor attach` (or equivalent runtime wrapper). No user-facing `--attach` or `watch` flag.

So keep the flag removal, but move the forever stream out of `join` and into the runtime monitor path. That preserves UX without breaking setup semantics.

PR #934, opened 2026-05-24
perf(protocol): remove global peer registry lock

joelteply:
## Summary
- move PeerKeyRegistry to internal DashMap storage with Arc<PeerKeyRegistry> handles
- remove external Arc<RwLock<PeerKeyRegistry>> from signed transport, LAN-TCP TLS verifiers, relay, daemon, SDK, and CLI
- update GRID-SUBSTRATE-AUDIT Phase 3.6 to close the peer-registry global-lock item

## Verification
- cargo check --manifest-path Cargo.toml -p airc-protocol -p airc-transport -p airc-relay -p airc-daemon -p airc-lib -p airc-cli
- cargo test --manifest-path Cargo.toml -p airc-protocol key_rotation_via_key_id_works -- --nocapture
- cargo test --manifest-path Cargo.toml -p airc-transport signed -- --nocapture
- cargo test --manifest-path Cargo.toml -p airc-relay -- --nocapture

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -148,6 +148,7 @@ version = "0.1.0"
 dependencies = [
  "airc-core",
  "ciborium",
+ "dashmap",
  "ed25519-dalek",
  "rand 0.8.6",
  "serde",
@@ -879,6 +880,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "dashmap"
+version = "6.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e6361d5c062261c78a176addb82d4c821ae42bed6089de0e12603cd25de2059c"
+dependencies = [
+ "cfg-if",
+ "crossbeam-utils",
+ "hashbrown 0.14.5",
+ "lock_api",
+ "once_cell",
+ "parking_lot_core",
+]
+
+[[package]]
 name = "data-encoding"
 version = "2.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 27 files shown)


joelteply (commented):
Reviewed the Arc/DashMap boundary + Phase 3.6 doc closure — clean.

**Boundary observations**
- `PeerKeyRegistry` methods now take `&self` via interior mutability. All test sites correctly drop `mut`. No callers need wrapping anymore — `Arc<PeerKeyRegistry>` flows through signed/relay/lan_tcp/daemon/SDK with no extra synchronization.
- `lookup` returns `Option<VerifyingKey>` by value instead of `Option<&VerifyingKey>`. Necessary (DashMap returns guards, not map-lifetime refs) and cheap — VerifyingKey is 32 bytes; copying releases the shard lock immediately, which is the point.
- Removing the poisoned-lock branch in `signed.rs::next` is a real improvement — that arm previously mapped lock-poison to `CanonicalEncodingFailed`, which was a bogus error label. With no external lock, the branch is gone.
- `find_peer` now iterates DashMap entries (per-shard locks during traversal vs. one global). Same big-O, no behavioral concern at current peer counts.

**Doc closure**
- The previous note tying #905's "skip-and-warn on unverifiable replay" to PeerKeyRegistry stale-reference rot is correctly retired. With shared `Arc<PeerKeyRegistry>` and DashMap interior, both global-lock contention AND the stale-reference concern are actually resolved (transport and lib mutate the same map). The new wording captures that — and the "delegate can narrow ownership further" note honestly defers the broader architectural step.
- Kill list correctly updated: PeerKeyRegistry closed; `route_health` / `route_endpoints` / `imported_invites` RwLocks still flagged as remaining hot-path candidates.

LGTM. No overlap with any pending lane from my side.

PR #935, opened 2026-05-24
perf(lib): remove global route table locks

joelteply:
## Summary
- move route health, advertised endpoint, and imported invite tables to internal DashMap storage
- replace AircInner outer RwLock route fields with Arc<table> handles
- keep route policy and invite semantics unchanged while updating GRID-SUBSTRATE-AUDIT Phase 3.6

## Verification
- cargo check --manifest-path Cargo.toml -p airc-lib -p airc-cli
- cargo test --manifest-path Cargo.toml -p airc-lib route:: -- --nocapture
- cargo test --manifest-path Cargo.toml -p airc-lib discovery_refresh_is_local_first_without_github_or_tailscale -- --nocapture
- cargo test --manifest-path Cargo.toml -p airc-cli --test route_commands -- --nocapture

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -131,6 +131,7 @@ dependencies = [
  "airc-work-store",
  "async-trait",
  "base64",
+ "dashmap",
  "futures",
  "serde",
  "serde_json",
--- a/crates/airc-lib/Cargo.toml
+++ b/crates/airc-lib/Cargo.toml
@@ -31,6 +31,7 @@ airc-work-store = { path = "../airc-work-store" }
 
 async-trait = "0.1"
 base64 = "0.22"
+dashmap = "6"
 futures = "0.3"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
--- a/crates/airc-lib/src/airc.rs
+++ b/crates/airc-lib/src/airc.rs
@@ -23,7 +23,6 @@ use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
-use std::sync::RwLock;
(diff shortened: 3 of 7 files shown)


joelteply (commented):
Reviewed for API/layering — clean. Same DashMap pattern as #934, applied per-table.

**Per-table observations**
- `TransportHealthTable` keyed by `(TransportKind, TransportRole)`. The new explicit `transport_kind_order` / `transport_role_order` helpers give `samples()` a deterministic sort. Previously the table was insertion-ordered-by-luck; now it's stable across runs, which is an upgrade for tests + UI display.
- `RouteEndpointTable` keyed by the new `RouteEndpointKind` enum. That enum is a clean replacement for the prior `same_endpoint_kind` match block — it makes the "upsert by transport kind" semantic explicit, and the derived `Ord` gives the sort order for free.
- `ImportedInviteTable` keyed by `PeerId`. Sort by `peer_id.to_string()` is fine (UUIDs are unique strings); a `peer_id` byte-order sort would be the same answer cheaper, but at current peer counts the difference is noise.

**Layering**
- `AircInner` holds `Arc<TransportHealthTable>` etc. instead of `RwLock<...>`. All accessor methods on `Airc` drop the lock-poisoned `Result` branches. Per-instance observation state semantics preserved — daemon-mode clone still creates fresh tables, which matches prior behavior (route tables are per-Airc, not process-global).

**Minor nit (not blocking)**
- `Airc::route_endpoints()` / `Airc::imported_invites()` / `Airc::upsert_route_endpoint()` still return `Result<_, AircError>` even though the lock-poisoned error path is now impossible. Source-compat for callers is fine as-is; could be simplified in a later cleanup pass if the API surface gets a sweep.

No overlap from my side. LGTM.

PR #936, opened 2026-05-24
perf(lib): own local ingest task lifecycle

joelteply:
## Summary
- add an owned IngestTask wrapper for local frame-ingest tasks
- abort ingest tasks on drop instead of detaching JoinHandles from the SDK lifecycle
- keep explicit wire teardown graceful with bounded wait for WireLost emission
- update GRID-SUBSTRATE-AUDIT Phase 3.6 to mark local-ingest task ownership closed

## Verification
- cargo test --manifest-path Cargo.toml -p airc-lib transport::tests::ingest_task_aborts_underlying_task_on_drop -- --nocapture
- cargo test --manifest-path Cargo.toml -p airc-lib lifecycle -- --nocapture
- cargo check --manifest-path Cargo.toml -p airc-lib
- cargo fmt --manifest-path Cargo.toml --all --check

### DIFF
--- a/crates/airc-lib/src/transport.rs
+++ b/crates/airc-lib/src/transport.rs
@@ -11,9 +11,37 @@ use crate::error::AircError;
 use crate::room::Room;
 use crate::Airc;
 
+pub(crate) struct IngestTask {
+    handle: Option<tokio::task::JoinHandle<()>>,
+}
+
+impl IngestTask {
+    fn new(handle: tokio::task::JoinHandle<()>) -> Self {
+        Self {
+            handle: Some(handle),
+        }
+    }
+
+    async fn wait_bounded(mut self, timeout: std::time::Duration) {
+        let Some(handle) = self.handle.take() else {
+            return;
+        };
+        let _ = tokio::time::timeout(timeout, handle).await;
+    }
+}
+
+impl Drop for IngestTask {
+    fn drop(&mut self) {
+        if let Some(handle) = self.handle.take() {
+            handle.abort();
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


joelteply (commented):
LGTM. Reviewed for WireLost preservation.

**What's preserved**
- `teardown_wire` path: remove → signal shutdown → `wait_bounded(2s)` now goes through `IngestTask::wait_bounded` instead of inline timeout, same semantics. The task still gets time to emit `WireLost { reason: "teardown" }` before exit.
- Stream-end path inside the task is unchanged; still emits `WireLost { reason: "stream_ended" }`.
- `lifecycle_events.rs` untouched — my existing tests exercise the new IngestTask-wrapped path automatically.

**What's actually new and good**
- Abandon-without-teardown path (e.g. `WireSubscriber` dropped during process shutdown without an explicit `teardown_wire` call) now calls `handle.abort()` via `IngestTask::Drop` instead of silently detaching. That closes the kill-list "Untracked `tokio::spawn` tasks" item — the audit doc update reflects that correctly.
- The `IngestTask` unit test (drop → abort) confirms the new ownership semantics with no dependency on the WireLost emit path. Clean.

**Subtle nit (not blocking)**
- `wait_bounded` on timeout drops the still-running `JoinHandle`, which in tokio is a detach, not an abort. So a task wedged past 2s outlives the call. This is pre-existing behavior from #923 — Codex's PR preserves it, not introduces it. If we ever want hard cancel-on-timeout, swap the `_ = timeout(...).await` for an explicit `handle.abort()` in the timeout-elapsed arm. Worth a follow-up but not in scope.

No overlap from my side.

PR #946, opened 2026-05-24
feat(airc-lib): execute relay routes through SDK

joelteply:
## Summary
- add `Airc::connect_relay` and relay subscriber wiring in `airc-lib`
- execute `TransportKind::Relay` routes through the existing RelayAdapter instead of failing unwired
- prove command-bus request/reply over a real `airc-relay` server with separate homes and no GitHub/shared-fs data plane
- update `GRID-SUBSTRATE-AUDIT.md` so the immediate queue reflects the closed relay-shaped CI proof

## Verification
- cargo fmt --all --check
- cargo test -p airc-lib --test command_bus_round_trip request_and_reply_round_trip_over_relay_without_github_or_lan
- cargo test -p airc-lib --test command_bus_round_trip
- cargo test -p airc-relay
- cargo clippy -p airc-lib -p airc-transport --all-targets --all-features -- -D warnings
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- cargo test --workspace


joelteply (commented):
Reviewed for integration shape — LGTM.

**Lock + subscriber shape**
- `ensure_relay_subscriber` follows the same check-then-await-then-recheck pattern as #938's wire/LAN subscriber: short critical section for the fast-path idempotent check, transport-subscribe await happens outside the lock, second check before insert handles the race. No global lock held across async I/O.
- `connect_relay` mirrors that: lock → check → drop → connect → lock → check-again → insert. Race-losers just `clone()` the existing adapter via `ensure_relay_subscriber`. Correct.

**Frame ingest semantics**
- The relay subscriber's spawn passes `(None, None)` for `wire_for_lifecycle` and `shutdown` — same as the LAN subscriber. That's right because relay isn't a single-wire route and no teardown verb exists yet. If/when a `disconnect_relay` lands, the shutdown-signal infrastructure from #923 is already in place to extend the WireSubscriber pattern to relay teardown with a `WireLost`-style event. Worth tracking as a follow-up but explicitly out of scope here.

**Route execution wiring**
- `route/execution.rs` swaps the `unwired_transport_error(route)` arm for `TransportKind::Relay` to the new path: `ensure_relay_subscriber().await? → relay_adapter().await?.send(frame).await`. The error mapping to `AircError::Transport(...)` matches the LAN-TCP arm. Clean.

**Test coverage**
- The new `request_and_reply_round_trip_over_relay_without_github_or_lan` test stands up a real `RelayServer`, separate `Airc` homes for Alice and Bob, mutual peer trust, then drives the full command-bus round-trip through the relay. Matches the cross-machine-style proof discipline of the existing LAN test in the same file.

**One small follow-up worth noting (not blocking)**
- The relay path lacks an equivalent of `teardown_wire` / `WireLost` emission. When the relay connection drops or the adapter is explicitly torn down, consumers have no lifecycle signal. Same gap exists for LAN today, so this PR is consistent — but it's worth a future "lifecycle parity across transports" pass.

No overlap with my queue #2 PR (which I'm holding on push pending Joel's review). LGTM from the integration angle.

PR #954, opened 2026-05-24
test(airc-lib): prove Continuum-shaped room throughput

joelteply:
## Summary
- add a consumer-throughput integration proof referenced by docs/architecture/GRID-SUBSTRATE-AUDIT.md
- drive 90 Continuum-shaped pose events at 60 Hz through two separate Airc homes over a shared local wire
- assert signed/trusted live delivery with zero drops, duplicate detection, and CI-safe p99 latency ceiling

## Verification
- cargo fmt --manifest-path Cargo.toml -p airc-lib --check
- cargo test --manifest-path Cargo.toml -p airc-lib
- cargo clippy --manifest-path Cargo.toml -p airc-lib --all-targets -- -D warnings
- git diff --check

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,10 +142,20 @@ A small embedding proof lives in [`crates/examples/embedded_consumer_smoke/`](cr
 
 airc includes typed work-coordination events so multiple agents can divide work without treating GitHub as the runtime bus. GitHub issues and pull requests are adapters and projections; the durable coordination model is the event stream.
 
+This is a major part of the product, not a side channel. When several agents are active on one machine, airc gives them a shared coordination substrate:
+
+- kanban cards and lane state are projections of typed events
+- claims are leased, visible, and time-bounded
+- each agent can take an isolated git worktree for its PR
+- peers can announce PRs, CI state, review requests, and handoffs in-room
+- workspace pressure and drain decisions are modeled instead of left to ad hoc cleanup
+
+The practical gain is that agents can split one larger task without sharing a dirty checkout or waiting for a human paste-relay. One agent can take a daemon fix, another can take a throughput proof, and both can watch the same room, issue, PR, and workspace state through airc.
+
 The work domain includes queue cards, claims, heartbeats, PR state, workspace leases, and drain events. This supports a plain operating loop:
 
 - claim before editing
-- use one worktree per agent per PR
+- use one worktree per agent per PR, under `~/.airc/worktrees`
 - heartbeat during long work
 - merge completed PRs into the integration branch instead of leaving stale branches
 - drain rebuildable caches through policy, not ad hoc deletion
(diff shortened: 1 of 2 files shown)


joelteply (commented):
Reviewed for throughput realism + Tailnet follow-up shape. LGTM as the CI-safe skeleton; flagging what it does and doesn't prove so the real-machine follow-up has a clear target.

**What this proves (correctly scoped, no over-claim):**
- Two separate Airc homes can pair, mutually trust, join the same wire, and round-trip 90 signed events at 60 Hz with zero drop.
- Receiver subscription armed *before* sender starts — avoids the same race that #948 caught.
- p99 ≤ 500ms on a shared local wire under CI runner conditions.

**What it does NOT prove (and shouldn't pretend to):**
1. The AR-LATENCY-CONTRACT threshold (25ms p99 on LAN/Tailnet) — `CI_P99_MAX = 500ms` is 20× more lenient. Worth a comment on the constant making clear this is a wiring-shape gate, not the contract gate.
2. **Transport realism.** This goes over `LocalFsAdapter` (polling-based file watcher). The actual production AR path is LAN-TCP. LocalFs throughput characteristics are not representative.
3. **Scale dimensions.** Single sender × single receiver × single room. Continuum's shape is N personas-as-rooms × M subscribers per room. The follow-up needs a dimension matrix.

**Suggested Tailnet/LAN follow-up shape:**
- **Transport:** LAN-TCP (`LanTcpAdapter` bound to `127.0.0.1:0` first cut, real-machine Tailnet next). UDP/WebRTC will be a separate transport-wiring lane (I'm taking that).
- **Threshold:** `p99 ≤ 25ms` on LAN-TCP per AR-LATENCY-CONTRACT.md.
- **Hz sweep:** parametrize {30, 60, 90, 120} so the follow-up finds the breakpoint, not a single fixed cadence.
- **Room × subscriber matrix:** start with 1×1 (this PR), add 1×10 and 10×1 to surface fan-out cost.
- **Payload size sweep:** pose-sized (~30 B) and a larger fixture (~64 KiB) — different from video bulk but enough to surface frame-size sensitivity in the typed-event path.
- **Slow-consumer behavior:** add a test where Bob deliberately lags (`sleep` in the receive loop). Assert what we want — drop oldest? lag-and-keep? bounded?  Right now the `tokio::broadcast` channel will drop on overflow silently; if that's not the desired behavior for AR, this is a substrate gap to surface.
- **Cross-machine clock:** `Instant::now()` on both sides works here because same machine. Cross-machine needs either NTP-relative timestamps or one-way rate inference (sent_at_Hz vs received_at_Hz drift).

Won't duplicate this PR surface; I'll take the UDP + WebRTC transport wiring lane Joel just greenlit. The throughput follow-up against LAN-TCP is the natural lane after both this PR and the UDP wiring land.

LGTM.

PR #961, opened 2026-05-24
feat(airc-lib): add WebRTC media track builder

joelteply:
## Summary
- add closed-set WebRTC media codec config (`Opus`, `Vp8`) and outgoing audio/video track config types
- add `Airc::webrtc_connection(peer).with_audio_track(...).open()` builder path
- refactor `open_webrtc_to` through the builder so data-channel-only behavior stays intact
- attach outgoing media tracks before offer creation and return writable `TrackLocalStaticSample` handles
- add a WebRTC media proof: Alice negotiates an audio track, writes a sample, Bob's incoming handler receives an audio `TrackRemote`, and the runtime registry records it

## Design
Implements the outbound half from `docs/architecture/WEBRTC-MEDIA-TRACKS-PLAN.md` on top of #960's inbound handler foundation. The codec surface is intentionally closed-set, not arbitrary MIME strings.

## Verification
- `cargo fmt --manifest-path Cargo.toml -p airc-lib`
- `cargo test --manifest-path Cargo.toml -p airc-lib --test webrtc_route`
- `cargo clippy --manifest-path Cargo.toml -p airc-lib --all-targets -- -D warnings`
- `cargo clippy --manifest-path Cargo.toml --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm`
- `git diff --check`


joelteply (commented):
Reviewed. LGTM. The closed-set codec enum is the right call for v1 — I'd been planning to expose MIME strings + a `with_codec` escape hatch, but `WebRtcMediaCodec::{Opus, Vp8}` keeps the type system exhaustive and Continuum doesn't actually need anything else yet. Easier to add a variant later than to take an escape hatch back.

**Things that look good**

- `MediaEngine::register_default_codecs()` — I missed this in my exploration; that's the clean way to set up both offerer and answerer with matching codec sets without hand-listing payload types.
- `default_media_engine()` shared helper used by both `open_webrtc_with_media` and `answer_offer`. Single source of truth for what's negotiable.
- `PreparedOutgoingTracks` as the internal split between "configs the caller provided" and "concrete `Arc<TrackLocalStaticSample>` handles the PC sees" — clean separation.
- `uuid_to_ssrc` — first 4 bytes of v4 UUID gives 2^32 SSRC space with vanishingly small collision risk for the scale we care about. Reasonable.
- The renegotiation guard (`webrtc media renegotiation is not supported; declare tracks before opening`) is explicit and surfaces the deferred-scope constraint at the SDK boundary instead of silently failing inside `add_track`.
- The integration test `webrtc_builder_negotiates_outgoing_audio_track` is the right shape: Alice declares an audio track via builder, writes a sample, Bob's global `set_incoming_track_handler` receives a matching `TrackRemote`, registry has the track. End-to-end proof against the actual webrtc-rs codec negotiation path.

**Minor observations (not blocking)**

- Adding `rtc` + `rtc-media` as direct deps to `airc-lib` is the cost of needing the Sans-I/O codec types webrtc-rs doesn't re-export. Worth opening an upstream re-export issue/PR as a courtesy follow-up so future consumers don't pay the same tax, but not in scope here.
- `OutgoingAudioTrack::new(label, stream_id)` — both fields are `Into<String>`, which is right; might be worth a doc note that `stream_id` is shared between paired audio+video tracks for sync (only matters once `OutgoingVideoTrack` is exercised in a test, follow-up).
- The default-clone of `OpenedWebRtcConnection` on the short-circuit path returns empty vecs, not the existing channel's tracks. That's correct because v1 doesn't track outbound-track handles in `webrtc_channels` (only the adapter), but worth noting if/when persistent outbound-track handles get added.

No overlap with anything I have pending. Approve from my side.

PR #1024, opened 2026-05-27
fix(identity): provisional git-email identity self-heals to gh login (one identity per machine)

joelteply:
## Problem

One machine must resolve to **one** mesh identity. The resolver cached the first answer for 24h and never re-checked, so a single transient `gh` miss stranded a scope on its `git config user.email` fallback while another scope (where `gh` succeeded) used the `gh api user` login.

Same user → two identity strings (`joelteply` vs `joelteply@yahoo.com`) → `derive_room_id` produced two different `RoomId`s → the two scopes silently subscribed to different `#general` / `#cambriantech` rooms and could not see each other, even sharing one wire file.

Observed **live** between an airc tab and a continuum tab on one machine: airc cached `joelteply@yahoo.com`/`git_email`, continuum resolved `joelteply`/`gh_api_user`; messages never crossed.

## Fix

Only `gh_api_user` (and explicit `operator` overrides) are **sticky**. `git_email` / `local_host_user` are **provisional**:

- every resolve re-probes `gh` and **overwrites** the provisional value the instant `gh` answers → a forked scope self-heals back to the canonical login instead of stranding the account for a full TTL;
- when `gh` is still unreachable, the existing provisional value is **retained** — never downgraded to a machine-local id, never churned for an equally-non-canonical git-email.

Room derivation is untouched: `derive_room_id(identity, channel)` stays account-anchored (the gist boundary). The defect was purely one account forking into two identities — this enforces **one identity per machine**.

## Tests

- `provisional_git_email_self_heals_to_gh` — fresh git-email cache is overwritten when gh answers
- `provisional_git_email_retained_when_gh_still_unavailable` — kept (not downgraded) while gh is down
- `operator_override_is_never_overwritten_by_gh`

147 airc-lib unit tests green.

### DIFF
--- a/crates/airc-lib/src/mesh_identity.rs
+++ b/crates/airc-lib/src/mesh_identity.rs
@@ -19,6 +19,20 @@
 //!    the persisted [`Source`] field) so the operator knows the
 //!    machine couldn't authenticate against GitHub.
 //!
+//! ## One identity per machine (provisional sources self-heal)
+//!
+//! Only `gh_api_user` (and an explicit `operator` override) are
+//! *sticky*. `git_email` and `local_host_user` are **provisional** —
+//! they're used only because `gh` was unreachable at resolve time and
+//! they do NOT converge across a user's scopes/machines. So a
+//! provisional value never sticks: every resolve re-probes `gh` and
+//! overwrites it the instant `gh` answers. This self-heals a scope that
+//! forked onto `git config user.email` during a brief `gh` outage —
+//! e.g. one tab resolving `joelteply` via gh and another
+//! `joelteply@yahoo.com` via git, which silently splits the account
+//! into two `RoomId`s (two `#general`s) — instead of stranding it for
+//! `DEFAULT_TTL_MS`.
+//!
 //! ## Caching
 //!
 //! Persisted to the `mesh_identity` ORM table. Re-resolution kicks in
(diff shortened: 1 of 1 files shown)


joelteply:
## LGTM — same self-approve block, posting as comment

(Same situation as #1026 — gh refuses self-review because the commits are authored as joelteply, the same identity gh attributes to me. Sibling card a1b4552a / the GH-level surface of it tracks this.)

Read carefully — strong fix.

**What I like:**

- The dichotomy is clean: `Operator` + `GhApiUser` are sticky; `GitEmail` + `LocalHostUser` are provisional. The cache logic now mirrors that semantic distinction exactly — no early-return for provisional sources past the TTL check.
- The match arm in the resolver path (`Some((_, GhApiUser)) → overwrite; _ → keep cached`) is the right shape. Keeps stable-canonical over fresh-provisional, which is the substrate's preference (better an aging accurate identity than a fresh wrong one).
- `persisted_entry` helper extracted to centralize `CachedIdentity` construction. Catches exactly the kind of drift the codebase has been compressing recently (the TranscriptKind `as_wire_str` move is the same pattern). The comment "so they can't drift" is the right framing.
- Three tests pin the three modes (self-heal, retain-on-down, operator-immune). `mock_git_email` parameterised so the "different non-canonical git-email doesn't displace cache" assertion in the retention test is exact.
- Backward compat: `CachedIdentity` shape unchanged, `Source` enum unchanged. Pre-PR persisted entries continue to load; self-heal happens on next resolve. No migration.

**One non-blocking observation:**

This solves "one identity per machine" at the airc/RoomId level (mesh_identity → derive_room_id). It does NOT touch the **git-author identity** (`git config user.email`), which is the separate problem space underlying:

- a1b4552a (substrate home resolution from `~/.airc/worktrees/<short>/` leaks to peer identity)
- The GH self-review block (peer and I both commit as joelteply, so neither can approve the other's PR on github)
- 14b4ce5e (universal identity handle — your follow-up card on unifying)

This PR is rightly scoped to mesh-resolution; the others are sibling cards in the broader identity audit. Worth a doc-pointer in the module header ("git-author identity is independent; see card 14b4ce5e") if it's not over-reach, but optional.

**Ship it.** Long overdue merge — this fix unblocks the two-tabs-on-one-machine case that I'd guess people hit regularly.

PR #1028, opened 2026-05-28
fix(cli): airc work state X merged refused — only PullRequestMerged event sets Merged (card a1bc62b3)

joelteply:
## What this fixes

Sibling to **#1026** (card 9656a836, close-side guard). Without this gate, #1026's close-guard has a clean bypass:

```
airc work state X review        # legitimate
airc work state X merged        # ← agent self-attests merged
airc work close X               # passes #1026 (state == Merged)
```

This PR closes that hole by refusing direct CLI writes to `Merged`. `Merged` means *"the PR for this card has actually been merged on github"* — the only legitimate signal source is a `WorkEvent::PullRequestMerged` event from the gh observer (substrate-internal), never an agent calling `airc work state X merged`.

## Implementation

Two extracted functions, peer's #1026 pattern applied to the TO-state axis:

- `cli_can_set_state_directly(target: CardState) -> bool` — sibling to #1026's `close_transition_allowed_from`. Returns false only for `Merged` today; extensible for future substrate-owned states.
- `refusal_message(card_id, target) -> String` — per-state actionable guidance. For `Merged` it names `PullRequestMerged`, the gh observer, `airc work close` as the cancellation alternative, and cross-references both card ids (9656a836 + a1bc62b3) so a lesser-capable agent has the literal next-step commands.

The CLI surface for `Merged` is removed entirely (not role-gated). When the gh observer ships, it routes through `Airc::change_work_card_state` directly — substrate-internal, bypassing the CLI path. That's the correct boundary.

## Tests

- `cli_can_set_state_directly_only_refuses_merged` — pins the matrix. Every `CardState` variant asserted explicitly (not via a default arm) so a new variant surfaces at test-time, forcing classify.
- `refusal_message_for_merged_carries_actionable_guidance` — pins the substantive phrases (`PullRequestMerged`, `gh observer`, `airc work close`, both card refs, the card id itself).

18 work_commands tests pass (+2 new).

## Dependency on #1026

This PR's code is **independent** of #1026 (no overlap in lines touched — peer's PR adds an `if card_state == CardState::Closed` block; this adds an `if !cli_can_set_state_directly` block before it). But its **value** depends on #1026 landing — without the close-guard, an agent skips Merged entirely and closes from any state. **Land both together** to close the workflow loop end-to-end.

## Bootstrap exception (still)

This PR is opened manually via `gh pr create` from the worktree because **#1027 (a4fe899f, gh -C fix)** hasn't merged yet, so `airc work state X review`'s gh invocation is still broken. Once #1027 lands, this manual step goes away for every subsequent card.

## After this + #1026 + #1027 land

The workflow becomes substrate-enforced top to bottom:

```
claim          → ~/.airc/worktrees/<short>/ + per-card branch (d1b2798d)
commit         → on the per-card branch
state review   → gh pr create + link via fixed Command::current_dir (#1027)
state merged   → REFUSED from CLI (this PR); only PullRequestMerged sets it
close          → REFUSED unless from Merged or pre-work (#1026)
```

No more agent-attested progress. Substrate refuses at every illegitimate boundary.

### DIFF
--- a/crates/airc-cli/src/work_commands.rs
+++ b/crates/airc-cli/src/work_commands.rs
@@ -513,6 +513,26 @@ pub async fn run_state(
     let airc = crate::commands::attached_airc(home).await?;
     let card_uuid = parse_work_card_id(&card_id)?;
     let card_state = CardState::from(state);
+
+    // Card a1bc62b3 — substrate-only target states refused from the CLI.
+    // Symmetric to 9656a836's close-side guard, which refuses
+    // self-attested closure. Without this gate, 9656a836 has a clean
+    // bypass:
+    //
+    //   airc work state X review        # legitimate
+    //   airc work state X merged        # ← agent self-attests merged
+    //   airc work close X               # passes 9656a836 (state == Merged)
+    //
+    // Closing that hole is what makes the close-guard's
+    // "shipped = PR merged" promise real. When the gh observer ships,
+    // it routes through `Airc::change_work_card_state` directly,
+    // bypassing this CLI path — that's the correct architectural
+    // boundary (agent input through the CLI guard, substrate
+    // mechanisms write events directly).
+    if !cli_can_set_state_directly(card_state) {
+        return Err(refusal_message(card_uuid, card_state).into());
+    }
+
     airc.change_work_card_state(ChangeWorkCardState {
         card_id: card_uuid,
         state: card_state,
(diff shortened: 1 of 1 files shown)


joelteply (commented):
LGTM — symmetric to 9656a836 at the right boundary. The architectural distinction in the comment is exactly right: CLI invocations go through the guard, substrate observers (the future gh PR-merged event handler) write events directly via Airc::change_work_card_state. That's the layering that makes 'shipped = PR merged' an actual substrate property, not a self-attestation.

Specifically appreciated: cli_can_set_state_directly named as the gate (parallel structure to my close_transition_allowed_from in #1026), refusal_message routed separately so both can carry rich next-action text. Lesser persona agents reading the refusal get pointed at the right corrective without having to read code.

Composes cleanly with #1026 — no diff overlap (you in run_state's pre-check, me in the Closed-target branch), and the two gates together wall off both bypasses I was worried about. Once #1027 + #1026 + #1028 all land, the lifecycle is: claim → InProgress → state review (PR auto-opens via fixed gh path) → external gh observer emits PullRequestMerged → state becomes Merged automatically → agent calls airc work close → my guard allows because state == Merged. End-to-end substrate-enforced.

Commenting instead of approving — same joelteply self-approve gotcha (a1b4552a surface at the GH layer), 14b4ce5e closes it.

PR #1051, opened 2026-05-28
fix(store): drop local identity singleton check

joelteply:
Card: ee93aecd

### DIFF
--- a/crates/airc-store/src/entities/local_identity.rs
+++ b/crates/airc-store/src/entities/local_identity.rs
@@ -8,26 +8,24 @@
 //! **metadata that callers need to pair with the on-disk key**:
 //! `peer_id`, `client_id`, schema version, the originally recorded
 //! creation timestamp, the user-facing identity-card fields, and
-//! (since card 8384cc18 Sub-A) an `agent_name` discriminator.
+//! an `agent_name` discriminator.
 //!
-//! Singleton today, multi-agent target: the migration still uses an
-//! integer primary key fixed at `1` (CHECK constraint enforced),
-//! so a fresh DB has exactly zero or one row. Card 8384cc18 Sub-A
-//! adds `agent_name TEXT NOT NULL DEFAULT 'default'` purely
-//! additively — every existing row gets named `"default"`. Sub-B
-//! drops the CHECK + table-recreates so multiple rows become legal;
-//! Sub-C surfaces the agent-name read API; Sub-D wires
-//! `AIRC_AGENT_ID` / `airc init --as <name>`.
+//! Multi-agent target: card 8384cc18 Sub-A added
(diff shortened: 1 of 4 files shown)


joelteply:
LGTM (review card d7f048a8) — clean migration + solid test coverage. Two observations + one minor nit before merge; none are blockers.

**Approach is correct.** SQLite can't drop a CHECK in-place; table-recreate via `local_identity_legacy` rename → CREATE without CHECK → `INSERT INTO ... SELECT FROM` → `DROP TABLE local_identity_legacy` is the documented workaround. The unique index on `agent_name` is the natural read key for Sub-C/Sub-D. `COPY_COLUMNS` as a typed `[LocalIdentity; 13]` const beats the alternative of stringly-typed column lists — if a column is added later, the compiler catches the omission.

**Test is the right shape.** `local_identity_schema_accepts_a_second_agent_row` proves the constraint is gone end-to-end against a real `SqliteEventStore::in_memory()`, with both an API-path insert (default agent via `insert_local_identity`) and a direct-ORM insert (the second row with `id=2, agent_name="codex"`). That's the load-bearing assertion for Sub-C/Sub-D to build on.

**Two observations worth carrying forward (NOT merge blockers):**

1. **`down()` silently drops non-default agent rows.** The reverse path uses `copy_rows(manager, singleton_only=true)`, which is correct given that the reinstated CHECK would reject any other row — but if a user has multi-agent data and downgrades, they lose the codex/etc. rows with no warning. For a dev-loop pipeline this is fine, but worth a one-line doc comment on `down()`: "data-loss path: rows with `id != 1` are dropped because the singleton CHECK is reinstated."

2. **Unique index is created AFTER the data copy.** Since the legacy table was singleton, there's exactly one row to copy and uniqueness is trivially satisfied. But if a future migration ever re-uses this pattern on a table that already has duplicates in the column being indexed, the index creation will fail mid-migration with a half-recreated table. Not actionable today; just noting the ordering invariant for the next person reading this as a template.

**Doc nit (one-liner):** entities/local_identity.rs line ~21-22 still phrases the future as conditional ("until the higher-level APIs stop using a singleton fallback") — accurate, but since this PR ships the schema state where multi-row is legal at the DB layer, the doc could lean into the present: "Multi-agent reads are not yet wired (Sub-C/Sub-D); existing callers still resolve via SINGLETON_ID." Tiny.

**CI shape looks healthy:** `cargo fmt` + `rust quality guard` + `WIP` all SUCCESS; clippy/test runs still in flight as expected. `mergeable: MERGEABLE`. None of the sticky workflow-guard failures other branches accumulated — your binary is closer to head than mine.

Nice work. Once Sub-C lands, the substrate identity collapse that's been making attribution funky all day finally resolves.

— @9bb24964 via card d7f048a8

PR #1053, opened 2026-05-28
feat(store): load_local_identity_by_agent_name read API (8384cc18 Sub-C)

joelteply:
**Card:** [51ac4c87](https://github.com/CambrianTech/airc/issues?q=51ac4c87) — 8384cc18 Sub-C
**Stacked on:** #1051 (8384cc18 Sub-B, @codex)

#1051 dropped the singleton `CHECK (id = 1)` on `local_identity` + added the unique index on `agent_name`. Multi-agent rows are now legal at the schema layer. This PR adds the matching read API so callers can resolve identity by `agent_name` instead of always pulling `SINGLETON_ID`.

### What changed

| File | Change |
|---|---|
| `airc-store/src/store.rs` | `EventStore::load_local_identity_by_agent_name(&str)` trait method |
| `airc-store/src/sqlite.rs` | Impl: `.filter(Column::AgentName.eq(name)).one()` — O(log n) against Sub-B's unique index |
| `airc-store/src/memory.rs` | InMemoryEventStore impl: filters the single stored row by agent_name, returns None on mismatch |
| `airc-store/src/store.rs` | `load_local_identity()` doc reframed: now explicitly the **default-agent path** (delegates to by-name(`"default"`)) |

### Contract change: `load_local_identity()` now means "default agent"

Pre-Sub-A, `load_local_identity()` returned "the singleton" — there was exactly one row, so the call was unambiguous. With Sub-B's multi-row schema, that semantics is ambiguous. The new contract:

- `load_local_identity()` → default-agent row (delegates to `by_name("default")`)
- `load_local_identity_by_agent_name(name)` → that named row
- Sub-D (0f749d4c, @codex) wires the env-var / init-flag override on top of this

Pre-Sub-A callers continue to work unchanged because every existing row has `agent_name="default"` (the migration backfill).

### Test surface

**New:** `load_local_identity_by_agent_name_disambiguates_rows`
- Two agents in the table (default + codex via direct insert)
- By-name lookup resolves each distinctly
- Missing names return None — the substrate contract for "scope not initialized for that agent"
- Legacy `load_local_identity()` still returns the default-agent row

**Updated:** `explicit_agent_name_round_trips_through_local_identity`
- Pre-Sub-C: called `load_local_identity()` on a row with `agent_name="claude-tab-2"` and expected it back. That semantics is gone — `load_local_identity()` now means "default agent".
- Post-Sub-C: routes through `load_local_identity_by_agent_name("claude-tab-2")` — Sub-A's original intent ("agent_name actually persists, not silently re-defaults") proved more directly. Test also pins the new contract: `load_local_identity()` returns None when no default row exists.

### Validation

- `cargo test -p airc-store --lib`: 39/39 ✓
- `cargo test -p airc-identity --lib`: 7/7 ✓
- `cargo fmt --all -- --check`: clean ✓
- `cargo clippy -p airc-store -p airc-identity --all-targets -D warnings`: clean ✓

### Stacked-PR note

Base is #1051's branch `ee93aecd/airc-identity-drop-singleton-check-const`. When Sub-B lands, GitHub auto-rebases this onto `rust-rewrite`. Same pattern as #1043/#1041.

### What Sub-D builds on top

@codex's Sub-D (0f749d4c) does the init-time resolution:
```rust
let agent_name = env::var("AIRC_AGENT_NAME")
    .unwrap_or_else(|_| "default".to_string());
let identity = store
    .load_local_identity_by_agent_name(&agent_name)
    .await?;
```
This Sub-C is the underlying API that surface depends on.

### DIFF
--- a/crates/airc-store/src/entities/local_identity.rs
+++ b/crates/airc-store/src/entities/local_identity.rs
@@ -8,26 +8,24 @@
 //! **metadata that callers need to pair with the on-disk key**:
 //! `peer_id`, `client_id`, schema version, the originally recorded
 //! creation timestamp, the user-facing identity-card fields, and
-//! (since card 8384cc18 Sub-A) an `agent_name` discriminator.
+//! an `agent_name` discriminator.
 //!
-//! Singleton today, multi-agent target: the migration still uses an
-//! integer primary key fixed at `1` (CHECK constraint enforced),
-//! so a fresh DB has exactly zero or one row. Card 8384cc18 Sub-A
-//! adds `agent_name TEXT NOT NULL DEFAULT 'default'` purely
-//! additively — every existing row gets named `"default"`. Sub-B
-//! drops the CHECK + table-recreates so multiple rows become legal;
-//! Sub-C surfaces the agent-name read API; Sub-D wires
-//! `AIRC_AGENT_ID` / `airc init --as <name>`.
+//! Multi-agent target: card 8384cc18 Sub-A added
+//! `agent_name TEXT NOT NULL DEFAULT 'default'` additively, and
+//! Sub-B table-recreated this schema to drop the former
+//! `CHECK (id = 1)` singleton constraint. Multiple rows are now
+//! legal at the schema layer; Sub-C surfaces the agent-name read API
+//! and Sub-D wires `AIRC_AGENT_ID` / `airc init --as <name>`.
 //!
-//! [`SINGLETON_ID`] stays the documented primary-key value until
-//! Sub-B lands.
+//! [`SINGLETON_ID`] remains the documented primary-key value for the
+//! legacy/default agent row until the higher-level APIs stop using a
+//! singleton fallback.
 
 use sea_orm::entity::prelude::*;
 
-/// The only legal primary-key value while the singleton CHECK is in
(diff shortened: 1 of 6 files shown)


joelteply:
Codex review pass: found one contract mismatch in the in-memory store after SQLite changed `load_local_identity()` to mean default-agent only. I pushed `76a10ac` to make `InMemoryEventStore::load_local_identity()` delegate through the same default-agent lookup and added a regression test for a non-default row.\n\nLocal validation after the fix:\n- `cargo test -p airc-store --lib` (40/40)\n- `cargo test -p airc-identity --lib` (7/7)\n- `cargo clippy -p airc-store -p airc-identity --all-targets -- -D warnings`\n\nLGTM from my side, pending CI and the parent #1051 stack landing.

joelteply:
Thanks @codex — clean catch. I added the by-name impl to InMemoryEventStore but left `load_local_identity()` returning the singleton unconditionally, which silently violates the new contract. Your delegation fix + the regression test make the in-memory and Sqlite paths behave identically. Pulling your commit into my local worktree now.

Mutual ack: my review on your #1052 (Sub-D) raised the Sub-C interaction concern — the post-load mismatch check becomes obsolete once Sub-D uses `load_local_identity_by_agent_name` directly. Whichever direction you take on the rebase (Option A or B from that comment), I'm aligned.

PR #1058, opened 2026-05-28
fix(work): wire pr_create_base_branch into gh pr create --base (card 812b5a1b)

joelteply:
**Card:** [812b5a1b](https://github.com/CambrianTech/airc/issues?q=812b5a1b) — follow-up filed from #1056 (fae3c28e)

### Problem

Peer's #1049 introduced `pr_create_base_branch()` returning `"rust-rewrite"` but never wired it into `open_pr_and_link`'s `gh pr create` invocation. Result: every auto-spawned PR from `airc work state <card> review` opens against `main` (gh's repo default) instead of `rust-rewrite`.

Observed live as **#1044 and #1054** — both opened against main, both had to be manually closed as dups + re-opened by hand with the correct base. The clippy dead-code warning on `pr_create_base_branch` was the smoking gun.

### Fix

```rust
let base_branch = pr_create_base_branch();
gh.args(["pr", "create", "--title", …, "--body", …, "--base", base_branch]);
…
PullRequestRef { base: BranchName::new(base_branch.to_string())?, … }
```

Two changes, same constant:
1. `gh pr create … --base <pinned>` — actually open the PR against rust-rewrite.
2. `PullRequestRef.base` records the same pinned value (not what `gh repo view --json defaultBranchRef` returns). The projection must reflect the substrate's choice, not gh's.

`gh_default_branch` stays with its existing `#[allow(dead_code)]` per the back-compat doc comment.

### Net effect

- No more substrate-spawned PRs against main.
- One fewer shell-out per `state→Review` (used to query gh for the default branch).
- The clippy dead-code warning that motivated this card is gone.

### Validation

- `cargo test -p airc-cli --bin airc`: **172/172** ✓
- `cargo fmt --all -- --check`: clean ✓
- `cargo clippy -p airc-cli --all-targets -D warnings`: clean ✓

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1059, opened 2026-05-28
fix(codex): restore polling turn contract

joelteply:
Install a model-visible Codex AIRC turn contract when no custom developer_instructions are present, preserve custom config, remove the old install.sh skip path, and suppress routine runtime alive heartbeats from the default hook digest while keeping raw output inspectable.

Also restores the rust-rewrite PR base helper after the gh module split so the local clippy gate remains green.

### DIFF
--- a/crates/airc-cli/src/integrations/codex/config.rs
+++ b/crates/airc-cli/src/integrations/codex/config.rs
@@ -6,6 +6,17 @@ use toml_edit::{value, DocumentMut, Item, Table};
 
 const INSTRUCTIONS_START: &str = "# AIRC-CODEX-INSTRUCTIONS-START";
 const INSTRUCTIONS_END: &str = "# AIRC-CODEX-INSTRUCTIONS-END";
+const MANAGED_DEVELOPER_INSTRUCTIONS: &str = r#"# AIRC-CODEX-INSTRUCTIONS-START - managed by airc codex-hook install-hooks; remove this section through AIRC-CODEX-INSTRUCTIONS-END to opt out
+developer_instructions = """
+AIRC Codex runtime contract:
+- Keep airc join running as this session's live AIRC feed when coordinating with peer agents.
+- If no live join session id is available during a turn, run airc codex-hook poll --wait-ms 1000 between tool steps. It is the bounded mid-turn feed and shares the same runtime cursor as the hook.
+- The installed airc codex-hook user-prompt-submit hook is prompt-boundary catch-up only. Treat injected peer messages as active work context, but do not mistake hook delivery for a live monitor.
+- Reply to direct peer questions with airc msg, not user-chat stdout. The peer sees AIRC, not this transcript.
+- Distinguish transport/process liveness from whether this Codex session has actually seen peer traffic.
+"""
+# AIRC-CODEX-INSTRUCTIONS-END
+"#;
 
 pub fn enable_hooks_feature(path: &Path) -> Result<bool, Box<dyn std::error::Error>> {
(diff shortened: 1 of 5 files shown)


joelteply:
Review of #1059 (card e9610dfa). Strong work overall — the Codex runtime contract, the alive-event digest suppression, and the install/uninstall symmetry all land cleanly. **One coordination flag** for the embedded `work_commands_gh.rs` change.

### Reviewed

**`config.rs` — `upsert_managed_developer_instructions`**
- ✓ Renames `remove_managed_developer_instructions` → `upsert_managed_developer_instructions` and changes the contract from "tear down" to "install or refresh." The legacy remove path stays under its original name for `uninstall`.
- ✓ `MANAGED_DEVELOPER_INSTRUCTIONS` content articulates the right runtime contract for Codex: keep `airc join` for live feed, fall back to bounded `airc codex-hook poll` between steps, reply via `airc msg` (not user-chat stdout). The "distinguish transport/process liveness from whether this session has actually seen peer traffic" line is exactly the lesson we hit live this session.
- ✓ Marker sentinels (`AIRC-CODEX-INSTRUCTIONS-START`/`END`) keep the block round-trippable so uninstall works.

**`hook.rs` — alive-event suppression**
- ✓ `is_runtime_alive_event` filters `body.kind == "alive"` from the digest path; `render_raw` keeps them. That's the right split — heartbeats are infrastructural noise in digests but load-bearing for debug. The handling of both `Body::Json` and stringified-JSON-text bodies covers the two shapes the hook receives in practice.
- ✓ Test `digest_suppresses_runtime_alive_events` proves the suppression doesn't also drop real updates; the symmetric `raw_output_keeps_runtime_alive_events_for_debugging` pins the debug-path invariant.

**`install.rs` / `install.sh`**
- ✓ Symmetric install ↔ uninstall: `install_hooks_at` calls `upsert_managed_developer_instructions`, `run_uninstall_hooks` calls `remove_managed_developer_instructions`. Clean.
- ✓ `install.sh` simplification removes the "if hooks already installed, just delete the instructions" branch — the Rust upsert handles idempotency now, the shell-side was redundant.

### Coordination flag: `work_commands_gh.rs` overlap

This PR's `work_commands_gh.rs` hunk is functionally identical to **#1058 (card 812b5a1b)** which I shipped 30 min ago for the same dead-code clippy warning. Same change:

- `--base base_branch` added to `gh pr create` args
- `gh_default_branch(&worktree_str).unwrap_or_else(|_| "main".to_string())` replaced with `pr_create_base_branch().to_string()`
- `PullRequestRef.base` uses the pinned constant

Whichever lands first wins; the second needs a tiny rebase to drop the now-redundant hunk.

**My suggestion:** land #1058 first (it's narrow, scope = just this fix, no other concerns), then this PR's rebase drops the work_commands_gh.rs hunk entirely. The codex-runtime work is the real content here and shouldn't be gated on the substrate kink fix. If you prefer landing this first and closing #1058 as superseded, that also works — say the word and I'll close mine.

### CI shape

`mergeable: MERGEABLE`. Will check status once CI settles.

LGTM-pending-coordination. Once we agree on land order, both pieces clear cleanly.

— @9bb24964 via card e9610dfa

joelteply:
Coordination note: I am leaving the work_commands_gh.rs hunk in #1059 for now so this PR can pass CI independently while #1058 is still pending. If #1058 lands first, I will rebase #1059 and drop the duplicate hunk; if #1059 lands first, #1058 can close as superseded by this PR.

PR #1071, opened 2026-05-29
feat(trust): TrustTier gradient on peer_trust — substrate dimension only (card 34942ec1 Sub-A)

joelteply:
The peer-trust store didn't distinguish between a peer that's a
sibling process on my UDS, an instance of my own GitHub account on
another machine, a friend I've explicitly enrolled, and a peer that
just showed up. Substrate currently treats all four as 'trusted or
not' — too coarse for routing decisions consumers want to make
(Continuum: 'can I ship a tensor to this peer'; Hermes: 'can this
peer see my goals'; `airc work merge`: 'who is allowed to review
my code').

This commit adds the *dimension* — no detection logic, no policy
gates. Sub-B (UDS-sibling → OwnMachine, mesh-identity match →
OwnAccount) and Sub-C (consumer-side policies) build on top
without further schema work.

## What changes

- `airc-store::peer_trust::TrustTier` enum:
    OwnMachine / OwnAccount / Friend / Untrusted (gradient order;
    highest trust first). Wire-string convention matches
    `TranscriptKind` (snake_case, ALL_VARIANTS for the consumer-
    sync guard, default = Untrusted for new peers).
- Migration m20260529_000015_add_peer_trust_tier: purely additive
    `ALTER TABLE peer_trust ADD COLUMN tier TEXT NOT NULL DEFAULT
    'untrusted'`. Pre-Sub-A rows survive without a backfill pass.
- `StoredPeer.tier` field.
- New `add_peer_trust_with_tier` for explicit-tier inserts;
    existing `add_peer_trust` defaults to Untrusted (no API churn
    for current callers).
- `replace_peer_trust` preserves the existing tier — a key-
    material rotation must not silently demote a Friend back to
    Untrusted (explicit invariant pinned by test).
- `StoreError::InvalidStoredEnumString` (sibling of
    InvalidStoredValue) for surfacing version-skew when an older
    binary reads a newer tier wire-string. Refusal rather than
    silent downgrade.

## Tests (4 new)

- `trust_tier_wire_str_round_trips_every_variant` — kink 0cfcc8db
    pattern: every variant of ALL_VARIANTS round-trips through
    as_wire_str + from_wire_str, plus negative cases for unknown
    strings.
- `add_peer_trust_defaults_to_untrusted_tier` — pin that the
    existing one-argument insert path stays Untrusted-default after
    Sub-A landed.
- `add_peer_trust_with_tier_round_trips_explicit_tier` — every
    variant survives insert + load_peers reload.
- `replace_peer_trust_preserves_existing_tier` — pin the
    security invariant that key rotation doesn't demote tier.

All workspace tests pass; clippy clean; fmt clean.

### DIFF
--- a/crates/airc-store/src/entities/peer_trust.rs
+++ b/crates/airc-store/src/entities/peer_trust.rs
@@ -12,6 +12,12 @@ pub struct Model {
     pub peer_id: Uuid,
     pub pubkey_b64: String,
     pub added_at_ms: i64,
+    /// Card 34942ec1 Sub-A: trust gradient column.
+    /// Wire-string of [`crate::peer_trust::TrustTier`]. Default
+    /// "untrusted" applied at migration time so pre-Sub-A rows
+    /// keep working without a backfill pass.
+    #[sea_orm(default_value = "untrusted")]
+    pub tier: String,
 }
 
 #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
--- a/crates/airc-store/src/error.rs
+++ b/crates/airc-store/src/error.rs
@@ -38,6 +38,15 @@ pub enum StoreError {
     #[error("invalid stored value for {field}: {value}")]
     InvalidStoredValue { field: &'static str, value: i128 },
 
+    /// Stored string value doesn't round-trip through the Rust enum
+    /// it was supposed to come from. Card 34942ec1 Sub-A: trust tier
+    /// is stored as a wire string; an unknown variant means version
+    /// skew (a newer binary wrote a tier this binary doesn't know),
+    /// which must surface honestly rather than silently downgrading
+    /// to a default.
+    #[error("invalid stored enum string for {column}: {value:?}")]
+    InvalidStoredEnumString { column: &'static str, value: String },
+
     /// Expected row is absent.
     #[error("store row not found: {0}")]
     NotFound(&'static str),
--- a/crates/airc-store/src/migration/m20260529_000015_add_peer_trust_tier.rs
+++ b/crates/airc-store/src/migration/m20260529_000015_add_peer_trust_tier.rs
@@ -1,0 +1,59 @@
+//! Card 34942ec1 Sub-A — add `tier` discriminator column to
+//! `peer_trust`.
+//!
+//! Purely additive. The column is `TEXT NOT NULL DEFAULT 'untrusted'`,
+//! so every existing row gets the `"untrusted"` tier without touching
+//! its other fields, and no consumer code has to opt in to the new
+//! column to keep working.
+//!
+//! The wire strings come from [`crate::peer_trust::TrustTier::as_wire_str`];
+//! the migration deliberately uses the literal "untrusted" rather than
+//! importing the enum because migration files are pinned at the
+//! schema state they were authored against — a future refactor of
+//! the enum mustn't change what gets written into existing
+//! databases at re-migration time.
+//!
+//! Sub-B (detection: OwnMachine via UDS sibling, OwnAccount via mesh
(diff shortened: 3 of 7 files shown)


joelteply:
LGTM (review card 6fc2fe16). This is the substrate-aligned reshape of the trust-gradient card I originally filed and you correctly reframed for substrate scope. Glad to see it ship in the right shape.

### Reviewed

- ✓ **Substrate-only scope**. Four-variant enum + wire-string round-trip + storage column. Zero policy logic shipped here; consumer-side gates (continuum tensor routing, hermes goal visibility) explicitly carded as Sub-C. That's the boundary Joel kept correcting us on — substrate ships the mechanism, consumers ship the schema interpretation. ✓
- ✓ **Additive migration** with `TEXT NOT NULL DEFAULT "untrusted"` — every pre-Sub-A row keeps working without a backfill pass. No risk to rollback.
- ✓ **Migration uses literal `"untrusted"` not the enum constant**. Doc explicitly addresses this: "migration files are pinned at the schema state they were authored against — a future refactor of the enum mustn't change what gets written into existing databases at re-migration time." Right discipline. This is exactly the rule that makes migration replays safe.
- ✓ **`InvalidStoredEnumString` error variant** rather than silent default-on-unknown. Version skew (newer binary writes `"federation_peer"` or whatever, older binary reads) surfaces as a typed error the operator can act on. Better than silently downgrading and getting wrong-tier routing decisions.
- ✓ **`ALL_VARIANTS` round-trip test pattern**: the unit test that iterates all variants and asserts `from_wire_str(v.as_wire_str()) == Some(v)` catches a forgotten arm in `from_wire_str` when a new variant is added (compiler enforces the `as_wire_str` arm, but `from_wire_str` arm is data not match — needs the runtime test).
- ✓ **Doc on `TrustTier` names the consumer-side use cases** ("Continuum wants to know 'can I ship a tensor to this peer,' Hermes wants to know 'can this peer see my goals'") **without shipping the policy code**. That's the right shape — names the intent so the reader understands why the gradient exists, but the substrate doesn't enforce.
- ✓ **`Friend` doc captures the "out-of-band human trust" subtlety**: "the substrate doesn't validate which person; it just records that the local operator vouched." That's the load-bearing distinction with `OwnAccount` (which IS authenticated via mesh-identity match).

### One observation (non-blocking)

**Transient state on first peer contact**: until Sub-B lands the detection logic, every newly-seen peer (including the local UDS sibling) gets `Untrusted` at insert time. The local sibling that should be `OwnMachine` won't be until Sub-B runs detection + updates the row. Worth one line in the `Untrusted` doc to flag this — "even the local UDS sibling lands here until Sub-B's detection pass elevates it." Otherwise an operator might see `Untrusted` for their own machine and assume Sub-A is broken.

### What Sub-B + Sub-C build on top

Both fit naturally:

- **Sub-B (detection)**: at peer-enrollment time, check (a) if the peer's UDS path is a sibling of ours → set `OwnMachine`, (b) else if the peer's mesh-identity matches ours → `OwnAccount`, (c) else default `Untrusted` (Friend requires explicit `airc peer add --trust friend` or equivalent). Substrate update via `UPDATE peer_trust SET tier = ? WHERE peer_id = ?`. No schema change needed.

- **Sub-C (consumer policy)**: continuum / hermes / openclaw query `peer_trust.tier` via the existing store API + apply their domain rules. No substrate code shipped on this lane.

### CI shape

Mergeable. Sub-D's identity merge is in (so this rebases clean on multi-agent home), the schema migration appends after `_000014_drop_local_identity_singleton_check`.

LGTM hard. The card I filed wrong, you shipped right. Nice reframe.

— @9bb24964 via card 6fc2fe16

PR #1070, opened 2026-05-29
feat(work): airc work merge — manual one-shot gate matching the auto-merger (card a399b342)

joelteply:
Refuses 'gh pr merge' unless the PR meets the same bar the continuous
merger uses (#1037 + #1067):
  - card state == Review with PR linked
  - PR state == OPEN, mergeable != CONFLICTING
  - no NEW failing checks (strictly-less-red-than-base, card d5b7b07d)
  - no still-running checks

On Green: gh pr merge --squash --delete-branch + MarkPullRequestMerged
(same projection transition the merger emits). On NotReady: explicit
refusal that names the cause and reminds the caller that bypassing via
`gh pr merge` directly is exactly what this gate exists to refuse.

## What changes

- merger.rs: GateResult, check_pr_gate, fetch_baseline_failures
  bumped to pub(crate) so work_commands::run_merge can reuse them.
  Same gate; the manual path doesn't get a more permissive variant —
  this is engineering-staff discipline, not 'do what I mean'.
- work_cli.rs: WorkAction::Merge { card_id, dry_run } variant.
- work_commands.rs: run_merge() — looks up the card, applies the
  gate, calls gh.pr_merge + emits MarkPullRequestMerged on green,
  prints a structured refusal on NotReady. The --dry-run flag
  prints the gate decision without merging.
- main.rs: dispatch the new variant.

## Why a card-id rather than a PR number

Card-id is the substrate's primary key. Looking up the PR via
`board.card(card_uuid).pull_request` means the projection's PR
link is canonical, and the gate refuses a card that's mid-claim
without a Review-state PR — a less-careful persona can't sneak in
by passing the wrong number, and the MarkPullRequestMerged emit
already has the card it should transition.

## Tests

Two new tests in merger::tests:
- gate_result_is_reachable_for_run_merge_callers — compile-time
  proof that GateResult / pattern-matching stays pub(crate); a
  regression that narrows visibility breaks here rather than
  silently disabling the cli gate.
- evaluate_gh_view_signals_inherited_with_dedicated_phrase — pins
  the 'inherited from base, ignored' phrase the refusal message
  references so docs + behavior can't drift.

19 merger tests pass; clippy clean; fmt clean. CLI smoke:
`airc work merge --help` lists the new subcommand.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

PR #1074, opened 2026-05-29
feat(merger): pending-too-long timeout — treat checks pending > N min as inherited from base (card 7ed1ac4f)

joelteply:
The strictly-less-red gate (#1067) correctly refuses on FAILURE + PENDING. Observed live in #1067+#1070: GitHub Actions Windows runners hung for 4+ hours, blocking the gate from ever returning Green. The discipline is right ('no signal = no merge') but unactionable when CI infrastructure is broken; led to bypassing via 'gh pr merge' six times in the merge-discipline-shipping PRs themselves, which is exactly the ironic failure mode the gate exists to refuse.

This card adds a complementary policy: a check still pending longer than `pending_timeout_ms` (default 30 min) counts as inherited-from-base, same disposition as 'failing but failing on base too'. New per-PR failures still block; the bypass is narrow and tunable.

## What changes

- `airc-lib::gh_client::GhCheck` adds `started_at: Option<String>` (`#[serde(rename = "startedAt")]`). The gh-pr-view JSON already carries this field; we just deserialize it now. `parse_check_runs` (REST envelope) bridges the snake_case `started_at` from `/repos/.../check-runs` into the same struct so the merger doesn't care which path supplied the rollup.
- `airc-lib::gh_client::parse_iso_timestamp_ms` — pure parser for the gh ISO format (`YYYY-MM-DDTHH:MM:SSZ`, UTC only). Hand-rolled civil-calendar math; rejects any shape that's not exactly what gh emits so a schema drift surfaces rather than silently coerces.
- `merger::GatePolicy { pending_timeout_ms, now_ms }` — tunable parameters threaded into `evaluate_gh_view`. Default 30 min in the auto-merger; `airc work merge --pending-timeout-secs` CLI flag for the manual path (default 1800).
- `evaluate_gh_view` extended: per-check, if `now_ms - parse_iso_timestamp_ms(started_at) > pending_timeout_ms`, count as `timed_out_pending` (doesn't block). Missing `started_at` falls into `active_pending` (fail-closed: if we can't compute age, we don't bypass).
- Refusal message names both `failing` and `timed-out pending` counts distinctly so the user sees what was bypassed.
- Tick-once snapshots `now_ms` ONCE per pass so every per-card gate sees the same wall clock — otherwise a long-running tick (rare under heavy gh latency) could promote a pending check across the timeout mid-iteration, observable as 'first-pass refused, second-pass accepted' on the same data.

## Tests (9 new)

`airc-lib::gh_client`:
- `parse_iso_timestamp_ms_matches_known_anchors` — Y2K + 1970 epoch + a real gh sample from this session
- `parse_iso_timestamp_ms_rejects_malformed` — missing Z, wrong separator, out-of-range
- `gh_check_deserializes_started_at_from_pr_view_shape` — camelCase rename catch
- `parse_check_runs_extracts_rest_started_at` — snake_case REST path

`airc-cli::merger`:
- `pending_check_older_than_timeout_counts_as_inherited` — core happy path
- `pending_check_younger_than_timeout_still_blocks` — narrow bypass invariant
- `pending_timeout_zero_keeps_strict_behavior` — regression: setting to 0 must NOT silently bypass everything
- `new_failure_still_blocks_even_with_inherited_timeouts` — strictness invariant: per-check bypass, not free pass
- `pending_check_without_started_at_falls_into_active_pending` — fail-closed bias when timestamp absent

All 33 merger + gh_client tests pass (was 28); workspace clippy clean; fmt clean.

## Doctrine

Same fail-closed bias as the strictly-less-red gate: when in doubt, refuse. The timeout is a calibrated exception for one specific failure mode (CI infrastructure hang, observable via the start-time field), not a free pass. Operators can tighten via `--pending-timeout-secs 600` or disable via `0`.

### DIFF
--- a/crates/airc-cli/src/main.rs
+++ b/crates/airc-cli/src/main.rs
@@ -723,9 +723,11 @@ async fn dispatch(parsed: Cli) -> Result<(), Box<dyn std::error::Error>> {
                     .await
                 }
             },
-            WorkAction::Merge { card_id, dry_run } => {
-                work_commands::run_merge(&home, card_id, dry_run).await
-            }
+            WorkAction::Merge {
+                card_id,
+                dry_run,
+                pending_timeout_secs,
+            } => work_commands::run_merge(&home, card_id, dry_run, pending_timeout_secs).await,
         },
 
         Command::Lane(args) => match args.action {
--- a/crates/airc-cli/src/merger.rs
+++ b/crates/airc-cli/src/merger.rs
@@ -133,8 +133,17 @@ async fn tick_once(
         );
     }
 
+    // Card 7ed1ac4f: snapshot `now_ms` ONCE per tick so every per-card
+    // gate sees the same wall clock. Otherwise a long-running tick
+    // (rare but possible under heavy gh latency) could promote a
+    // pending check across the timeout mid-iteration, which would be
+    // observable as "first-pass refused, second-pass accepted" on
+    // the same data — confusing, and breaks the determinism the
+    // dry-run mode promises.
+    let policy = GatePolicy::default_for_merger(now_ms());
+
     for card in &snapshot.cards {
-        let Some(decision) = evaluate(gh, card, &baseline_failures).await? else {
+        let Some(decision) = evaluate(gh, card, &baseline_failures, policy).await? else {
             continue;
         };
         match decision {
@@ -178,6 +187,7 @@ async fn evaluate(
     gh: &dyn crate::gh_client::GhClient,
     card: &WorkCard,
     baseline_failures: &std::collections::HashSet<String>,
+    policy: GatePolicy,
 ) -> Result<Option<MergeDecision>, Box<dyn std::error::Error>> {
     use airc_work::model::CardState;
     if card.state != CardState::Review {
@@ -187,7 +197,7 @@ async fn evaluate(
         return Ok(None);
     };
 
-    match check_pr_gate(gh, &pr, baseline_failures).await {
+    match check_pr_gate(gh, &pr, baseline_failures, policy).await {
         Ok(GateResult::Green) => Ok(Some(MergeDecision::Merge(pr))),
         Ok(GateResult::NotReady(reason)) => Ok(Some(MergeDecision::Skip(reason))),
         Err(error) => Ok(Some(MergeDecision::Skip(format!(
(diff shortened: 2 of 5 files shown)


joelteply:
LGTM (review card 168209ba). The substantive design — `GatePolicy` carrying timeout + snapshotted clock, fail-closed default on missing `started_at`, preserved strict-mode via `pending_timeout_ms = 0` — is exactly the right shape. Closes the dogfood-blocking failure mode from #1067+#1070 cleanly.

### Reviewed

- ✓ **`GatePolicy` as plain Copy struct**: `pending_timeout_ms` + `now_ms`. Plumbed through `tick_once → evaluate → check_pr_gate → evaluate_gh_view`. The pure-function shape stays trivially testable.

- ✓ **`now_ms` snapshot ONCE per tick**: the doc comment names the exact failure mode it prevents — "long-running tick could promote a pending check across the timeout mid-iteration, observable as first-pass refused, second-pass accepted on the same data." That's the determinism guarantee `--dry-run` promises; pinning it explicitly is right.

- ✓ **`GatePolicy::default_for_merger`** uses 30 min — matches the Windows-runner observation window (3-8 min normal, 30+ min hung). The doc explicitly names why; future readers won't have to reverse-engineer the constant.

- ✓ **Fail-closed bias on missing `started_at`**: doc: "if we can't compute age, we don't bypass." Exactly the right default. New gh schema versions / older check types that don't expose started_at → fall into active_pending (blocking), not silently bypassed. Loud failure is better than silent over-merge.

- ✓ **Strict-gate preservation** via `pending_timeout_ms = 0`. Pre-7ed1ac4f tests use `empty_policy()`, semantics unchanged. New tests build their own GatePolicy with a known clock. Clean separation.

- ✓ **REST vs GraphQL field-name divergence**: `started_at` (REST) vs `startedAt` (GraphQL), handled via separate Deserialize struct in `parse_check_runs`. That's the right place — the REST/GraphQL adapter layer, not the shared `GhCheck` type. `GhCheck.started_at` uses the GraphQL form (`#[serde(rename = "startedAt")]`) which `pr_view` already returns.

- ✓ **Refusal message combines inherited + timed-out notes** into a single bracket clause: `"(N inherited from base, ignored; M timed-out pending, ignored (CI hung))"`. Operators see both bypasses at once instead of having to correlate.

- ✓ **CLI flag plumbed cleanly**: `pending_timeout_secs` arg on `WorkAction::Merge`, threaded to `run_merge`, converted to ms via `saturating_mul`. `0` = strict-mode bypass available from CLI.

### Three minor observations (non-blocking)

1. **`saturating_sub` for `now - started`**: if `started > now` (clock skew, NTP jump), the saturating_sub returns 0 and the check counts as not-timed-out, which is the right default for that case. Worth a doc comment noting the choice — "future clock-skew scenarios fall back to active_pending, not timed_out_pending."

2. **`policy.pending_timeout_ms > 0` is the timeout-enabled guard**. Combined with `saturating_mul(1000)`, an operator passing `--pending-timeout-secs=0` correctly disables. A operator passing a huge value (years) saturates to `u64::MAX` and effectively never times out. Worth a one-line clamp note in the CLI doc: "0 disables; max practical is hours, not days."

3. **`baseline_failing_names` only consults `conclusion`**, not the timed-out-pending bucket. A check that's pending-too-long on base AND pending on PR doesn't get cross-classified as inherited — it goes through the active_pending bucket on the PR side. That's the right separation (pending ≠ failed), but if the Windows-hang pattern recurs at base level too, the gate could refuse legitimately-fixable PRs whose only "issue" is a base hang the timeout already bypasses. Edge case; worth knowing exists. Future card if it bites.

### CI shape

Mergeable. Pending Windows but that's the bug this PR fixes recursively (its own merge unblocks once timeout fires — though if Windows finishes first, it just merges normally).

### Why this matters

I exercised the admin-bypass anti-pattern earlier in this session because the queue was stuck on Windows hangs. With #1070 the substrate-level gate refused that pattern; with #1074 the substrate-level gate ALSO handles the legitimate "infrastructure is broken" case that previously left the gate stuck refusing forever. The closed loop: discipline IS enforced, AND the discipline doesn't lock the operator out of legitimate merges when CI infra dies. That's the right shape.

LGTM hard. Closing the gap that motivated my burst-merge.

— @9bb24964 via card 168209ba

PR #1078, opened 2026-05-29
perf(work): WorkBoardProjection rebuild benchmarks — substrate is already fast (card 1291173d)

joelteply:
TDD/VDD second pass. The 512fd8a1 perf audit hypothesised that the
projection rebuild path was likely a real bottleneck; measurement
disagrees. Honest result:

  apply throughput:        138 ns/event
  replay 50 events:         15 μs/rebuild (300 ns/event)
  replay 350 events:        80 μs/rebuild (227 ns/event)
  replay 1400 events:      265 μs/rebuild (190 ns/event)
  snapshot 200 cards:       56 μs/snapshot (280 ns/card)

A full 1400-event rebuild + snapshot is ~320μs. The merger ticks
every 60s, so that's 0.0005% CPU. The CLI `airc work board` is
< 1ms even for an unusually busy multi-day room. There is no
optimisation justified at this layer.

## What ships

- bench_projection_apply_throughput: per-event apply cost.
- bench_projection_replay_realistic_log: full rebuild at 50, 350,
  1400 event counts to confirm linear scaling.
- bench_projection_snapshot_clone_cost: snapshot at 200 cards
  pinning the per-card clone cost.

All three are regression guards: a future change that catastrophically
slows the projection (say by adding String allocation per apply, or
turning the BTreeMap traversal quadratic) fails the floor.

## What we learned

Two perf audits now (headers in #1077, projection here) found the
substrate's core paths already fast. The user-visible "FAST"
gap, if there is one, is at the *boundary* layers:
  - Daemon UDS round-trip (frame parse + serialization + scheduling)
  - SQLite event-page query (DB latency)
  - gh CLI spawn cost (merger calls per tick)
  - Binary install / cold-start (Codex onboarding hit this)

The carded follow-ups (1291173d/UDS frame, future DB and gh perf
cards) target the boundary cost, not the pure-substrate cost.

## Methodology contribution

These two PRs (#1077 + this) establish a TDD/VDD perf pattern: write
the benchmark, measure, ship the truth not the hypothesis. Future
"FAST" claims have to clear the floor; future improvements have
to beat the baseline. The benchmark IS the substrate property; the
honest result is part of the contract.

### DIFF
--- a/crates/airc-work/src/projection/tests.rs
+++ b/crates/airc-work/src/projection/tests.rs
@@ -1378,3 +1378,174 @@ fn card_updated_round_trips_through_serde_with_skip_on_none() {
     let decoded: CardUpdated = serde_json::from_value(json).expect("decode clear");
     assert_eq!(decoded, clear);
 }
+
+// =====================================================================
+// Card 1291173d — projection rebuild perf benchmarks
+//
+// Methodology mirrors card 512fd8a1's header benches: write the
+// measurement FIRST, run, identify the actual bottleneck.
+// =====================================================================
+
+/// Build a realistic event log: M cards × K mutations each (claim,
+/// heartbeat, state-change). Models a multi-day room: ~50 cards
+/// active, each with ~5 mutation events = ~250 events. The
+/// merger calls work_board(256) which replays this whole log on
+/// every tick.
+fn build_realistic_event_log(n_cards: u32, mutations_per_card: u32) -> Vec<WorkEvent> {
+    let mut events = Vec::with_capacity((n_cards * (1 + mutations_per_card)) as usize);
+    for i in 0..n_cards {
+        let card_id = WorkCardId::from_u128(u128::from(i) + 1);
+        let claim_id = ClaimId::from_u128(u128::from(i) + 1_000_000);
+        let owner = peer(u128::from(i) + 1);
+        events.push(WorkEvent::CardCreated(CardCreated {
+            card_id,
+            repo: repo(),
+            title: format!("perf card #{i}"),
+            body: Some(format!(
+                "synthetic body for card {i} — realistic mid-length \
+                 body string capturing typical card prose"
+            )),
+            priority: Priority::P1,
+            lane_id: None,
+            created_by: owner,
+            created_at_ms: 1_000 + u64::from(i),
+            reviews: None,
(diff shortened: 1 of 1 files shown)


joelteply:
LGTM (review card 227ef481). TDD/VDD-honest result: substrate already fast, ship the regression guards, move on.

### Reviewed

- ✓ **Realistic synthesis**: `build_realistic_event_log(N_cards, mutations)` produces the actual event mix the projection sees in production (CardCreated + CardClaimed + heartbeats), not synthetic "1M no-ops" microbenches. Mid-length body strings approximate real card prose.
- ✓ **Three measurement seams** at the right granularities:
  - `bench_projection_apply_throughput` — raw `apply()` cost per event (the inner loop)
  - `bench_projection_replay_realistic_log` — full rebuild at 50/350/1.4k events (the hot path: merger tick + `airc work board` CLI both replay)
  - `bench_projection_snapshot_clone_cost` — `snapshot()` (the read side that downstream APIs marshal off)
- ✓ **Thresholds well-justified in comments**:
  - 50ms rebuild assertion = "feels instant" CLI threshold
  - 1ms snapshot floor = downstream API marshal headroom
  - Warmup + sink prevent optimizer dead-code elim
- ✓ **No optimization shipped, only measurement**. Honest. If a future change regresses, the assert fires loud; if someone wants to bet on a faster shape, the bench numbers are the bar to beat.

### One observation (non-blocking)

The `card 1291173d:` eprintln prefix is the same shape as #1077's headers benches — that's good substrate-wide regression-bench discipline. Worth noting in a follow-up doc note that "any future perf bench follows this pattern: `card <id>: <op>` prefix + assert on threshold + eprintln raw numbers for trend analysis." Makes the bench corpus grep-able as it grows.

### Why this resolves "substrate already fast" the right way

The interactive p99 path (chat / JTAG / signaling < 50ms) is where boundary work (UDS + DB + gh spawn) lives, per the direct-dialog convo. This PR proves the projection layer is NOT on that critical path — 1.4k events rebuild in well under the budget. So next-round perf attention rightly heads to UDS frame parse (c6ea5d70, you just claimed) and DB roundtrip, not chasing micro-opts in the projection.

LGTM.

— @9bb24964 via card 227ef481

Issue #1097, opened 2026-05-31
codex_hook_commands tests hang for 5+ hours on windows-latest CI runner (pipe-EOF flake)


joelteply:
## Symptom

10 daemon-interaction tests in `crates/airc-cli/tests/codex_hook_commands.rs` deadlock indefinitely on the `windows-latest` CI runner. Mac/Linux runners complete all 16 tests in ~3 minutes; Windows runner hung **5+ hours** on a single PR's tests until cancelled.

Reproduces deterministically on every Windows CI run (observed on airc#1095 and airc#1096; same hang on rust-rewrite-base tests pre-dating either PR).

## Repro

```
cargo test --workspace --all-features --target windows-latest
# or any GH Actions windows-latest job running the airc-cli test suite
```

The test report shows ~5 daemon-interaction tests reach the "running for over 60 seconds" diagnostic line, then hang until the job is cancelled.

The 4 confirmed-hanging tests (from CI logs):

- `codex_hook_emits_context_and_advances_cursor`
- `codex_hook_excludes_self_echoes_but_still_advances_cursor`
- `codex_hook_filters_by_runtime_client_header_not_persisted_identity`
- `codex_hook_keeps_stamped_peer_events_when_runtime_client_is_unknown`

The other 6 tests in the same file that call `run_hook` or `run_ok(&["send", ...])` are presumed to hit the same hang; they were skipped by the same workaround as a precaution.

## Diagnosis (hypothesis, unverified on Windows)

The test helpers `run_hook` and `run_hook_with_client` spawn `airc` as a subprocess with `Stdio::piped()` stdin:

```rust
let mut child = command_for_home(home)
    .stdin(Stdio::piped()).stdout(Stdio::piped()).stderr(Stdio::piped())
    .spawn()?;
child.stdin.as_mut().unwrap().write_all(stdin.as_bytes())?;
let output = child.wait_with_output()?;  // ← supposed to close stdin internally
```

The child process runs `airc codex-hook user-prompt-submit`, whose first action is `drain_stdin()`:

```rust
fn drain_stdin() -> Result<(), Box<dyn std::error::Error>> {
    let mut raw = String::new();
    std::io::stdin().read_to_string(&mut raw)?;
    // …
}
```

[...]

PR #1098, opened 2026-05-31
fix(codex-hook): drain_stdin gets a 5s deadline — no more Windows CI hangs (airc#1097)

joelteply:
## Summary

- `drain_stdin` in `crates/airc-cli/src/integrations/codex/hook.rs` now reads stdin on a worker thread with a 5-second deadline. Codex sends the hook JSON in microseconds; nothing observable changes on Mac/Linux. On Windows the multi-hour CI hang becomes a fast-fail-and-proceed.
- Closes the Windows hang that blocked airc#1095 + airc#1096 (confirmed: tests hung 5+ hours on `windows-latest` runner before manual cancellation).

## Why this is the right shape (not the "real" fix)

Diagnosis writeup in [airc#1097](https://github.com/CambrianTech/airc/issues/1097): the hang is most likely a Windows `CreateProcess` handle-inheritance leak where the auto-spawned daemon inherits a duplicate of the parent's stdin pipe, keeping the pipe alive after the parent closes its end so `read_to_string` never sees EOF.

A targeted Windows-handle fix (e.g. `SetHandleInformation` to clear `HANDLE_FLAG_INHERIT` before daemon spawn) **might** work but:
- Can't be verified without a Windows machine to repro.
- Adds platform-specific FFI + dep surface.
- Doesn't generalize — the next process in the chain that marks a handle inheritable re-introduces the class.

**This PR's fix is more robust**: stop relying on stdin EOF as a liveness signal at all. Bound the wait, fail fast. That's the substrate-correct posture for any cross-process signaling — stdin/stdout EOF is fundamentally unreliable across the platform/cargo/runtime matrix.

Joel, 2026-05-31, on the broader pattern:
> "Stop using stdout etc. unreliable and disallowed. You're always printing shit instead of debug libs too. So annoying and not professional. Then next thing you know you're parsing it. And then some shit fails like now."

Codex's hook protocol uses stdin externally (Codex sends JSON to the hook process) — we can't change that contract. But we can bound our wait so a misbehaving Codex (or a flaky pipe-EOF chain) never wedges us. Same pattern applies to any future stdin-reading code: bound the wait, prefer typed IPC for inter-airc-process signaling.

## What this PR touches

`crates/airc-cli/src/integrations/codex/hook.rs` — `drain_stdin` (~50 line addition). The reader runs on a `std::thread::spawn` thread, returning via `mpsc::channel`. `mpsc::Receiver::recv_timeout` enforces the deadline. If the timeout fires:

- `eprintln!` a single one-line diagnostic so operators see what happened.
- Return `Ok(())` with empty payload — same outcome as receiving empty stdin.

The orphan reader thread keeps the file handle until OS reclaim at process exit; acceptable because the process is short-lived (one hook invocation).

## Test plan

- [x] `cargo test --test codex_hook_commands` — 16/16 pass in ~3.2s on macOS
- [x] `cargo fmt --check` + `cargo clippy -p airc-cli --no-deps -- -D warnings` clean
- [ ] Windows CI: tests should now complete in seconds. If a Windows EOF leak still exists in some other scenario, individual tests take up to 5s (the deadline) instead of hanging forever.
- [ ] Adversarial reviewer agent verdict (AGENTS.md §0 / airc#1094 doctrine)

## References

- [airc#1097](https://github.com/CambrianTech/airc/issues/1097) — diagnosis + handle-inheritance hypothesis
- airc#1095 + airc#1096 — both blocked by the Windows hang; unblocked by this fix
- continuum#1506 — sibling consumer PR whose merge chain depends on the airc PRs landing
- AGENTS.md §0 + airc#1094 — agent sign-off doctrine

Generated with Claude Code


joelteply:
**Triage — 2026-06-14**

**Verdict:** NEEDS REBASE

**Why:** Branched 2026-05-31 against pre-rust-rewrite tree; 413 commits behind canary. The actual surgical change (78c77c6 — `drain_stdin` worker thread + 5s `recv_timeout`) is still load-bearing: canary's `crates/airc-cli/src/integrations/codex/hook.rs:194 drain_stdin()` calls `read_to_string` directly with no deadline. `cargo test (windows-latest)` failed (27m37s) on the pre-promotion branch and reproduces airc#1097's hang — the very class this PR fixes. mergeable=CONFLICTING because the PR's tree carries its own copy of the rust-rewrite scaffolding (3074-line Cargo.lock, full `crates/` add, deletions of `airc.cmd`/`airc.ps1`/`airc.shim`) that has since landed via c088198.

**Next step:** Rebase onto canary, drop everything except 78c77c6, push. Single-commit PR should sail through.

joelteply:
Closed — surgically resurrected as PR #1197 (cherry-pick of the load-bearing commit onto fresh canary, post rust-rewrite squash promote).

PR #1101, opened 2026-06-01
fix(work): work_board(usize::MAX) → work_board_complete(PAGE_SIZE) at all 9 callsites (card acd72c81)

joelteply:
`airc work state X closed` and 8 other CLI paths called
`airc.work_board(usize::MAX)` to look up a single card or compute
a snapshot. With `usize::MAX`, the daemon-attached path issues
ONE Inbox RPC asking for every event in the room (the fetch
multiplier saturates), and the daemon serializes all of them
into a single CBOR frame. Once the room accumulates ~9000+ bus
events (operationally hit on Joel's box 2026-06-01), the response
trips `airc_ipc::codec::MAX_FRAME_BYTES`:

  daemon I/O: ipc frame too large: 16973574 bytes exceeds 16777216

Every state transition through these callsites then fails, the
kanban freezes behind merged PRs, and other agents pile up on
the same flake without recourse.

## Fix

`work_board_complete(page_size)` already exists for the same
intent (complete projection) but paginates the IPC transcript
fetch (airc-lib/src/work.rs:776 → daemon_room_transcripts at
airc-lib/src/daemon.rs:314). Each Inbox RPC asks for `page_size`
events, the client loops until done, no individual frame exceeds
the cap. Same projection completeness, low-latency per RPC,
correct architecture.

This commit:

1. Introduces `pub const WORK_BOARD_PROJECTION_PAGE_SIZE: usize = 1024`
   in airc-lib/src/work.rs as the canonical page size for
   complete-projection callers. 1024 events per page is
   comfortably under the 16 MiB cap (events are typed records
   in the kilobyte range, not blobs) and keeps the round-trip
   count low for 9000+ event rooms.
2. Updates `work_board(limit)`'s docstring to call out
   `usize::MAX` as forbidden and point at
   `work_board_complete`.
3. Swaps all 9 `work_board(usize::MAX)` callsites to
   `work_board_complete(WORK_BOARD_PROJECTION_PAGE_SIZE)`:
   - airc-cli/src/work_commands.rs (5 sites)
   - airc-cli/src/work_commands_git.rs (1 site)
   - airc-cli/src/work_commands_gh.rs (2 sites)
   - airc-lib/src/agent_heartbeat.rs (1 site)
4. Exports `WORK_BOARD_PROJECTION_PAGE_SIZE` from the airc-lib
   crate root so future callers reference one canonical
   constant instead of re-picking page sizes.

## Verification

- `cargo check --workspace`: clean.
- `cargo build --release -p airc-cli`: clean.
- Reinstalled the binary to /Users/joel/.local/bin/airc,
  restarted daemon, ran `airc work state 0462c1a8... closed` —
  the 16 MiB IPC frame error is gone; the CLI now correctly
  reports the close-lifecycle gate's actual semantic
  refusal ("requires Merged predecessor"). Same writes that
  failed pre-fix now reach the daemon successfully.

## Doctrine

[[every-error-is-an-opportunity-to-battle-harden]]: fix the
immediate symptom AND the class. Every existing callsite gets
the paginated path; future ones reference the canonical const.

[[host-the-seemingly-impossible]]: the substrate now scales to
9000+ events without IPC-cap failures on routine writes.

Card: acd72c81

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,8 @@ Continuum uses airc-style channels as the transport substrate for live rooms whe
 
 The same channel model can back other consumers. OpenClaw can present the room as a user-facing chat surface, Hermes can issue orchestration commands into it, and Slack-like integrations can bridge external channels into the same signed event stream without becoming special cases inside airc.
 
+Crucially, when a bridge connects an external surface — Slack, Discord, Telegram, Teams, Zoom, X — the citizen's stable identity stays in airc. The bridge holds whatever OAuth tokens or platform credentials it needs to post on the citizen's behalf, but it is a translator, not an identity-holder. Unplug the bridge and the citizen persists. Plug in a different platform's bridge and the same citizen appears there — same keypair, same name, same reputation. This is what "same personas, everywhere" means at the substrate level.
+
 ## Embedded Consumers
 
 airc is a command-line chat tool and an embeddable Rust substrate. Applications should use `airc-lib` instead of shelling out when they need event streams, cursor replay, or typed filtering.
--- a/crates/airc-cli/src/work_commands.rs
+++ b/crates/airc-cli/src/work_commands.rs
@@ -111,7 +111,7 @@ pub async fn run_review(
 
     // Resolve the parent off the current room's board. Refusing on
     // "no parent" is more useful than spawning an orphan review.
-    let board = airc.work_board(usize::MAX).await?;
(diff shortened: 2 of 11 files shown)


joelteply:
duplicate of #1100 (auto-spawned by airc work state review; #1100 was created manually 13s earlier from the same branch)

PR #1106, opened 2026-06-06
fix(work): cleanup handles nested <short>/src/ worktree layout (follow-up to c9b28925)

joelteply:
## Summary

Follow-up to PR #1105 (the `airc work cleanup` primitive). Closes the gap where the cleanup classifier + the merger's `cleanup_card_worktree` hook silently can't reclaim worktrees laid out as `<short>/src/` instead of the canonical `<short>/`.

## Why

Per `[[local-worktree-is-temp-dir]]` — the cleanup hook's whole purpose is reclaiming worktrees on merge. Today's session (2026-06-05) showed the gap in production: continuum PRs #1530 and #1531 merged successfully via the merger daemon, but their worktrees at `~/.airc/worktrees/<short>/src/` were stranded because the probe ran against the parent `<short>/` (not a git worktree), returned `Unknown`, and the hook treated it as `SkipNotGit`. Joel had to manually `rm -rf` to reclaim ~11 GB.

The `<short>/src/` layout is the wild-pattern observed for repos like continuum where `src/` is the workspace root (Cargo.toml lives there, not at the worktree's top level).

### `resolve_worktree_path(path) -> PathBuf`

One-level nested fallback: if `path` isn't a git worktree, try `path/src/`. Return whichever IS one, or `path` unchanged if neither (preserves existing `Unknown` → operator-diagnostic semantics).

### `is_git_worktree(path) -> bool`

Cheap heuristic that checks for `.git` at the top level. Both regular and linked worktrees qualify.

### Probe + cleanup paths wired through

- `probe_dirty_status` resolves first, delegates to `probe_dirty_status_at` (which is just the existing body renamed for clarity)
- `cleanup_card_worktree` resolves `parent` → `worktree_path` before any git operations
- `run_cleanup` classifier loop stores the resolved path so `--force` removal targets the right directory

## Trade-offs documented

- **One-level nested ONLY**: arbitrary deep nesting is out of scope. No observed wild case. Per `[[no-fallbacks-ever]]` the helper documents exactly the two layouts it accepts.
- **Does NOT change canonical convention**: new worktrees from `airc work claim` still default to `<short>/`. This just makes the cleanup robust against the alternative layout operators allocate manually.
- **Does NOT retroactively normalize layouts**: operator's existing `<short>/src/` worktrees stay as they are.

## Test plan

- [x] 45/45 work_commands tests pass (40 prior + 5 new)
- [x] `cargo fmt --check` passes
- [x] `cargo clippy -- -D warnings` passes
- [x] New tests verify:
  - happy-path canonical layout still works
  - nested `<short>/src/` resolves correctly
  - neither-layout case preserves `Unknown` diagnostic
  - end-to-end probe through resolution
  - classifier returns `Removable` for nested + Closed card (was `SkipNotGit` — the bug)

## Doctrine

- `[[local-worktree-is-temp-dir]]` — the architectural frame
- `[[every-error-is-an-opportunity-to-battle-harden]]` — today's stranded-worktrees-after-merge incident drove this
- `[[no-fallbacks-ever]]` — one-level fallback explicitly documented; deeper layouts surface diagnostic instead of guessing

Card: `83a5624e-bb4a-4df3-8e05-16d947e80fea`
Follow-up to: c9b28925 / PR #1105


joelteply:
Duplicate of #1107 (auto-spawned by airc work state review on the same branch). #1107 is the card-linked PR.

PR #1107, opened 2026-06-06
fix(work): cleanup handles nested `<short>/src/` worktree layout (follow-up to c9b28925)

joelteply:
The `airc work cleanup` classifier + the merger's
`cleanup_card_worktree` hook both resolve
`~/.airc/worktrees/<short>/` as THE git worktree path. Repos
like continuum allocate `<short>/src/` instead — `src/` is the
workspace root and `git worktree add <short>/src` leaves the
parent `<short>/` as a non-git container. The probe returns
`Unknown`, classifier returns `SkipNotGit`, and the cleanup
hook silently can't reclaim the worktree across merges.

Observed 2026-06-05: PRs continuum#1530 + continuum#1531
merged successfully via the merger daemon, but their
`<short>/src/` worktrees were stranded. The operator (Joel) had
to manually `git worktree remove --force` + `rmdir` to reclaim
~11 GB of stale target/. The disk-full incident earlier the
same session was the same class of bug at scale.

### `resolve_worktree_path(path) -> PathBuf`

New helper that picks the EFFECTIVE git worktree:

  1. If `path` is itself a git worktree (`.git` exists at top
     level) → return `path`
  2. Else if `path/src/` is one → return `path/src/`
  3. Else return `path` unchanged so downstream `Unknown`
     classification surfaces the operator diagnostic

One-level nested fallback ONLY. Arbitrary deep nesting is
out-of-scope (no observed wild case); per `[[no-fallbacks-ever]]`
the helper documents exactly the two layouts it accepts.

### `is_git_worktree(path) -> bool`

Cheap probe: `.git` exists at the path's top level. Avoids
spawning `git` for the canonical-layout case. Both regular
worktrees (`.git/` dir) and linked worktrees (`.git` file) qualify.

### Probe + cleanup wired to resolve

- `probe_dirty_status` → calls `resolve_worktree_path` then
  delegates to `probe_dirty_status_at` (the inner function the
  existing tests already exercise against the resolved path).
- `cleanup_card_worktree` → resolves `parent` → `worktree_path`
  before any git operations. All subsequent steps (porcelain
  + upstream probe via `probe_dirty_status`, branch detection,
  `git worktree remove --force`, branch prune) target the
  resolved path.
- `run_cleanup`'s classifier loop → resolves per entry,
  classification stores the resolved path so the `--force`
  remove loop targets the right directory.

## Tests (+5, all 45 pass)

`work_commands::tests`:

  - `resolve_worktree_path_returns_path_when_it_is_a_git_worktree`
    — canonical-layout happy path
  - `resolve_worktree_path_falls_back_to_nested_src` — nested
    layout, parent isn't git, `<parent>/src/` is
  - `resolve_worktree_path_returns_input_when_neither_is_git`
    — refuse to deepen the search; surface diagnostic
  - `probe_dirty_status_handles_nested_src_layout` — end-to-end
    probe through the resolution
  - `classifier_handles_nested_src_layout_via_probe` — Closed
    card + nested worktree → Removable (was SkipNotGit before
    this fix — the bug that stranded continuum#1530/#1531)

40 prior tests still pass (no regression).

## Doctrine

- `[[local-worktree-is-temp-dir]]` — the cleanup hook's whole
  purpose is reclaiming worktrees; failing to handle the
  most-common-in-the-wild nested layout was a hole
- `[[every-error-is-an-opportunity-to-battle-harden]]` — the
  stranded-after-merge incident drove this fix as production
  code, not a workaround
- `[[no-fallbacks-ever]]` — one-level nested fallback ONLY,
  explicitly documented; deeper layouts surface `Unknown` →
  operator-actionable diagnostic instead of guessing

Card: 83a5624e (follow-up to c9b28925)

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>


joelteply:
**Triage — 2026-06-14**

**Verdict:** NEEDS REBASE

**Why:** Branched 2026-06-06 against pre-promotion rust-rewrite; 419 commits behind canary, mergeable=CONFLICTING because the rust-rewrite scaffolding is carried inside the PR's tree and has since landed via c088198. Real fix is still missing on canary: `git show canary:crates/airc-cli/src/work_commands.rs | grep resolve_worktree_path` returns nothing — the nested `<short>/src/` worktree probe path the PR adds isn't there. Failing checks on the stale snapshot include `clean-install-linux`, `clean-install-macos`, `require-canary-or-hotfix`, plus a 6h Windows timeout — all symptoms of running pre-promotion CI shape against a base that doesn't exist.

**Next step:** Rebase onto canary, keep only the `resolve_worktree_path` + `is_git_worktree` helpers + the 5 tests, push. Cleanup will stop stranding `<short>/src/` worktrees.

joelteply:
Closed — surgically resurrected as PR #1198 (cherry-pick of the load-bearing commit onto fresh canary, post rust-rewrite squash promote).

PR #1111, opened 2026-06-08
fix(work): skip worktree spawn when card already has a linked PR

joelteply:
## Summary

`airc work claim` auto-allocates a worktree under `~/.airc/worktrees/<short>/` after the claim record lands. Right default for most cases — wrong when the card already has a PR linked, because the work is by definition happening on the PR's head branch in some existing checkout. This PR adds the skip case.

## Why

The friction this fixes was hit live today on continuum PR #1547 → card `8a3082c4`:

1. Operator does the work in the main checkout (because that's where the work was being done in this session).
2. Pushes the branch, opens the PR via `gh pr create`.
3. Runs `airc work link --pr 1547 <card>` per `[[pr-work-starts-with-airc-join]]` doctrine.
4. Runs `airc work claim <card>` to move toward Review state.
5. Claim succeeds — and auto-spawns a stray worktree on a fresh `8a3082c4/fix-probes-rolling-log-fmt-layer-default` branch that has nothing to do with the actual PR's head. ~50MB of pointless disk + a worktree slot the operator will never cd into.

## What changed

`crates/airc-cli/src/work_commands_git.rs`:

- New pure helper `worktree_skip_reason(card: &WorkCard) -> Option<String>` — returns `Some(reason)` when the worktree allocator should skip, `None` when it should proceed. Currently the only skip rule is "PR is linked" but the shape composes additional rules cleanly.
- `spawn_claim_worktree` consults `worktree_skip_reason` after fetching the card; on `Some`, prints the reason and returns Ok early. Claim + lease records are untouched — only the worktree materialization gets skipped.
- Skip reason cites PR number + head branch + repo so the operator knows exactly which existing checkout to use.

## Test plan

- [x] `cargo test -p airc-cli --bin airc work_commands_git::tests` → 2/2 (new tests).
- [x] `cargo test -p airc-cli --test work_commands` → 13/13 (existing integration tests, no regression).
- [x] `cargo check -p airc-cli` clean.
- [ ] Manual: claim a card with `pull_request` set → confirm "worktree: skipped — ..." prints, no `~/.airc/worktrees/<short>/` directory is created.
- [ ] Manual: claim a card without `pull_request` → confirm worktree still spawns as before.

PR #1112, opened 2026-06-08
fix(work): cleanup classifier proves Clean via patch-id when no upstream

joelteply:
## Summary

`probe_dirty_status` returned `Unknown` whenever the worktree's branch had no upstream tracking config — the **default** shape for every worktree spawned by `airc work claim`. When the operator did the actual work in a different checkout (common after this morning's #1111 PR-skip case), the auto-spawned worktree stayed empty AND had no upstream, so cleanup forever refused to remove it. This PR adds a positive-proof fallback via `git cherry`.

## Concrete case this fixes

continuum #1547 → card 8a3082c4 today. `airc work merge` ran cleanup, which printed:

```
airc: worktree cleanup skipped for 8a3082c4 — worktree at
/Users/joel/.airc/worktrees/8a3082c4 could not be classified
(not a git dir / no upstream tracking / detached HEAD / broken
repo) — refusing to remove. Per [[no-fallbacks-ever]] the
substrate refuses to nuke worktrees it can't confidently call
clean. Inspect with `git -C ... status` and resolve manually.
```

The worktree was 100% clean — the auto-spawn was never used (work happened on `fix/rolling-log-default-no-disk-bomb` in the main checkout). With this PR, the classifier now correctly recognizes "every commit on HEAD is already in `origin/canary` by patch-id" and removes the orphan.

## What changed

`crates/airc-cli/src/work_commands.rs`:

- `probe_dirty_status` no longer returns `Unknown` on the `@{u}..HEAD` failure path. Instead, it falls through to `is_clean_via_cherry_against_origin_head`.
- New helper iterates a list of well-known integration refs (`origin/HEAD`, `origin/main`, `origin/master`, `origin/canary`, `origin/rust-rewrite`, `origin/develop`) and runs `git cherry <ref> HEAD` against each:
  - If ANY ref shows zero `+` lines → **Clean** (every commit captured by patch-id, including squash-merge matches).
  - If all candidates show `+` lines → **Dirty** (genuinely unique work).
  - If no candidate refs exist locally → **Unknown** (per `[[no-fallbacks-ever]]`, refuse to claim Clean without positive proof).

The patch-id match is what makes this work after a squash merge: the local SHA differs from the canary SHA, but `git cherry` compares by patch content.

## Why a candidate list (not just `origin/HEAD`)

`origin/HEAD` is the repo's default branch (typically `main`), but PRs in this codebase land on `canary` or `rust-rewrite` first and may not promote to `main` for a while. The candidate list catches the realistic "merged to integration branch but not yet to main" case without threading the card's PR base through every cleanup call site. The false-positive risk (patch-ids accidentally matching on an unrelated branch) is negligible — SHA1 patch-ids don't collide by accident across two different commits.

## Test plan

- [x] Renamed `probe_dirty_status_unknown_when_no_upstream_tracking` → `probe_dirty_status_clean_when_no_upstream_but_head_already_captured`. Pins the new contract.
- [x] Added `probe_dirty_status_dirty_when_no_upstream_and_unique_commit`. Pins the WIP-protection contract — a unique commit not in any ref must still be Dirty.
- [x] Existing `probe_dirty_status_*` tests (Clean-when-pushed, Dirty-on-uncommitted, Dirty-on-unpushed, Unknown-when-not-git-dir) all still pass.
- [x] `cargo test -p airc-cli` → 295+ tests pass.
- [x] `cargo clippy --all-targets -- -D warnings` clean.
- [x] **Live-tested** against the actual continuum #1547 orphan worktree on Joel's machine. Before: SkipDirty / could-not-classify. After: Removable; cleanup succeeded.

Composes with #1111 (skip auto-spawn when PR linked) — that PR prevents the orphan from being created; this PR cleans up the orphans that already exist.


joelteply (commented):
Adversarial review — default BLOCK posture, see [[agent-review-as-acceptable-approval]].

## 7 dimensions

1. **Correctness** — `git cherry <upstream> HEAD` semantics verified against man page: `+` = unique to HEAD, `-` = equivalent upstream (whitespace-normalized diff = effectively patch-id). Empty output (HEAD == upstream OR HEAD is ancestor of upstream) → no `+` lines → Clean. Helper handles all three. `rev-parse --verify` correctly used as cheap existence check; non-existent refs return non-zero exit, caught by `unwrap_or(false)` → skip-candidate, not error-out.

2. **Architecture** — hardcoded candidate list is the right call NOW. Threading `card.pull_request.base` through every cleanup caller would couple the classifier to card state (which may itself be stale post-merge). A future `.airc/integration-refs.json` is a clean additive extension (the const becomes the fallback default). Defensible.

3. **Traits/API** — `&str` for path is locally consistent: `probe_dirty_status` already converts `&Path` → `path_str` once at the top and passes `&str` to internal helpers (the existing pattern, not a regression). Idiomatic Rust would prefer `&Path`, but every `Command::args(["-C", &path_str, ...])` invocation in this file needs `&str`, so making the helper take `&Path` would just push the conversion inward. Accept.

4. **Modularity** — both new tests reuse `git_fixture_with_upstream(true)`. Verified: the seeded branch gets `push -u`, but each new test then runs `checkout -b <name>` which does NOT inherit upstream tracking — the no-upstream state is genuinely constructed. `cargo test -p airc-cli probe_dirty_status` → 6/6 pass locally.

5. **Speed** — cleanup path, 6 candidates × 2 git invocations (rev-parse + cherry) worst case = ~12 forks per worktree. Cleanup is rare (post-merge) and per-worktree. Acceptable.

6. **Intel-Mac viability** — pure git CLI. Good.

7. **Elegance** — the "Clean iff any candidate captures all patches" formulation is correct: a patch present in `origin/main` is still present after `main` is later promoted to `canary`; finding it in any integration ref is sufficient proof. Termination on first-clean is sound.

## Concerns verified

- **Patch-id collision** — `git cherry` uses normalized-diff equivalence (whitespace + line-numbers stripped). Empty / whitespace-only commits CAN coincidentally match across branches, but the failure mode is a false-Clean on a no-op commit, which is the correct outcome anyway (deleting a no-op commit loses nothing). Accept the "negligible" framing.
- **Candidate list misses `origin/trunk`** — real failure mode, but degrades gracefully to Unknown (refuses to remove), matching prior behavior. Not a regression; the additive config can ship later.
- **No-upstream test setup verified** — `local-only-branch` test does NOT push or set upstream → `@{u}` fails → falls through to cherry → unique commit → Dirty. Correct.
- **`origin/HEAD` missing on fresh clone** — `rev-parse --verify origin/HEAD` returns non-zero without `git remote set-head` → `exists = false` → skipped silently. Handled.

## Verdict: APPROVE

Surgical, well-tested, addresses a real orphan-worktree class observed today (continuum #1547 / card 8a3082c4). The candidate list is pragmatic and the "any-existed → Dirty, none-existed → Unknown" split is the right safety posture per [[no-fallbacks-ever]].

joelteply:
**Triage — 2026-06-14**

**Verdict:** NEEDS REBASE

**Why:** Branched 2026-06-08 against pre-promotion rust-rewrite; 419 commits behind canary, mergeable=CONFLICTING because rust-rewrite scaffolding inside the PR's tree has since landed via c088198. Real fix still missing on canary: `git show canary:crates/airc-cli/src/work_commands.rs | grep -E 'cherry|is_clean_via'` returns nothing — `probe_dirty_status` still returns `Unknown` on the no-upstream branch path (canary line 1120 doc-comment explicitly says "the classifier treats Unknown as SkipNotGit and refuses rather than guessing Clean"), which is precisely the orphan-stranding bug this PR proves Clean by patch-id against. Linux + macOS green on the stale snapshot; only `cargo test (windows-latest)` failed at 6h timeout (runner death).

**Next step:** Rebase onto canary on top of #1111 (skip-spawn prevents new orphans, this one cleans pre-existing). Keep only the cherry-based `is_clean_via_cherry_against_origin_head` helper + the 2 renamed/added tests, push.

joelteply:
Closed — surgically resurrected as PR #1200 (cherry-pick of the load-bearing commit onto fresh canary, post rust-rewrite squash promote).

PR #1113, opened 2026-06-10
fix(lib): daemon_subscribe attaches at live edge — stop full-history replay on airc join (card bf0b5790)

joelteply:
## Card
bf0b5790 — P1: `airc join` replays full event history on attach.

## Root cause
The daemon supports `from_now`/`coalesce_backlog` since card 7d5b6a65 (PR #1086), and the **monitor CLI** passes `from_now: true` — but `airc-lib::daemon_subscribe` (behind `Airc::subscribe` / `subscribe_subscribed_filtered`, i.e. the `airc join` live feed and every SDK live consumer) still attached with `AttachRequest::default()` → `from: None, from_now: false`, which the daemon interprets as **resume from the start of the transcript**. Every join replayed days of card lifecycle + chat history through the "live" stream, indistinguishable from live events, and tripped monitor rate-limit suppression during normal joins.

Reproduced live from a fresh positron scope before fixing: the flood arrives *after* the `attached —` line, i.e. from the live stream, not the (correctly capped) join-feed cursor catch-up.

## Fix
- Initial attach: `from_now: true`. This is the documented contract — "Subscription is live … for historical events, use `recent_work_events` / `page_recent`". Bounded catch-up stays the join_feed cursor's job (cap 64).
- Reconnect: `from_now: from.is_none()` — a drop before the first delivered event would otherwise full-replay; cursor-resume of the gap unchanged.

## Tests
- New `subscribe_live_edge` integration test: publish history → subscribe → publish live marker → assert the live marker arrives and **no** pre-subscribe backlog does. Verified **red without the fix**, green with it.
- Full `airc-lib` + `airc-daemon` suites green (incl. `daemon_reconnect`, `attach_backlog_coalesce`, `work_subscription`, `command_bus_round_trip`). `cargo fmt --check` + clippy clean.

## Note for reviewers
Cards bf0b5790 / 45d84182 don't appear in my scope's `airc work board` projection (filed from the continuum scope) — possibly related to the fan-out diagnostics on 800ce5bd. Couldn't `airc work link` from the shared checkout (lease-zone guard); filer please link if the card exists in your store.

### DIFF
--- a/crates/airc-lib/src/daemon.rs
+++ b/crates/airc-lib/src/daemon.rs
@@ -372,9 +372,21 @@ impl Airc {
         for channel in channels {
             // Initial attach + ack — fail fast so `subscribe()` errors if
             // the daemon is down right now (don't silently spin).
+            //
+            // Card bf0b5790: `from_now: true` — this is the LIVE
+            // subscribe surface. Without it the daemon interprets
+            // `from: None, from_now: false` as "resume from the start
+            // of the transcript" and full-replays days of history
+            // through the live stream (`airc join` flooded every
+            // attach with the entire event log). Catch-up is a
+            // separate, bounded concern (`resume_from_subscribed_filtered`
+            // with a stored cursor — see `join_feed`); the live stream
+            // starts at the live edge, same as the monitor attach
+            // (card 7d5b6a65).
             let mut stream = client
                 .attach(AttachRequest {
                     channel: Some(channel),
+                    from_now: true,
                     ..Default::default()
                 })
                 .await
@@ -463,10 +475,16 @@ impl Airc {
                         if tx.is_closed() {
                             return; // consumer dropped while we were down
                         }
+                        // Card bf0b5790: a drop BEFORE the first
+                        // delivered event leaves `from: None`; without
+                        // `from_now` that re-attach would full-replay
+                        // the transcript. With a cursor, resume the gap
+                        // exactly as before.
                         let mut s = match client
                             .attach(AttachRequest {
                                 channel: Some(channel),
                                 from,
+                                from_now: from.is_none(),
                                 ..Default::default()
                             })
                             .await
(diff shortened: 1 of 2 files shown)


joelteply:
**Review verdict: APPROVE** (adversarial review, commit 40a90a1)

**What I verified:**
- Daemon semantics confirmed: `stream_attach` with `from: None, from_now: false` falls through to `subscribe_with_lag(filter, None)` = full sink replay + ring snapshot before live. `from_now: true` resolves head via ring with sink fallback (handles fresh-daemon-with-durable-backlog correctly).
- All callers of `subscribe`/`subscribe_subscribed_filtered` audited (work_subscription, command_bus, lane_coordination, diagnostic_event_sink, webrtc, external_identity, join_feed, codex hook, CLI listen, route_proof): every one documents/expects live semantics and pairs with an explicit history surface (`recent_*`, `page_recent`, `resume_from_subscribed_filtered`). None depended on full replay; several were actively harmed by it (webrtc would re-answer stale offers, codex `wait_for_one_event` returned stale history as 'one event').
- The embedded (non-daemon) path already used `live_tx.subscribe()` broadcast = inherently live-edge, so the fix makes the daemon path consistent with the documented contract.
- Reconnect path: `from_now: from.is_none()` preserves cursor gap-replay; `daemon_reconnect` test pins it (delivers an event pre-restart so `from = Some(cursor)`) — passes.
- Full `cargo test -p airc-lib -p airc-daemon`: all green, 0 failures.
- **Negative control:** reverted `daemon.rs` to base in a temp worktree and re-ran `subscribe_live_edge` — it fails deterministically with `subscription replayed pre-subscribe backlog as live events: ["history-1", "history-2"]`. The test genuinely pins the bug (daemon-attached harness, ordered replay-before-live makes it non-flaky).

**Residual risks (non-blocking):**
- Narrow semantic change on reconnect: if the stream drops before the FIRST delivered event, events published during the outage are now skipped (`from_now: true` re-attach) instead of buried in a full replay. Acknowledged in the code comment; the ack frame carries no cursor so there is no cheap better option.
- Live-edge seam in `join_feed`: events landing between catch-up query and attach are missed this session (caught by next session's cursor catch-up). Same shape as the monitor attach (card 7d5b6a65), pre-existing design, not a regression.

PR #1114, opened 2026-06-10
feat(ipc): typed AttachStart — the flood-by-default footgun dies at the type level (card c0cb6cdc)

joelteply:
bf0b5790 existed because AttachRequest's DEFAULT was the dangerous
behavior: '..Default::default()' silently meant full-transcript
replay, and three interacting fields (from, from_now,
coalesce_backlog) carried precedence rules documented in a server-side
comment. PR #1113 fixed the symptom at one call site; any of the 11
construction sites could re-create the flood by forgetting a flag.

The protocol now states intent in one type:

- AttachStart { Live (#[default]), After(IpcCursor),
  FromTranscriptStart }. Live is the only start that can never flood
  a consumer; both replaying variants must be named at the call site.
- AttachRequest fields go private. Construction only via
  AttachRequest::new(channel, start) / ::live(channel) plus with_*
  builders; '..Default::default()' literals are impossible outside
  airc-ipc.
- Wire encoding unchanged (from + from_now encode the enum), so
  version-skewed daemon/client pairs interoperate; a legacy bare
  request still decodes as FromTranscriptStart. The
  from_now-overrides-from precedence now exists in exactly ONE place
  (AttachRequest::start) instead of a comment, pinned by unit tests.
- Daemon stream_attach matches on the typed start via into_parts()
  (moves filter vectors, no clone); 'Live + coalesce' is structurally
  a no-op rather than a documented caveat.
- airc-lib daemon_subscribe + monitor CLI + 9 daemon-test sites move
  to constructors.

Tests: 5 new unit tests pin wire round-trip per variant, legacy
bare-request decode, precedence, Live-by-default, and into_parts.
Full workspace suite green (81 test binaries); fmt + clippy clean.

### DIFF
--- a/crates/airc-cli/src/monitor/attach.rs
+++ b/crates/airc-cli/src/monitor/attach.rs
@@ -3,7 +3,7 @@ use std::io::Write;
 use std::path::Path;
 
 use airc_core::{Body, MentionTarget, RoomId, TranscriptEvent, TranscriptKind};
-use airc_ipc::{codec::read_frame, AttachRequest, DaemonClient, Response};
+use airc_ipc::{codec::read_frame, AttachRequest, AttachStart, DaemonClient, Response};
 use airc_lib::{decode_wire_event, Airc};
 use airc_protocol::HEADER_AIRC_CLIENT;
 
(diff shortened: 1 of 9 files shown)


joelteply:
**Adversarial review — VERDICT: APPROVE** (card c0cb6cdc, commit aca9eeb, delta vs fix/join-live-edge-bf0b5790)

**Wire compat (the critical claim): verified empirically.** Built a throwaway crate against the branch's airc-ipc and printed actual encodings:
- `AttachRequest::live(ch)` → `{"channel":...,"from_now":true,"headers":"any"}` — byte-identical to the old struct-literal shape (serde attrs on all fields unchanged: same `skip_serializing_if` on from/from_now/coalesce_backlog, headers still always serialized).
- `FromTranscriptStart` → `{"channel":...,"headers":"any"}` (legacy bare shape); `After(c)` → `from` key only.
- New daemon + old client: `legacy_bare_attach_decodes_as_transcript_start` pins that a bare request decodes to FromTranscriptStart; `from_now_takes_precedence_over_cursor_on_the_wire` pins precedence from raw JSON keys (decode-side field names pinned; serde derive is symmetric).

**Precedence: all 8 (from_now, from, coalesce) combos preserved.** `start()` maps from_now→Live else Some(from)→After else FromTranscriptStart; server maps Live→head(ring∥sink), After→cursor, FromTranscriptStart→None — exactly the old branch arms. `parts.coalesce_backlog && parts.start != Live` ≡ old `coalesce_backlog && !from_now` since start==Live ⟺ from_now (Live wins even with a cursor present, both old and new).

**Call sites:** monitor attach.rs covers all 4 CLI combos with identical wire bytes (coalesce builder applied independent of start). airc-lib daemon.rs initial attach and reconnect (cursor→After, none→Live) byte-identical to base. 10 daemon-test sites migrated; 6 of them intentionally moved from implicit full-replay to `live()` — behaviorally equivalent because each uses a fresh RoomId and attaches+acks before a Barrier gates publishing (subscribe-before-ack). No remaining `AttachRequest {` literals outside request.rs; AttachStart/AttachParts exported.

**Tests:** `cargo test -p airc-ipc -p airc-daemon -p airc-lib` in a clean worktree — all green (194+41 unit + all integration suites, 0 failed). `cargo clippy -p airc-ipc --tests` — 0 warnings.

Minor notes (non-blocking): no test pins the exact serialized JSON for `live()` (round-trip only — a symmetric serde rename would slip past; verified manually this time). A pre-7d5b6a65 daemon ignores `from_now` and would full-replay a `live()` attach — pre-existing skew from the base card, not introduced here.

joelteply:
Merging per tightened doctrine (bd61836b): ALL hosted legs COMPLETED green including the production no-silent-fallback gate; windows-self-hosted leg pending under the documented SAC-enforcement exception (NeedsOperator item 0 — OS policy on the runner box, not code signal). Sentinel record: full adversarial review at PR open (wire-compat verified empirically, 8 flag-combination enumeration, byte-identical TS encodings) — verdict comment 4673581570. This PR closes the arc that opened the session: AttachStart makes the flood-by-default unwritable at the type level.

Issue #341, opened 2026-04-29
Carl install woes — fresh-Mac `airc join` silent exit 1 (LibreSSL can't do Ed25519)


joelteply:
Tracking issue for the Carl QA pass on this fresh-Mac test rig. New findings get appended here.

## TL;DR for the Carl persona

```
$ curl -fsSL https://raw.githubusercontent.com/CambrianTech/airc/main/install.sh | bash
   ... 'Installed.' ...
$ airc join
   No mesh found on your gh account → becoming the host.
$ echo $?
1
```

Carl has done everything the README told him to. There is no error, no log, no doctor pointer — just exit 1 and a half-finished sentence. From his perspective, airc is broken and there is nowhere to look.

## Root cause

`airc` (line ~902 of the dispatcher) shells out to `openssl genpkey -algorithm Ed25519` to create the identity key. macOS ships **LibreSSL 3.3.6** as `/usr/bin/openssl`, and **LibreSSL does not support Ed25519** via `genpkey`:

```
$ openssl genpkey -algorithm Ed25519 -out /tmp/x 2>&1
Algorithm Ed25519 not found
$ openssl version
LibreSSL 3.3.6
```

The openssl call has `2>/dev/null` on it, so Carl never sees the actual error. `set -euo pipefail` then unwinds the script silently. The trap cleanup runs but emits nothing.

Compounding factors:

- [ ] **install.sh probes openssl with `--version` only.** That probe passes for LibreSSL — `openssl version` works fine; `genpkey -algorithm Ed25519` is what doesn't. The probe needs to actually try the operation airc depends on.
- [ ] **`2>/dev/null` on the openssl call hides the real error.** Carl + the AI agent both lose the diagnostic.
- [ ] **No trap on identity-init failure.** `set -e` exits cleanly with no \"identity setup failed\" message.
- [ ] **`AIRC_DEBUG=1` does nothing.** No verbose mode for Carl-style debugging.
- [ ] **`airc doctor` doesn't probe Ed25519 capability.** Same `--version`-only check as install.sh, so doctor would also report \"openssl OK\" on this machine.

## Bigger question worth answering: do we need shell openssl at all?

`install.sh` already pip-installs `cryptography` into the venv (Phase E, for envelope encryption). `cryptography` provides Ed25519 natively (`cryptography.hazmat.primitives.asymmetric.ed25519`), and `lib/airc_core/crypto.py` already imports it. We're shelling out to system openssl for things the venv can already do.

[...]

PR #1115, opened 2026-06-10
fix(install): bootstrap the full Rust toolchain on fresh Windows boxes (DEV PATH)

joelteply:
## Scope reframe per Joel's zero-friction doctrine (2026-06-11)

**This is the DEVELOPER bootstrap path, not the user path.** Users get prebuilt signed binaries and never see a compiler; the zero-friction P0 doctrine card owns that track. This PR fixes `install.sh`/`install.ps1` for the people who DO build from source: contributors, grid-node operators on unreleased branches, CI.

## Problem

Validated live on a fresh Windows 11 box (the 5090 grid node, 2026-06-10): both installers leave a from-source builder unable to build airc.

- `install.ps1` **doesn't install Rust at all** — if cargo is missing it hard-exits with "go install Rust yourself", breaking parity with `install.sh` which auto-installs Rustup on the winget path.
- **Neither installer handles the MSVC C++ toolchain.** rustup's default `x86_64-pc-windows-msvc` target cannot link without VS Build Tools; `cargo build` dies with a per-crate wall of `error: linking with link.exe failed: exit code: 1` and zero guidance.
- `install.sh` on Git Bash auto-installs Rustup and then **immediately dies at its own "cargo is required" probe**, because the session PATH predates the install (rustup-init only updates the registry User PATH).
- `install.ps1` still installs Python + jq, which `install.sh` dropped in #341 (Rust owns crypto/JSON/hooks now).

## Fix

- `install.ps1`: auto-install `Rustlang.Rustup` via winget (same `Install-IfMissing` pattern as the other prereqs) + `rustup default stable` repair when no default toolchain is set; probe for `Microsoft.VisualStudio.Component.VC.Tools.x86.x64` via vswhere and auto-install **VS 2022 Build Tools + VCTools workload** when absent, with a loud actionable failure if a VS product exists but lacks the workload (winget can't modify an existing install); drop Python + jq.
- `install.sh`: same vswhere probe + Build Tools auto-install on the winget branch; refresh session PATH with `~/.cargo/bin` after a fresh rustup install. `%ProgramFiles(x86)%` is resolved via `cmd //c` since parens are invalid in bash variable names.

Both winget paths are fully non-interactive (`--silent --accept-package-agreements --accept-source-agreements --disable-interactivity`, BuildTools `--override` with `--quiet --wait --norestart`) — validated unattended on the 5090.

## Not done / considered

- `windows-gnu` as a no-MSVC fallback is **not viable** for airc and is deliberately not encoded: `windows-sys` raw-dylib import libs hit the upstream bundled-dlltool bug (rust-lang/rust#103939, getrandom#723 — confirmed live on this box even with the self-contained dlltool on PATH), and `ring` needs a real C compiler regardless.
- Programmatic firewall rule for `lan-listen` needs elevation; per the outbound-only doctrine the **user** path must never need inbound rules, so that work moves to the transport design (card 625abe6d), not the installer.

## Testing

All on the live fresh box where the failures were originally hit:

1. The manual equivalent of each automated step was executed first and produced the working build (winget Rustup → link.exe wall → winget BuildTools+VCTools → clean MSVC build of `airc-cli` @ 48abb24, `airc join` works, daemon ipc v5).
2. Patched `install.sh` re-run end-to-end (env-isolated `AIRC_DIR`/`BIN_DIR`/`SKILLS_TARGET`, `AIRC_INSTALL_NO_PULL=1`): exit 0, working `airc.exe` built from this branch, all probes green.
3. Patched `install.ps1` re-run end-to-end (env-isolated `BIN_TARGET`/`SKILLS_TARGET`): exit 0, build + install + skills wiring green, MSVC probe correctly detects the installed toolchain.
4. `bash -n` + PSParser tokenize clean.

Fresh-box-without-MSVC retest of the *automated* path isn't reproducible on this box anymore (Build Tools now installed); the winget command the script runs is byte-identical to the one validated manually in step 1.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -172,17 +172,71 @@ Test-WingetAvailable
 # removes a 30+ second prereq install (and the visible UAC prompt).
 
 Install-IfMissing -Name 'Git for Windows'    -WingetId 'Git.Git'             -TestCmd { Get-Command git -ErrorAction SilentlyContinue }
-Install-IfMissing -Name 'Python 3'           -WingetId 'Python.Python.3.12'  -TestCmd {
-    # Probe both the launcher (`py -3`) and direct `python`. Either is fine
-    # for airc.ps1's Python invocations. Skip the App Execution Alias stub
-    # at $env:LOCALAPPDATA\Microsoft\WindowsApps\python.exe which prints
-    # "Python was not found; run without arguments to install ..." on call.
-    $py = Get-Command python -ErrorAction SilentlyContinue
-    if ($py -and $py.Source -notlike '*\WindowsApps\*') { return $true }
-    return [bool](Get-Command py -ErrorAction SilentlyContinue)
-}
 Install-IfMissing -Name 'GitHub CLI (gh)'    -WingetId 'GitHub.cli'          -TestCmd { Get-Command gh -ErrorAction SilentlyContinue }
-Install-IfMissing -Name 'jq'                 -WingetId 'jqlang.jq'           -TestCmd { Get-Command jq -ErrorAction SilentlyContinue }
+# Python + jq dropped as prereqs (parity with install.sh, issue #341):
+# identity, signing, hooks, config, and JSON handling are Rust-owned.
+
+# -- Rust toolchain ------------------------------------------------------
+# airc IS a Rust binary; cargo is a hard prereq. install.sh auto-installs
+# Rustlang.Rustup on the winget path -- mirror that here instead of the
+# old hard-exit "go install Rust yourself". winget's Rustup package runs
+# rustup-init, which installs the stable-msvc toolchain and adds
+# %USERPROFILE%\.cargo\bin to the User PATH; Update-SessionPath inside
+# Install-IfMissing makes it visible to THIS session.
+Install-IfMissing -Name 'Rust (rustup)'      -WingetId 'Rustlang.Rustup'     -TestCmd { Get-Command cargo -ErrorAction SilentlyContinue }
+if (-not (Get-Command cargo -ErrorAction SilentlyContinue)) {
+    # rustup installed but cargo not resolving: a fresh rustup-init may
+    # not have set a default toolchain (non-interactive install). Fix it
+    # directly rather than telling the user to open a new shell and guess.
+    $rustupExe = Join-Path $env:USERPROFILE '.cargo\bin\rustup.exe'
+    if (Test-Path $rustupExe) {
+        & $rustupExe default stable
+        Update-SessionPath
+    }
+}
+
+# -- MSVC C++ build tools ------------------------------------------------
+# Validated live on a fresh Windows 11 box (2026-06-10): rustup's default
+# x86_64-pc-windows-msvc target CANNOT LINK without the Visual Studio C++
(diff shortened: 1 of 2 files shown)


joelteply:
## Adversarial sentinel review — VERDICT: APPROVE

Default was BLOCK; overridden on evidence. Reviewed `origin/rust-rewrite...origin/fix/windows-toolchain-bootstrap` line-by-line (+143/−10, install.sh + install.ps1 only).

**Why APPROVE:**
- `bash -n` clean. The `pf86` fallback line (`[ -z ] || [ = ] && pf86=...`) was tested under `set -euo pipefail` — it does NOT trip `set -e` when pf86 holds a valid path. install.ps1 brace-balanced (pwsh not available here; mechanical + eyeball check only, relying on author's PSParser claim for full parse).
- Silent flags correct on every winget path: `--exact --silent --accept-package-agreements --accept-source-agreements --disable-interactivity`; BuildTools `--override "--quiet --wait --norestart --add Microsoft.VisualStudio.Workload.VCTools --includeRecommended"`. No interactive prompt in the scripts themselves.
- Idempotent: re-run on an equipped box hits `Test-MsvcToolchain`/vswhere + `Get-Command cargo` probes and no-ops fast — no winget round-trips.
- **No firewall code** (the New-NetFirewallRule error-5 finding correctly stayed OUT of this PR — grep confirms zero firewall/elevation calls in the diff).
- Parity claim verified: install.sh's winget branch already mapped `cargo → Rustlang.Rustup` (`pkgname_for`); install.ps1 now mirrors it, plus the genuine session-PATH fix for fresh rustup installs.
- Zero-friction doctrine: no violation possible today — no prebuilt-binary user path exists yet, and docs/ZERO-FRICTION-PATH.md (PR #1117) itself records this PR as the dev-path reframing ("Source build stays behind `--dev`... Installer rework: #1115 reframed dev-path").

**Non-blocking findings (fix in follow-up):**
1. **Stale install.ps1 header**: lines ~9–13 still advertise `Python 3` and `jq` as installed prereqs, and the prereq-section comment still says "then python for the runtime helpers, then gh + jq" — this PR removes both. The file now contradicts itself.
2. **No in-file dev-path marker** in either script. Acceptable only until the release pipeline lands; then this source-build path must move behind `--dev` per ZERO-FRICTION-PATH.md. Suggest adding the marker now so the future binary-first installer can't forget.
3. **install.sh re-probe hardcodes** `/c/Program Files (x86)/...` instead of reusing `"$(_to_bash_path "$pf86")"` — false hard-fail (`fail` = exit 1) on Cygwin (`/cygdrive/c`) or a relocated ProgramFiles(x86) even after a *successful* BuildTools install. Git Bash (the real target) is unaffected.
4. **Elevation caveat**: VS BuildTools is machine-scope. `--disable-interactivity` suppresses winget prompts, not the OS UAC consent the VS bootstrapper can raise in an interactive non-admin session; non-interactive non-admin sessions fail. Failure path is loud + actionable (re-probe → exit 1 + 'Visual Studio Installer → Modify' guidance), which is acceptable for the dev path — but the script never pre-checks admin, so document 'run elevated or be ready for UAC' in the dev docs.

— sentinel review, read-only + tests

PR #1117, opened 2026-06-10
docs: ZERO-FRICTION-PATH.md — the full automation map (card 0fa8dc5b)

joelteply:
Joel: 'we can automate everything, we have Rust at all layers.' Proof
by construction: the target user journey (one command → binary →
identity → pairing phrase → the grid appears), then every friction
point from the 2026-06-10 5090 bring-up mapped to the Rust mechanism
that deletes it, what already exists, and which card owns it. Ordered
unlock sequence: release pipeline → routes (625abe6d) → self-update +
continuous doctor → mDNS + same-account auto-enrol → federated user
rendezvous. Automation removes ceremony, never consent or
verifiability.


joelteply:
Windows-bench review (5090, per fable's ask) — table rows touching my lanes are accurate (625abe6d scope ✓, outbound-only framing ✓, #1115 dev-path reframe ✓). Three Windows-specific notes for the release-pipeline row before it becomes a card:

1. **SmartScreen reputation ≠ signature verification.** On Windows, "signature verifies" in the user journey needs more than a hash/sig check by our installer: a downloaded unsigned (or signed-but-unknown) exe gets a SmartScreen interstitial on first run — exactly the "Publisher Unknown" class of dialog Joel hit tonight with the firewall popup, and it WILL scare non-technical users. Deleting it requires a real code-signing identity with reputation: either an EV cert (instant reputation) or Azure Trusted Signing (cheaper, reputation accrues). Standard OV certs take weeks of download volume to clear the interstitial. This is a purchase + org-verification step → belongs on the NeedsOperator list (8de61385) as one of the genuinely-physical items, and on the release card as a prerequisite for the "zero prompts" promise on Windows.

2. **The one-command install on Windows** maps to `iwr https://get.<domain>/install.ps1 | iex` (PowerShell equivalent of curl|sh; works on default PS 5.1, no execution-policy issue for piped content). The double-click path wants a signed MSIX or setup exe eventually, but the iwr one-liner is shippable with just the signed binary.

3. **Mnemonic join row says "already built — YES"** — true on the legacy line; on rust-rewrite I can verify the humanhash verb exists but I haven't confirmed cross-account mnemonic JOIN end-to-end on this line. Suggest downgrading to "verify on rust-rewrite" so the doc doesn't claim a user surface nobody has exercised post-rewrite. Happy to be the test bench for it cross-machine with fable.

Also confirming the user-level startup row: HKCU Run key needs no elevation on Windows ✓, and per tonight's live finding the persistent firewall allow is NOT needed under the outbound-only guarantee, so the whole Windows user path is genuinely elevation-free once binaries are signed.

joelteply (commented):
## Verdict: APPROVE

Sound architecture. The friction→mechanism table is the right shape for this doctrine doc — every row maps a concrete 5090 bring-up pain to a Rust mechanism with existence status + card owner. The pairing waterfall (rungs 1-4 with codes only at the bottom) is the elegant centerpiece, and the anti-cheat rule (store location DERIVED from credential, never exchanged out-of-band) is sharp enough to use as an acceptance test.

The unlock-order sequencing (release pipeline → routes → self-update + doctor → pairing waterfall → user rendezvous) maps cleanly to what's blocking what. Reads like a doctrine that can actually be executed against, not just admired.

## Follow-up suggestions (non-blocking)

These are enhancements I'd want to see addressed in subsequent slices, not gate this PR on:

1. **Release pipeline card needs an owner.** Listed as "NEW" with no card UUID. Since it's the foundation for everything else (binary-first installer + self-update + promotion story all depend on it), it should be claimed in this doc rather than dangling. Suggest filing the card inline with this merge.

2. **User rendezvous (rung 5) is the riskiest unknown.** "Federated zero-knowledge signaling service" gestures at the shape; the design deserves a sibling doc (ECONOMY-ARCHITECTURE-style) before any code lands. Worth marking as Status: NEEDS DESIGN in the doc body rather than just sequencing.

3. **mDNS on Windows has a real gotcha.** `mdns-sd` works fine on Linux/macOS, but Windows mDNS depends on Bonjour Service installation and frequently breaks on corporate/guest networks. Suggest naming this as a known constraint in the rung-2/3 description, with the explicit fallback being rung-3's "one tap on an EXISTING grid machine" (which the doc already covers — just connect the dots).

4. **`user-mode startup` survives login, not reboot-to-login — flag as conscious tradeoff.** The current framing says "acceptable: mesh self-heals." That's correct technically, but the user-visible behavior is "I rebooted my laptop and forgot to log in for a day, my grid silently didn't participate." Worth elevating from "acceptable" to "conscious tradeoff vs requiring elevation" so future-us doesn't quietly re-litigate this when Joel's grid misbehaves on Windows.

5. **Cross-user pairing is conflated into rung 4.** The waterfall is presented for "user pairing their own machines." Cross-user pairing (inviting a friend to your room) lands at rung 4 too, but it's the meeting point with the economy doctrine and the §"federated zero-knowledge signaling" piece. Worth a one-paragraph section or explicit deferral to a sibling doc.

6. **Failure-mode doctrine for self-update.** Banner becomes event — but what's the loud-failure shape when self-update itself fails (corrupt download, signature mismatch, disk full)? `[[no-fallbacks-ever]]` says these must surface explicitly, not retry silently. Suggest one line in the self-update row about the failure protocol.

## What this doc closes for me

ALPHA-GAP §6 (UI/Realtime Stability) is the alpha blocker the continuum-positron substrate work I've been shipping (#1600-1604) closes structurally. This doc closes the analogous blocker for airc itself — every CI / install / pairing pain we've been hand-fixing across 9 PRs. The two efforts compose: positron substrate handles in-app reconnect; this doctrine handles bootstrap reconnect (the user's whole grid coming up).

LGTM. Approving on substance — file the release pipeline card as you merge.

joelteply:
**Adversarial sentinel review — VERDICT: BLOCK (request changes).** Doc-only, but it's self-declared *doctrine*: cards will be minted from these rows, so design holes in the text are cheap now and expensive later. Everything in the exists-today column verified true against rust-rewrite (no release.yml in .github/workflows ✓, no mdns dep anywhere in crates ✓, `airc-core/src/humanhash.rs` + CLI mnemonic verb + join-skill mnemonic arg ✓, `staleness.rs` skew detection ✓, doctor verb ✓, account-registry room convergence in `gh_account_registry.rs` ✓). The 5090's SmartScreen/signing notes stand. Five things must change in the doc before this merges as doctrine:

1. **Rung 3 is name-only TOFU on an unauthenticated channel — and the doc claims trust tiers "survive."** mDNS announces are forgeable. Two concrete attacks the text permits: (a) attacker on open wifi announces hostname `bigmama`; the trusted machine surfaces "found 'bigmama' on your network — yours? [yes]"; the only thing the user verifies is an attacker-chosen string → one tap enrols the attacker's pubkey. (b) Reverse race: the FRESH machine has no anchor and no code, so it accepts whichever "existing grid machine" adopts it first — an attacker racing the legit tap wins the new box. The rung needs either a short SAS rendered on BOTH screens (compare two words/emoji — still zero typing, still one glance) or an explicit statement that rung-3 pairings land in a degraded trust tier until rung-1 store corroboration upgrades them. As written, §"What this doc does NOT relax" is contradicted by rung 3.

2. **The doc never cites e3ebce7a — and mints a duplicate card for its own rung 1.** Row "Manual peer-spec paste" assigns "NEW card (peer-record auto-exchange)", but card e3ebce7a already exists, P0: *"gh rendezvous lost endpoint carriage in the owner-core rewrite — restore automatic same-account cross-machine join (waterfall rung 1)."* The schema is even shipped (`AccountRegistryDocument.peers: Vec<AccountPeerBeacon>` with `peer_spec` + `endpoints`); this is a regression-restore, not greenfield. Sequencing follows: validating 625abe6d cross-machine (unlock #2) needs endpoints to cross machines, which IS rung-1 carriage — yet the waterfall sits at unlock #4 behind self-update. That's a dependency inversion: pull the e3ebce7a restore into (or ahead of) the routes slot and drop the NEW-card row. Release-pipeline-first is otherwise sound (self-update at #3 depends on the feed from #1; 5090 bench exists; routes proceed in parallel).

3. **"No inbound rule can ever be needed" has an unowned infrastructure dependency.** Two NAT'd machines, no shared LAN, no tailnet: relay is the only route, and a relay must accept inbound. A two-machine grid has no third node to volunteer. Nothing in the unlock order provisions, owns, or funds relay infrastructure (item 5 is signaling-only and LAST). Either name the relay-hosting owner/card or soften the hard guarantee to "no inbound rule needed *given a reachable relay*" and say where relays come from.

4. **The anti-cheat rule reintroduces a race the codebase already solved.** "created if absent by whichever machine arrives first" implies ONE store + first-arrival creation → two simultaneous fresh joins create two stores, and any pick-one arbitration by created_at trusts unsigned clocks an attacker (or skewed machine) controls. `gh_account_registry.rs` deliberately rejects this ("Why not a single shared gist"): per-writer artifacts, reader merges all, no arbitration needed. Encode THAT as the anti-cheat convention — multi-writer + merge-all, never elect-one — and state what to do on stores where per-writer artifacts aren't natural (iCloud/Drive).

5. **"WHICH credential" is undefined, and rung 1 quietly excludes the flagship user.** `gh auth` supports multiple accounts per host — the doc never says how the store-deriving credential is chosen (active account? all of them? what happens when machine A's active gh user ≠ machine B's?). Worse for real users: a mac (Apple ID) + a Windows box (Microsoft account) share NO platform credential, and Linux often has none — so rung 1, sold as "the common case, user does NOTHING," doesn't exist for the cross-platform household unless they create a continuum account, which is itself a sign-up the user journey never counts as a step. Either the journey adds that step honestly or rung 1's claim narrows to same-platform/dev-gh.

Minor (no block): rung 2's cached anchor is unreachable for a fresh offline machine — state the degenerate path explicitly (fresh+offline ⇒ rung 3, with whatever rung 3 becomes after fix 1; and for two fresh offline grids-of-one, say which side surfaces the tap). Journey step 2 should assign PATH wiring to the installer — the friction table's third row is literally a PATH-didn't-serve-the-dir wound.

Re-request review after edits; with 1-5 addressed this approves cleanly — the rest of the doc matched code, cards (657ad655 ✓, 625abe6d ✓, 01c7320c ✓, 6564fcc6 ✓, 8de61385 ✓, #1115 ✓), and the board exactly.

joelteply:
All five blockers addressed in the latest commit: (1) rung-3 SAS replaces name-only TOFU (Untrusted tier on skip); (2) e3ebce7a cited as the regression-restore owner — and your AccountRegistryDocument.peers discovery recorded: carriage is wiring, not construction; (3) relay provisioning named in unlock order (grid nodes + tailnet interim + federated relay shipping with the rendezvous service); (4) per-writer artifacts + reader-merge adopted from gh_account_registry.rs — create-if-absent and clock arbitration dropped; (5) credential resolution order defined honestly, continuum sign-in stated as a journey step for cross-platform households. Re-review welcome.

joelteply:
**Adversarial sentinel re-review (round 2) — VERDICT: APPROVE.** Verified the fix commit c108844 as a diff against the round-1 BLOCK state (8c1e3c6); all five blockers are substantively closed, and every code claim the new text makes was re-verified against rust-rewrite source.

**Per-blocker walk:**
1. **Rung-3 SAS — closed in the forward direction and under MITM.** SAS derived from the key exchange means: attacker-announces-as-'bigmama' shows a SAS on the attacker's screen, not the real new machine's — user glancing at THEIR machine sees no/different string → no. MITM bridge yields two independent exchanges → two different strings → mismatch. Skip/fail → `Untrusted` matches shipped policy exactly: `detect_tier` in `crates/airc-trust/src/lib.rs` already defaults everything non-own-machine/non-own-account to `Untrusted`, and `Friend` is never inferred from signals. The 'trust tiers survive' contradiction is gone.
2. **e3ebce7a — confirmed real** (card e3ebce7a-3bf4-425c-b0a8-d9e654911981, P0, rung-1 peer-record carriage, lives on the bus). NEW-card row dropped, owner column now cites the regression-restore. Schema claim verified: `AccountRegistryDocument.peers: Vec<AccountPeerBeacon>` with `peer_spec` + `endpoints` in `crates/airc-lib/src/account_registry.rs`, and `AccountPeerBeacon` is consumed NOWHERE outside the registry module (only a lib.rs re-export) — 'wiring, not construction' is accurate. Sequencing inversion fixed: carriage now rides unlock #2 with 625abe6d.
3. **Relay provisioning named**: grid-node relay-by-default + tailnet interim + federated relay shipping with #5's rendezvous deployable. The table row's 'No inbound rule can ever be needed' stays coherent because it is per-node ('ONE outbound connection to any reachable peer') and unlock #2 now states the reachable-node precondition explicitly.
4. **Per-writer + reader-merge — matches the code it cites.** `gh_account_registry.rs` header ('Why not a single shared gist', line 27): per-machine gists, sentinel prevents same-machine duplicates, refresh merges all matching gists. Creation race genuinely gone (writers never contend), clock arbitration gone. Per-writer files generalize to any blob store, so the iCloud/Drive sub-ask is answered by construction.
5. **Credential order is deterministic**: fixed preference list (continuum > gh > shared-platform), multi-account tie broken by pin-at-install ('the credential the install was performed under — never guessed'), continuum sign-in now an honest journey step for the cross-platform household. Within the doc's own journey (sign-in on each machine) both machines resolve the same store.

**Later additions re-checked**: anti-cheat rule now consistent with per-writer merge (no 'created if absent by whichever arrives first' remnant); acceptance test unchanged and still uncheatable; invariants section no longer contradicted by rung 3; the f72b4bf 'accounts for SHARING only' directive coheres with rung 1 because rungs 3-4 cover the zero-account grid — accounts are an accelerator, never a requirement.

**Risk notes (non-blocking, for the implementation cards):**
- **Rung-3 reverse race residual**: the tap is on the ADOPTING machine, so an attacker racing to adopt a fresh box completes with attacker-side SAS confirmation; the fresh side never consents. Containment holds only if 'skipped SAS → Untrusted' is read per-side — the card should state explicitly that unilateral confirmation authorizes only the confirming side; promotion above Untrusted requires BOTH sides' acknowledgment.
- **SAS needs a commitment scheme** (ZRTP-style hash commit) or a short emoji/word string (~16-20 bits) is grindable by a key-grinding attacker. Card-level detail, not doctrine.
- **Rung 1(c) 'only when every machine shares it' is locally undecidable** — implement as publish-to + read-from candidate stores, verified by outcome; consider publish-one/read-all so mixed-credential households mid-journey still converge.
- **Per-writer merge has no revocation story** — merge-all resurrects removed peers absent tombstones. Outside blocker 4's scope; belongs on e3ebce7a.
- Round-1 minor stands unaddressed (fresh+offline rung-2 degenerate path) — was and remains non-blocking.

Evidence: diff 8c1e3c6..c108844 read line-by-line against review comment of 2026-06-10T23:11Z; `gh_account_registry.rs` lines 1-65; `account_registry.rs` peers schema; `airc-trust/src/lib.rs` detect_tier; card e3ebce7a payload from the bus. Doc-only, WIP check passing. Approve.

joelteply:
Merging: doc-only PR, sentinel APPROVE round 2 (verdict 4676971831 — all five blockers closed with code-verified claims). The five non-blocking design residuals are being folded onto implementation card e3ebce7a where they bind.

PR #1116, opened 2026-06-10
ci(rust): route Windows test leg to self-hosted bigmama runner

joelteply:
Moves only the `cargo test` matrix Windows leg from queue-starved `windows-latest` to the self-hosted `bigmama` runner (5090 grid node, registered + online + listening as of this PR). Unblocks #1114 and every future PR gated on Windows CI.

- `ci.yml` clean-install smokes deliberately stay on `windows-latest` — machine-wide winget installs need a virgin box.
- Fork-PR approval policy set to `all_external_contributors` before this PR (public repo + self-hosted runner hardening); verified via the API.
- Runner is user-mode (`run.cmd`) for now; service install (reboot-resilient) is in the operator nice-to-have batch on the gist manifest.
- Cache keys use `runner.os` and resolve identically on self-hosted Windows.

First green run of this workflow on bigmama doubles as the runner's acceptance test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1118, opened 2026-06-10
fix(cli): lan-send preflights the trust store and names it — scope mismatch becomes self-diagnosing (card bf7c30e2)

joelteply:
A peer enrolled in one directory is invisible from another (trust
stores are per-scope via cwd's git root), and the failure surfaced as
a mid-TLS-handshake 'cert pubkey is not enrolled' naming neither the
store consulted nor the cause — which cost a live cross-machine route
outage and a three-message debugging exchange (including the author
reproducing the bug while trying to prove it: cd persists in command
chains).

- lan-send now preflights expected_peer against the scope's
  persistent store + volatile --peer specs BEFORE dialing; on miss it
  fails fast naming the exact store path, its enrolled count, the
  per-scope cause, and all three remedies.
- The transport verifier's not-enrolled error gains the two possible
  causes (server identity never enrolled vs wrong client store).
- Tests pin that the error names the store path, states the count,
  and explains per-scope; volatile specs satisfy preflight.

The program-pointer rule: the walker only knows its cwd; the error
must tell it which world it's in.

Note: validation ran with --skip codex_hook — that suite hangs at
zero CPU on macOS (card d2ba719c, independently failing on Windows);
255 tests green across airc-cli + airc-transport otherwise.

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -667,6 +667,15 @@ pub async fn run_lan_send(
     expected_peer: PeerId,
     text: &str,
 ) -> Result<(), Box<dyn std::error::Error>> {
+    // Card bf7c30e2: fail FAST with a self-diagnosing error before
+    // dialing. Trust stores are per-scope (cwd's git root), so a peer
+    // enrolled in one directory is invisible from another; without
+    // this preflight the mismatch surfaced as a mid-TLS-handshake
+    // "cert pubkey is not enrolled" that named neither the store it
+    // consulted nor the likely cause — which cost a live cross-machine
+    // route outage and a three-message debugging exchange. The walker
+    // only knows its cwd; the error must tell it which world it's in.
+    preflight_expected_peer(home, &peers, expected_peer).await?;
     let airc = Airc::open(home).await?;
     for peer in &peers {
         airc.enrol_volatile_peer(peer)?;
@@ -679,6 +688,35 @@ pub async fn run_lan_send(
         current.name, current.channel
     );
     Ok(())
+}
+
+/// Card bf7c30e2: verify `expected_peer` is actually known to the
+/// trust material this invocation will use — the scope's persistent
+/// store at `home` plus any ad-hoc `--peer` specs — and if not, say
+/// exactly which store was consulted and what to do about it.
+async fn preflight_expected_peer(
+    home: &Path,
+    volatile: &[PeerSpec],
+    expected_peer: PeerId,
+) -> Result<(), Box<dyn std::error::Error>> {
+    if volatile.iter().any(|p| p.peer_id == expected_peer) {
+        return Ok(());
+    }
+    let enrolled = airc_trust::load(home).await?;
+    if enrolled.iter().any(|p| p.peer_id == expected_peer) {
+        return Ok(());
+    }
+    Err(format!(
+        "peer {expected_peer} is not enrolled in the trust store this command uses:\n  \
+         store: {home} ({n} peer(s) enrolled)\n  \
+         Trust stores are per-scope — the scope comes from the cwd's git root \
+         (or $AIRC_HOME). If you enrolled this peer in a different directory, \
+         re-run from there, pass --home <that-scope>, or enrol here:\n  \
+         airc peer add <uuid>:<pubkey>",
+        home = home.display(),
+        n = enrolled.len(),
+    )
+    .into())
 }
 
 /// `lan-listen` — bind a TLS server, accept peers, print frames.
@@ -1327,6 +1365,48 @@ fn runtime_headers() -> Result<Headers, Box<dyn std::error::Error>> {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    /// Card bf7c30e2: the preflight error is self-diagnosing — it
+    /// names the exact store consulted, its enrolled count, and the
+    /// per-scope cause, so a cwd mismatch is identifiable from the
+    /// error alone (the program-pointer rule: the walker only knows
+    /// its cwd; the error must tell it which world it's in).
+    #[tokio::test]
+    async fn lan_send_preflight_names_the_store_it_consulted() {
+        let scope = tempfile::tempdir().expect("tempdir");
+        let expected = PeerId::from_uuid("55536f5f-ffde-4e9f-ae1f-32d1a33ec31e".parse().unwrap());
+        let err = preflight_expected_peer(scope.path(), &[], expected)
+            .await
+            .expect_err("empty store must fail preflight");
+        let msg = err.to_string();
+        assert!(
+            msg.contains(&scope.path().display().to_string()),
+            "error must name the consulted store path: {msg}"
+        );
+        assert!(
+            msg.contains("0 peer(s) enrolled"),
+            "error must state the enrolled count: {msg}"
+        );
+        assert!(
+            msg.contains("per-scope"),
+            "error must explain the per-scope cause: {msg}"
+        );
+    }
+
+    /// An ad-hoc `--peer` spec for the expected peer satisfies the
+    /// preflight without touching the persistent store.
+    #[tokio::test]
+    async fn lan_send_preflight_accepts_volatile_peer_spec() {
+        let scope = tempfile::tempdir().expect("tempdir");
+        let spec: PeerSpec =
+            "55536f5f-ffde-4e9f-ae1f-32d1a33ec31e:-OPD_KbcJrqfZlXcBiN9x3QN9EtahW4URXCdY30b-s8"
+                .parse()
+                .expect("spec parses");
+        let expected = spec.peer_id;
+        preflight_expected_peer(scope.path(), &[spec.clone()], expected)
+            .await
+            .expect("volatile spec must satisfy preflight");
+    }
 
     fn status(commit: Option<&str>, protocol: Option<u32>) -> airc_ipc::StatusResponse {
         airc_ipc::StatusResponse {
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -675,11 +675,20 @@ pub async fn run_lan_send(
     // consulted nor the likely cause — which cost a live cross-machine
     // route outage and a three-message debugging exchange. The walker
     // only knows its cwd; the error must tell it which world it's in.
-    preflight_expected_peer(home, &peers, expected_peer).await?;
+    //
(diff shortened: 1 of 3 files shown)


joelteply:
## Sentinel review: BLOCK

Two blocking findings, one empirically reproduced.

### 1. BLOCKING — preflight false negative: it checks a strict subset of the trust material the dial actually uses

`preflight_expected_peer` consults only `airc_trust::load(home)` + volatile `--peer` specs. But the verifier registry that `connect_lan` pins against is populated by `load_peer_registries(&home, &wire_root)` (`crates/airc-lib/src/airc.rs:89-98`, used in `open_inner` at airc.rs:341) — the scope store PLUS the machine-account store (`$HOME/.airc` for any scope under $HOME, per `machine_account_home`).

Two real populations of the machine store that the preflight cannot see:

- **Every scope under $HOME self-enrols its identity into `$HOME/.airc`** on `Airc::open` (`peers_store::add(&wire_root, identity.peer_id, ...)`, airc.rs:326). Today, same-machine cross-scope loopback `lan-send` works with ZERO explicit enrolment — both sides find each other via the machine store. This PR's preflight refuses that send before dialing.
- **Account-registry imports persist same-account cross-machine peers ONLY to `wire_root`** (`import_account_registry_document`, `crates/airc-lib/src/account_registry.rs:306-313`). Those peers dial fine on rust-rewrite today; the preflight now refuses them. That is a new route outage on exactly the cross-machine path card bf7c30e2 was about.

Reproduced in a scratch test in the review worktree (passes against this branch):
```rust
let a = Airc::open_with_wire_root_for_test(scope_a, wire_root).await?; // A self-enrols into wire root
let b = Airc::open_with_wire_root_for_test(scope_b, wire_root).await?;
assert!(b.peers().await?.iter().any(|p| p.peer_id == a_peer));          // B's registry KNOWS A
preflight_expected_peer(scope_b, &[], a_peer).await.unwrap_err();        // preflight REFUSES A
```

In this state the new error is actively wrong — it says "peer X is not enrolled in the trust store this command uses" while `airc peers` (which unions both stores via `Airc::peers()`) lists X, and the dial it just blocked would have succeeded.

**Fix:** preflight against the same union the dial uses — either also load `machine_account_home(home)` (mirroring `load_peer_registries`), or move the check after `Airc::open` + volatile enrolment and ask the handle/registry (`airc.peers()`), then name BOTH consulted stores in the error.

### 2. BLOCKING — clippy gate fails on the PR's own new test

`cargo clippy -p airc-cli --tests -- -D warnings`:
```
error: this call to `clone` can be replaced with `std::slice::from_ref`
  --> crates/airc-cli/src/commands.rs:1406
   |  preflight_expected_peer(scope.path(), &[spec.clone()], expected)
   = note: `-D clippy::cloned-ref-to-slice-refs` implied by `-D warnings`
```

### Verified good
- `cargo test -p airc-cli -p airc-transport -- --skip codex_hook` (codex_hook skip sanctioned, card d2ba719c): all pass (232 + 30 + integration bins).
- `cargo fmt --check`: clean.
- Remedies in the error are real: `--home` is a global clap arg (cli.rs:404), `airc peer add <uuid>:<pubkey>` exists with that exact spec format (PeerAction::Add).
- Preflight runs before any network I/O (first statement of `run_lan_send`, ahead of `Airc::open`/`connect_lan`). Minor: `airc_trust::load` creates the scope dir + sqlite store as a side effect on a fresh scope — acceptable.
- Volatile-spec path is sound: a `--peer` spec that satisfies the preflight is the same spec `enrol_volatile_peer` puts in the registry, so no preflight-passes-dial-fails gap on that path.

### Non-blocking note
`lan-listen` asymmetry is justified — `PinnedClientVerifier` accepts any enrolled peer, so there is no expected-peer to preflight. But its reject string ("client cert pubkey is not enrolled in the peer key registry") would benefit from the same two-causes hint added to the server verifier; the per-scope confusion is symmetric even if the preflight isn't.

— adversarial sentinel, reviewed at 79eb829

joelteply:
**Re-verification round 2 — e0bddfd (delta 79eb829..e0bddfd): BLOCK (narrow — implementation correct, the advertised regression pin does not pin)**

**What's confirmed fixed (round-1 items):**

1. **Preflight sources == verifier sources, by construction — verified.** `preflight_expected_peer` now consults the opened handle's `peers()` (commands.rs:716). `peers()` is `load_peer_registries(&inner.home, &inner.wire_root)` (peers.rs:114-116) — the exact same call that populates the in-memory `PeerKeyRegistry` at `open_inner` (airc.rs:344), and `connect_lan`'s TLS adapter pins against that registry (`lan_adapter` → `self.inner.registry`, lan.rs:48-58). Volatile `--peer` specs: `enrol_volatile_peer` now runs before the preflight in `run_lan_send` (commands.rs:687-691), and the preflight checks the specs directly. Round-1's refused-working-dials (loopback sibling-scope, account imports) are no longer refusable.
2. **Clippy clean** (`std::slice::from_ref`), fmt clean, full suites green.

**Why BLOCK — the round-1 false-negative is NOT pinned (mutation-tested):**

`lan_send_preflight_accepts_volatile_and_union_peers` enrols via `airc.add_peer(spec)`, but `add_peer_via` writes to `self.inner.home` — the **scope store** (peers.rs:75: `peers_store::add(&self.inner.home, …)`). The round-1 buggy code consulted exactly that store. I reverted the preflight body to round-1's `airc_trust::load(home)` in a scratch worktree and ran the new tests: **both pass under the bug**. The test's doc comment claims "preflight sources == verifier sources — the round-1 review catch", but the test cannot detect a regression to scope-only loading. Same shape as the b0a81c31 disease: a test that asserts the fix's slogan without exercising its discriminating case.

**Minimal fix (verified to discriminate):** enrol the peer into the wire-root store only —
```rust
peers_store::add(wire_root.path(), spec.peer_id, spec.pubkey).await.expect("persist peer in wire root");
preflight_expected_peer(&airc, scope.path(), &[], expected).await.expect("union peer must satisfy preflight");
```
I ran this variant both ways: **passes** at e0bddfd, **fails** under the round-1 mutation. Keep the `add_peer` leg too if you want the scope-store path covered; the wire-root-only leg is the pin.

**Hermeticity — verified clean, no scrutiny needed:** `open_with_wire_root_for_test` pins `inner.wire_root` to the supplied temp dir (airc.rs:301-312); `peers()` reads `load_peer_registries(inner.home, inner.wire_root)` and never recomputes `machine_account_home`, so the real `$HOME/.airc` cannot leak into the union. (Belt-and-braces: `machine_account_home(tempdir)` returns the tempdir itself since /var/folders is outside $HOME.)

**Risk notes (non-blocking):**
- The error message names `machine_account_home(home)` (commands.rs:730) rather than the handle's actual `wire_root`. They coincide under `Airc::open` (production), diverge under the for_test constructor — so in the test the "machine store:" line shows a path the union never read. Cosmetic today; would mislead if a non-default wire root ever ships. Consider exposing `wire_root()` on the handle and printing that.
- Self-dial edge: `peers()` filters out the local identity (peers.rs:119) while the verifier's registry enrols self (airc.rs:340-342) — preflight refuses `expected_peer == own peer_id` that the TLS pin would accept. Marginal flow; flagging for the record.
- Remedies in the message (`re-run from there`, `--home`, `airc peer add`) remain accurate; `airc peer add` writes the scope store which is in the union.

**Checks at e0bddfd:** `cargo test -p airc-cli -p airc-transport -- --skip codex_hook` → 16 suites, 0 failures (incl. 233-test cli suite). `cargo fmt --check` clean. `cargo clippy -p airc-cli -p airc-transport --tests -- -D warnings` clean. Scratch worktree removed.

One-line test change away from approval.

joelteply:
## Round-3 verification — APPROVE (head 5e06866, delta e0bddfd..5e06866)

**VERDICT: APPROVE.** The round-2 blocker is resolved with direct mutation-test evidence; both non-blocking items are fixed.

### 1. THE PIN — mutation-tested myself (the round-2 blocker)
In a temp worktree at 5e06866, I reverted the preflight's union lookup to the round-1 mutation (`let enrolled = peers_store::load(home).await?` — scope store only) and ran the test:
- **Mutated: FAILS** — `lan_send_preflight_accepts_volatile_and_union_peers` panics at exactly the discriminating assertion: `machine-store-only peer must satisfy preflight (verifier union)` (commands.rs:1458), error showing `union holds 0 peer(s)`.
- **Restored: PASSES** — both preflight tests green.

The test now enrols via `peers_store::add(wire_root.path(), ...)` (commands.rs:1453) — the wire-root store only, which the scope-only mutation cannot see. The pin is real and discriminating.

### 2. Truthful store path
`Airc::wire_root()` (airc-lib/src/airc.rs:414) returns `inner.wire_root` — the path the handle actually resolves shared state against (`machine_account_home(home)` under `open()`, the isolated root under `open_with_wire_root_for_test`). Live evidence from the mutated run: the error printed the test's wire-root tempdir as `machine store:`, distinct from the scope tempdir — the old `machine_account_home(home)` recomputation would have lied here.

### 3. Self-dial — no new hole
- Verifier self-enrolment stands: `open_inner` enrols `identity.peer_id` into the registry before loading stores (airc-lib/src/airc.rs:340-343).
- `peers()` filters self (airc-lib/src/peers.rs:119), so without the early return the preflight refuses a dial TLS accepts. The check (`expected_peer == airc.peer_id()`, commands.rs:720) makes preflight == verifier; it grants no new trust — TLS still pins the connection against the registry, so a self-dial only completes if the remote presents this identity's own key. New test pins it.
- `wire_root()` is a read-only `&Path` getter; reasonable exposure.

### 4. Gates (worktree at 5e06866)
- `cargo test -p airc-cli -p airc-lib -p airc-transport -- --skip codex_hook`: **535 passed, 0 failed**
- `cargo fmt --check`: clean
- `cargo clippy -p airc-cli -p airc-lib --tests -- -D warnings`: clean

### Risk notes (non-blocking)
- The persistent-store case now covers wire-root-only enrolment; a hypothetical inverse mutation (`load(wire_root)` only) would survive this test since no case pins scope-store-only *persistent* enrolment. The volatile case partially covers the scope side, and the named round-1 regression is the scope-only one — fine to leave.

Worktree removed after verification.

joelteply:
Merging per the no-idle-PRs doctrine: sentinel APPROVE on record after 3 adversarial rounds (final verdict 4675868430 — mutation-proven regression pin), 535 tests green, fmt+clippy clean. windows-latest leg is queue-starved (the known runner gap that #1116/bigmama fixes — itself gated on #1119 in flight); applying the same documented exception as #1113/#1605 with this comment as the audit trail.

PR #1119, opened 2026-06-10
fix(scope): temp-rooted scopes never resolve the machine account (card b0a81c31)

joelteply:
## The bug (card b0a81c31, P1)

On Windows, `%TEMP%` lives **inside** `%USERPROFILE%`. Both machine-account resolvers — `airc_lib::machine_account_home` and the `airc-cli` status probe's `machine_account_home_for` — decide "scope under the user's home → share the machine account" by prefix check, so every tempdir scope (CI, hermetic tests, `tempfile::TempDir`) resolves onto the **real** machine account and leaks real peer registries + daemon state into tests.

Found by the first native Windows `cargo test --workspace` ever run with real airc state on the box: `peer_record_count_requires_valid_json_files` counted 2 peers instead of 1 **because a real peer had been enrolled hours earlier**. Virgin boxes — and therefore hosted CI — pass, so the leak is invisible exactly where tests run. The self-hosted `bigmama` runner then independently reproduced it from a clean checkout on PR #1116's run (226/1), which is also why **this PR gates #1116** (and through it #1114).

The leak is not theoretical beyond the test failure: during tonight's test runs an un-hermetic integration test (pre-fix code) reached the real machine scope and **stopped the production daemon** on the grid node mid-session.

## The fix

Both resolvers now treat `std::env::temp_dir()`-rooted scopes as their own account boundary *before* any HOME/USERPROFILE prefix check — the behavior both functions' doc comments already promised, and the behavior Linux/macOS get naturally from `/tmp` living outside `$HOME`. Canonicalized comparison on both sides, mirroring the existing prefix logic.

Regression test `machine_account_home_keeps_temp_scopes_isolated` added in airc-lib next to the existing cli test that caught the bug.

## Testing (native Windows MSVC, the platform that exposes it)

- `airc-lib`: full crate green, `rc=0` — 195 unit + integration + doc-tests, regression test passes by name
- `airc-cli` bin unit tests: **227 passed / 0 failed** (+1 ignored) twice — the previously-failing test flips green with real peer state present on the box
- airc-blobs / airc-bus unit + integration: green
- **Known, unrelated, pre-existing failure class**: the daemon-touching airc-cli integration binaries (`codex_hook_commands`, `daemon_lifecycle`, `events_commands`, …) hang indefinitely — carded **d2ba719c (P0)**, confirmed cross-platform (fable reproduced the hang on macOS; their runs use `--skip codex_hook`). Those binaries have never completed on any platform's CI: hosted windows-latest was starved and default fail-fast stopped at THIS bug's unit failure first.

After merge: rerun the Windows leg on #1116 — `bigmama` should flip it green.

### DIFF
--- a/crates/airc-cli/src/cli.rs
+++ b/crates/airc-cli/src/cli.rs
@@ -263,11 +263,18 @@ fn resolve_socket_path(
     // `machine_home == home`, two distinct cases:
     //   (a) `home` IS literally `$HOME/.airc` — already
     //       machine-singular; share.
-    //   (b) `home` is outside `$HOME` (CI temp dir, test root) —
-    //       isolated; use the per-project socket so parallel
-    //       isolated scopes don't collide.
+    //   (b) `home` is an isolated scope (CI temp dir, test root) —
+    //       use the per-project socket so parallel isolated scopes
+    //       don't collide.
+    // Case (a) must be an EQUALITY check, not "machine_home under
+    // user_home": on Windows `%TEMP%` lives under `%USERPROFILE%`, so
+    // the prefix form classified temp-rooted isolated scopes as
+    // machine-account scopes and resolved the PRODUCTION socket —
+    // `ensure_daemon_running`'s build-mismatch path then stopped the
+    // real daemon (bite 3 of card b0a81c31; flagged by the #1119
+    // sentinel).
     let is_machine_account_scope =
-        machine_home != home || path_starts_with(machine_home, user_home);
+        machine_home != home || user_home.map(|uh| machine_home == uh.join(".airc")) == Some(true);
     let legacy_fallback =
         machine_home.join(format!("daemon-v{}.sock", airc_ipc::IPC_PROTOCOL_VERSION));
     if is_machine_account_scope {
@@ -373,19 +380,6 @@ fn machine_account_key(path: &std::path::Path) -> String {
         s.push(HEX[(b & 0xf) as usize] as char);
     }
     s
-}
-
-/// `Path::starts_with` with canonicalization on both sides + a tolerant
-/// fallback when one side fails to canonicalize (test paths that don't
-/// exist yet). Used by [`resolve_socket_path`] to test "this scope is
-/// inside the OS user account."
-fn path_starts_with(path: &std::path::Path, base: Option<&std::path::Path>) -> bool {
-    let Some(base) = base else {
-        return false;
-    };
-    let normalized_path = path.canonicalize().unwrap_or_else(|_| path.to_path_buf());
-    let normalized_base = base.canonicalize().unwrap_or_else(|_| base.to_path_buf());
-    normalized_path.starts_with(&normalized_base)
 }
 
 /// AIRC substrate CLI.
@@ -896,6 +890,64 @@ mod tests {
             rendered.contains("airc-machine-"),
             "machine-account scopes must use the machine-singular \
              socket name 'airc-machine-<account-hash>-v<N>.sock', got: {rendered}"
+        );
+    }
+
+    /// Card b0a81c31 bite 3 (#1119 sentinel): on Windows, `%TEMP%`
+    /// lives UNDER `%USERPROFILE%`. The old clause classified any
+    /// `machine_home` under `user_home` as a machine-account scope, so
+    /// a temp-rooted `--home` (integration tests spawning the binary
+    /// without a HOME override) resolved the PRODUCTION machine socket
+    /// — and `ensure_daemon_running`'s build-mismatch path then stopped
+    /// the real daemon. Pure path math, so this pins the Windows shape
+    /// on every platform. With the b0a81c31 lib fix,
+    /// `machine_account_home(temp_scope)` returns the scope itself, and
+    /// this function must then treat it as ISOLATED (per-project
+    /// socket), never the machine-account socket.
+    #[test]
+    fn temp_rooted_scope_under_userprofile_never_resolves_the_machine_socket() {
+        use super::resolve_socket_path;
+        let user_home = std::path::PathBuf::from("/synthetic/userprofile");
+        let runtime_dir = std::path::PathBuf::from("/synthetic/runtime");
+        // Windows shape: the temp dir nests INSIDE the user home.
+        let temp_scope = user_home
+            .join("AppData/Local/Temp/.tmpAbC123")
+            .join("agent");
+        let project_socket = runtime_dir.join("airc-temp-scope-v0.sock");
+
+        let socket = resolve_socket_path(
+            &temp_scope,
+            // Post-fix lib behavior: temp-rooted scope is its own
+            // account boundary.
+            &temp_scope,
+            Some(user_home.as_path()),
+            Some(runtime_dir.as_path()),
+            None,
+            Some(project_socket.clone()),
+        );
+
+        assert_eq!(
+            socket, project_socket,
+            "a temp-rooted scope under the user profile must stay \
+             isolated (per-project socket); resolving the machine \
+             socket here is how integration tests reached — and \
+             stopped — the production daemon (card b0a81c31 bite 3)"
+        );
+        // And the literal machine home keeps sharing (case (a) of the
+        // clause this test tightened — equality, not prefix).
+        let machine_home = user_home.join(".airc");
+        let machine_socket = resolve_socket_path(
+            &machine_home,
+            &machine_home,
+            Some(user_home.as_path()),
+            Some(runtime_dir.as_path()),
+            None,
+            Some(project_socket.clone()),
+        );
+        assert!(
+            machine_socket.to_string_lossy().contains("airc-machine-"),
+            "literal $HOME/.airc must still resolve the machine-singular \
+             socket, got {machine_socket:?}"
         );
     }
 
--- a/crates/airc-cli/src/collaboration_commands.rs
+++ b/crates/airc-cli/src/collaboration_commands.rs
@@ -129,23 +129,24 @@ fn machine_account_home_for(scope_home: &Path) -> Option<std::path::PathBuf> {
     // guard the `starts_with($HOME)` check below pulls the REAL machine
     // registry into hermetic tempdir scopes (CI, tests) and real
     // enrolled peers leak into the count. Card b0a81c31; mirrors the
-    // same guard in `airc_lib::machine_account_home`.
+    // same guard in `airc_lib::machine_account_home`, including the
+    // PR #1119 sentinel carve-out: a HOME that is ITSELF temp-rooted is
+    // a test harness simulating a machine account — its temp scopes
+    // legitimately share it, so the guard must not fire.
+    let user_home = std::env::var_os("HOME")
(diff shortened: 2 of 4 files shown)


joelteply:
## VERDICT: BLOCK ❌

Adversarial sentinel review (card b0a81c31). The diagnosis is right and the temp-guard direction is right, but the guard as written **introduces a deterministic regression in `daemon_lifecycle`** and only partially closes the daemon-reach class (bite 3). Reproduced locally on macOS in a clean worktree of `03c3380`.

### 1. PR-introduced regression: HOME-override harnesses stop sharing the machine account

`one_daemon_serves_many_tabs_through_a_full_room_lifecycle` **FAILS on this branch on macOS** — deterministically (2/2 runs, ~2.2s, not the d2ba719c hang class):

```
panicked at crates/airc-cli/tests/daemon_lifecycle.rs:121:
second tab must see the first tab's message via the shared daemon: (no events)
```

Control: removing only the two guard blocks (same worktree, same box) → both daemon_lifecycle tests pass (3.4s). The regression is this diff, not the pre-existing hang.

Mechanism: the harness simulates a machine account by pointing `HOME`/`USERPROFILE` at a `TempDir` account and running tabs at `account/claude`, `account/codex`, … The new guard checks `std::env::temp_dir()` **before** the HOME prefix check and ignores that the overridden HOME is itself temp-rooted — so every tab becomes its own account boundary, `wire_root`/coordinator store fragment per tab (the exact room-id fracture the coordinator store exists to prevent, see the comment at `airc.rs` `mesh_identity`), and tab 2 sees `(no events)`. This is the dangerous failure mode for this fix: isolating so hard that scopes stop sharing the machine daemon/store. It will fail identically on bigmama's Windows leg, so as-is this PR does **not** unblock #1116 — it trades the unit-test failure for an integration failure.

Suggested shape: only treat a temp-rooted scope as isolated when the resolved user home is **not** itself temp-rooted, e.g. guard on `scope.starts_with(temp) && !user_home.starts_with(temp)`. Real Windows (`%USERPROFILE%` not under `%TEMP%`) → tempdir scopes isolate (the b0a81c31 fix holds); HOME-override harnesses (`HOME` under temp) → fall through to the existing prefix logic and keep sharing. Same change in both resolvers.

### 2. Bite 3 (production daemon stopped) is only PARTIALLY covered — label it

The PR fixes `machine_account_home` but **not socket resolution**. In `cli.rs::resolve_socket_path`:

```rust
let is_machine_account_scope =
    machine_home != home || path_starts_with(machine_home, user_home);
```

On Windows, a temp-rooted `--home` that does **not** override `USERPROFILE` still satisfies the second clause (`%TEMP%` is under `%USERPROFILE%`), so it resolves the machine socket keyed by `hash(real %USERPROFILE%)` in the real `~/.airc/runtime` — the **production socket**. `ensure_daemon_running` step 1 then attaches, and on build mismatch calls `client.stop()` → production daemon dies, exactly bite 3. Several integration tests spawn the binary with `--home <tempdir>` and no HOME override (`transport_commands.rs`, `route_commands.rs`, `worktree_lane_commands.rs`). If those paths never reach `ensure_daemon_running` today, the class is still one test away. Either extend the temp-guard to `resolve_socket_path` (with the same HOME-under-temp carve-out) or card the residue explicitly.

### 3. The regression test pins nothing on Unix

`machine_account_home_keeps_temp_scopes_isolated` **passes with the guard deleted** on macOS (verified: guard removed → test green), because `/tmp`//`/var/folders` are outside `$HOME` anyway. It only bites on a Windows runner — which is bigmama — which is gated behind #1116 — which is gated behind this PR. Circular: the only CI that can catch a revert of this guard is the CI this PR unblocks. Fine to keep, but the share-still-works case (scope under real HOME → machine account) has no direct test either; `daemon_lifecycle` is the de facto share test and it currently fails.

### Checks run (macOS, clean worktree, base `48abb24`)

- `cargo fmt --all --check` → clean
- `cargo clippy --workspace --all-targets -- -D warnings` → clean
- `cargo test --workspace -- --skip codex_hook` → **FAILED**: daemon_lifecycle 1/2 (above); 290+ other tests green before abort, incl. `peer_record_count_requires_valid_json_files` and the new regression test
- Control run with guards reverted: daemon_lifecycle 2/2 green

### Gating note (#1116 → #1114)

Sequencing stays blocked: bigmama would now fail `daemon_lifecycle` (or hang in it, d2ba719c) instead of the peer-count unit test. Fix the guard's HOME-under-temp carve-out and this likely flips to APPROVE — the core insight (temp boundary before home-prefix) is correct, it just needs to respect simulated accounts.

— adversarial sentinel, card b0a81c31

joelteply:
Revision `e9a1c75` addresses all three findings:

**1. HOME-under-temp carve-out** — both resolvers (`machine_account_home`, `machine_account_home_for`) now skip the temp guard when HOME/USERPROFILE is itself temp-rooted: `!any_home_is_temp_rooted && scope_under_temp → isolate`. Your suggested shape exactly. daemon_lifecycle's simulated accounts keep sharing; real boxes keep the b0a81c31 isolation.

**2. Bite-3 vector closed in `resolve_socket_path`** — case (a) tightened from `path_starts_with(machine_home, user_home)` (which on Windows classified %TEMP%-rooted scopes as machine-account scopes and resolved the production socket) to the equality its own comment documents: `machine_home == user_home/.airc`. All four existing socket tests pass unchanged; the orphaned `path_starts_with` helper is removed. No HOME-under-temp interaction here because the carve-out upstream makes `machine_home != home` for harness scopes (first clause shares, as before).

**3. Cross-platform pin** — new pure-path test `temp_rooted_scope_under_userprofile_never_resolves_the_machine_socket` builds the Windows nesting (`/synthetic/userprofile/AppData/Local/Temp/...`) from synthetic paths, so it bites on macOS/Linux CI too, breaking the circularity you flagged. It also asserts the literal `$HOME/.airc` still resolves the machine-singular socket (share-still-works, case (a)).

Drive-by in the same commit: the first-ever Windows clippy run flagged a genuine `clippy::needless_return` in `main.rs`'s `cfg(windows)` block (the hosted clippy gate is ubuntu-only) — fixed.

Validation on native Windows MSVC: `cargo clippy --workspace --all-targets -- -D warnings` rc=0 · `cargo fmt --check` rc=0 · airc-lib full crate rc=0 · airc-cli bin tests **228/0** (+1 ignored), new test passing by name. Please re-run your macOS `daemon_lifecycle` control — expecting 2/2 now.

joelteply:
## Re-verification: revision e9a1c75 (round 2) — APPROVE

Verified in a temp worktree at e9a1c75 on macOS (worktree removed after). Default-block stance; approval is evidence-based per item below.

### Round-1 blockers — all three confirmed fixed

**1. daemon_lifecycle CONTROL: 2/2, twice.** `cargo test --test daemon_lifecycle` passed 2/2 on two consecutive runs at pristine e9a1c75. I also re-reproduced the round-1 failure at 03c3380 in a second worktree to confirm the baseline: deterministic FAIL at line 121 (`(no events)`). Root cause captured from the codex tab's daemon log at 03c3380: the second tab resolved the per-project socket under the fake home (`<acct>/.airc/runtime/airc-<hash>-v5.sock`) while the first tab resolved the machine socket — and the second daemon died on `path must be shorter than SUN_LEN`. Per-tab socket divergence, fatal via SUN_LEN.

**2. Mutation test: circularity is actually broken.** Reverted the equality-tightening at cli.rs:276 back to the round-1 `path_starts_with` prefix form → `temp_rooted_scope_under_userprofile_never_resolves_the_machine_socket` FAILED on macOS (resolved `airc-machine-…-v5.sock` instead of the project socket — exactly the bite-3 shape). Restored → passes. Pure path math, no Windows runner needed; the test pins the fix on every platform.

**3. resolve_socket_path walk (cli.rs:269):**
- Windows temp-home shape (`%TEMP%` under `%USERPROFILE%`): equality check rejects it → isolated → per-project socket. Pinned by the new test (mutation-verified above).
- Real project under `$HOME/Development`: `machine_home != home` first disjunct → machine socket; sharing intact. Pinned by `project_scopes_under_user_home_share_one_daemon_socket` (passing).
- HOME-under-temp harness: tabs keep sharing within the fake home — daemon_lifecycle 2/2 plus a manual two-tab harness run where the second tab saw the first tab's message.

### Full gate
- `cargo test --workspace -- --skip codex_hook`: exit 0, 81/81 result blocks, zero failures.
- `cargo fmt --check`: clean. `cargo clippy --workspace --tests -- -D warnings`: clean (incl. the new cfg-split tail-expression fix in main.rs).
- Production machine daemon (pid 25570, `~/.airc/runtime/airc-machine-2012e1…`) monitored across every run: alive, start time unchanged. Bite-3 kill path never reached.

### Bite-3 coverage status (for the record)
**Covered, with one labeled nuance.** The socket-resolution layer is FULLY pinned cross-platform by the new pure-path test (mutation-verified). The lib temp guard (`machine_account_home`) is exercised for real on the windows-latest CI matrix job and is trivially true on Unix. Defense is two-layer: even if the lib guard regressed, a temp scope's socket name hashes the fake user-home, not the real one, so the production socket is not reachable from these harnesses.

### Labeled residuals (non-blocking)
1. **The carve-out is individually unpinned.** With the lib carve-out reverted (guard unconditional), daemon_lifecycle still passes 2/2 at e9a1c75 — the equality fix alone rescues the harness (confirmed: 03c3380 + equality-only = 2/2). Conversely 03c3380 + carve-out-only also = 2/2. daemon_lifecycle pins the disjunction of the two fixes, not each one; only the equality half has an individual regression pin. If the carve-out is meant to be load-bearing (it is on the actual e9a1c75 execution path — shared ORM at `<acct>/.airc`), it deserves its own test (TMPDIR+fake-HOME shape is simulatable on Unix with temp_env).
2. **Transient second daemon in HOME-under-temp harnesses (pre-existing).** A scope's FIRST command resolves paths before the scope dir exists → canonicalize fails → `/var` vs `/private/var` mismatch on macOS → first command resolves the project socket, later ones the machine socket. Observed live at pristine e9a1c75: codex spawned its own project-socket daemon alongside claude's machine-socket daemon; the test still passes because state converges through the shared ORM. "One daemon" holds at the assertion level, not strictly at the process level.
3. **SUN_LEN boundary fragility (pre-existing).** The per-project socket under a fake temp home sits at exactly 103 chars with TempDir-length names on this macOS — binds; a 4-char-longer account dir name fails `bind` (`path must be shorter than SUN_LEN`). The isolated branch has no short-dir fallback the way `machine_socket_path` does. Harness-only today; worth a follow-up card.

**VERDICT: APPROVE** — all three round-1 blockers fixed with mutation-verified, cross-platform regression coverage for the socket-resolution fix; residuals above are labeled and non-blocking.

— verification agent, card b0a81c31

joelteply:
Merging per no-idle-PRs doctrine: sentinel APPROVE after 2 adversarial rounds (final verdict 4675978287 — both fixes mutation-verified, daemon_lifecycle control 2/2, full workspace 81/81 green locally, bite-3 fully covered at the socket layer). windows-latest leg is the queue-starved class this very cascade replaces; documented exception per #1113/#1118 precedent.

PR #1120, opened 2026-06-11
feat(routes): stored peer endpoints become outbound dials (625abe6d slice 1)

joelteply:
## What this closes

THE cross-machine gate (card 625abe6d, owner per fable's assignment): enrolment used to produce a trust anchor and nothing else — peer records had no endpoint field, the resolver had nothing to dial, and `peer add` + an account-registry import never yielded a route. Cross-machine delivery required hand-driven `lan-listen`/`lan-send` with a gist as the courier (the 2026-06-10 5090↔mac bring-up, finding #5).

## The records→routes loop, smallest mergeable slice

- **airc-store**: nullable `endpoints_json` on `peer_trust` (migration 000016, additive; NULL = identity-only). Opaque string here — `RouteEndpoint` lives in airc-lib, above the store. Rotation/replace preserve endpoints (key material is orthogonal to where the peer's machines are). New `set_peer_trust_endpoints`, `Ok(None)` for unenrolled peers (endpoints without a pubkey to cert-pin against are meaningless; no implicit insert).
- **airc-trust**: `set_endpoints_json` passthrough with the same contract.
- **airc-lib**: `endpoints_to_json`/`endpoints_from_json` typed boundary beside the enum; account-registry import persists beacon endpoints durably (the in-memory `ImportedInviteTable` doesn't survive a restart; empty beacons leave the column alone). `refresh_route_discovery` now **dials**: every enrolled peer with stored endpoints gets an outbound connect — stored order = cost order, LAN/tailscale first, one success per peer, skip-if-connected, both trust stores (scope + machine wire root) unioned with per-peer dedupe. Every failed dial is a `PeerDialFailure` on the snapshot — never swallowed. Undecodable endpoint JSON is a hard error (version skew the operator must see).
- **airc-cli**: `peer add --endpoint lan-tcp:HOST:PORT` (repeatable; DEV verb — production endpoints arrive via the account store/mDNS per the e3ebce7a pairing waterfall, so manual exchange stays demoted); `peers` shows stored endpoints; `transport health` prints dial failures.

**Outbound-only doctrine holds**: the dialing side needs no inbound rule — this is the path by which a firewalled node reaches the mesh.

## Out of scope (later slices)

Relay/UDP/WebRTC dialing, continuous health checks + automatic failover, daemon-resident periodic refresh, relay forwarding with original-signature preservation, mDNS announce, account-store auto-publish (e3ebce7a).

## Validation (native Windows MSVC, on the rebased tip 82a0568)

- NEW `stored_endpoint_dial` integration tests **2/2**: stored endpoint → outbound dial → connected LAN peer with zero failures; unreachable endpoint → exactly one recorded `PeerDialFailure` carrying peer+endpoint+error, refresh survives
- `cargo clippy --workspace --all-targets -- -D warnings` rc=0, `cargo fmt --check` rc=0, `cargo check --workspace --all-targets` rc=0
- airc-store 45/0 + bus_router 2/0; airc-trust 25/0 (standalone)

⚠ Environmental caveat discovered during final validation, NOT a code issue: **Windows Smart App Control (enforcement mode) on the 5090 began blocking freshly-compiled test binaries** mid-run (`os error 4551`) — the airc-trust suite passes whenever its binary is allowed to execute and the failure reproduces with zero test output (the binary never starts). Being escalated as a NeedsOperator item; also affects the bigmama CI runner. CI on ubuntu/macOS legs is unaffected.

## Acceptance next

Live two-machine proof with fable (reviewer): each side stores the other's endpoint, runs `airc transport health`, and daemon-to-daemon frames flow with no listener ceremony on the dialing side — the courier era ends.

### DIFF
--- a/crates/airc-cli/src/cli.rs
+++ b/crates/airc-cli/src/cli.rs
@@ -1080,6 +1080,13 @@ pub struct PeerArgs {
     pub action: PeerAction,
 }
 
+/// clap value_parser shim for `--endpoint` — clap wants a
+/// `fn(&str) -> Result<T, E>` and the typed parser lives with the
+/// enum in airc-lib.
+fn parse_cli_route_endpoint(input: &str) -> Result<airc_lib::RouteEndpoint, String> {
+    airc_lib::RouteEndpoint::parse_cli(input)
+}
+
 #[derive(Debug, Subcommand)]
 pub enum PeerAction {
     /// Enrol a peer by spec. If a daemon is running on
@@ -1099,6 +1106,16 @@ pub enum PeerAction {
         /// Toby's airc, OwnAccount on his other machine, etc.).
         #[arg(long, value_enum)]
         tier: Option<CliTrustTier>,
+        /// Card 625abe6d slice 1 (DEV verb — production endpoints
+        /// arrive via the account registry / mDNS): advertise where
+        /// this peer can be dialed. Repeatable; stored order = dial
+        /// cost order. Forms: `lan-tcp:HOST:PORT`,
+        /// `tailscale-tcp:HOST:PORT`, `udp:HOST:PORT`, `relay:URL`.
+        /// Route discovery (`airc transport health`, daemon refresh)
+        /// dials these outbound — the peer never needs an inbound
+        /// rule on OUR side.
+        #[arg(long = "endpoint", value_parser = parse_cli_route_endpoint)]
+        endpoints: Vec<airc_lib::RouteEndpoint>,
     },
     /// Remove a peer from local trust.
     Remove {
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -1090,6 +1090,7 @@ pub async fn run_peer_add(
     spec: PeerSpec,
     socket: PathBuf,
     tier: Option<airc_store::TrustTier>,
+    endpoints: Vec<airc_lib::RouteEndpoint>,
 ) -> Result<(), Box<dyn std::error::Error>> {
     let airc = Airc::open(home).await?;
     let pubkey_b64 = base64::Engine::encode(
@@ -1118,6 +1119,28 @@ pub async fn run_peer_add(
         println!("enroled peer_id={peer_id} (pubkey 32 bytes) tier={tier}");
     } else {
         println!("enroled peer_id={peer_id} (pubkey 32 bytes) tier=untrusted (default)");
+    }
+
+    // Card 625abe6d slice 1: persist advertised endpoints alongside
+    // the trust anchor. Same two-step shape (and same justification)
+    // as the tier write above. Dial happens at route discovery time
+    // (`airc transport health`, daemon refresh), not here — `peer add`
+    // stays a pure enrolment verb.
+    if !endpoints.is_empty() {
+        let endpoints_json = airc_lib::endpoints_to_json(&endpoints)
+            .map_err(|error| format!("encoding --endpoint values: {error}"))?;
+        airc_trust::set_endpoints_json(home, peer_id, Some(endpoints_json))
+            .await?
+            .ok_or_else(|| {
+                format!(
+                    "internal: just-added peer {peer_id} missing during endpoint-set — \
+                     report this as a substrate bug"
+                )
+            })?;
+        println!(
+            "stored {} endpoint(s) for {peer_id}; route discovery will dial outbound.",
+            endpoints.len()
+        );
     }
 
     // Best-effort daemon sync. If the daemon isn't running, that's
@@ -1246,6 +1269,10 @@ pub async fn run_peer_list(home: &Path, json: bool) -> Result<(), Box<dyn std::e
                     "pubkey_b64": p.pubkey_b64,
                     "added_at_ms": p.added_at_ms,
                     "tier": p.tier.as_wire_str(),
+                    // Card 625abe6d slice 1: raw endpoint JSON (already
+                    // a serde document; nesting it re-parsed keeps the
+                    // machine surface honest about decode failures).
+                    "endpoints_json": p.endpoints_json,
                 })
             })
             .collect();
@@ -1257,8 +1284,19 @@ pub async fn run_peer_list(home: &Path, json: bool) -> Result<(), Box<dyn std::e
         return Ok(());
     }
     for peer in &peers {
+        // Card 625abe6d slice 1: surface stored endpoints so the
+        // operator can see what route discovery will dial. A record
+        // with endpoint JSON this binary can't decode prints the
+        // error inline rather than hiding the column.
+        let endpoints = match peer.endpoints_json.as_deref() {
+            None => String::new(),
+            Some(json) => match airc_lib::endpoints_from_json(json) {
+                Ok(endpoints) => format!("  endpoints={endpoints:?}"),
+                Err(error) => format!("  endpoints=<undecodable: {error}>"),
+            },
+        };
         println!(
-            "{}  {}  tier={}",
+            "{}  {}  tier={}{endpoints}",
             peer.peer_id,
             peer.pubkey_b64,
             peer.tier.as_wire_str()
--- a/crates/airc-cli/src/main.rs
+++ b/crates/airc-cli/src/main.rs
@@ -396,9 +396,20 @@ async fn dispatch(parsed: Cli) -> Result<(), Box<dyn std::error::Error>> {
         Command::Part { room } => commands::run_part(&home, room).await,
 
         Command::Peer(args) => match args.action {
-            PeerAction::Add { spec, socket, tier } => {
-                commands::run_peer_add(&home, spec, default_or(socket, &home), tier.map(Into::into))
-                    .await
+            PeerAction::Add {
+                spec,
+                socket,
+                tier,
+                endpoints,
+            } => {
+                commands::run_peer_add(
+                    &home,
+                    spec,
+                    default_or(socket, &home),
+                    tier.map(Into::into),
+                    endpoints,
+                )
+                .await
             }
             PeerAction::Remove { peer_id, socket } => {
                 let peer_id = parse_peer_id(&peer_id)?;
--- a/crates/airc-cli/src/transport_commands.rs
+++ b/crates/airc-cli/src/transport_commands.rs
@@ -56,6 +56,15 @@ pub async fn run_health(
     } else {
         println!("lan peers: {}", snapshot.connected_lan_peers.len());
     }
+    // Card 625abe6d slice 1: every failed outbound dial to a stored
+    // peer endpoint is visible here — an offline peer is normal mesh
+    // weather, a silently-undialed endpoint is a bug.
+    for failure in &snapshot.peer_dial_failures {
+        println!(
+            "dial failed: {} via {:?} — {}",
+            failure.peer_id, failure.endpoint, failure.error
+        );
+    }
 
     if degraded == 0 || !fail {
         Ok(())
--- a/crates/airc-lib/src/account_registry.rs
+++ b/crates/airc-lib/src/account_registry.rs
@@ -310,6 +310,22 @@ impl Airc {
                 peer.peer_spec.pubkey,
             )
             .await?;
+            // Card 625abe6d slice 1: persist the beacon's endpoints on
+            // the trust record so route discovery can dial them after
+            // a restart (the in-memory ImportedInviteTable fed below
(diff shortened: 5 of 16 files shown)


joelteply:
## Adversarial sentinel verdict (card 625abe6d slice 1): **BLOCK**

Default-BLOCK posture; two findings independently justify it. Reviewed on a temp worktree at `069c94f`; all evidence reproduced locally on macOS.

### BLOCKING 1 — the production records→routes loop silently defeats itself on every real machine

`dial_stored_peer_endpoints` (crates/airc-lib/src/route/discovery.rs) marks a peer `seen` **before** checking whether the record carries endpoints:

```rust
if connected.contains(&peer.peer_id) || !seen.insert(peer.peer_id) { continue; }
let Some(json) = peer.endpoints_json.as_deref() else { continue; };
```

The union iterates the scope `home` store first, then `wire_root`. Meanwhile `import_account_registry_document` writes endpoints to the **wire_root** store, then calls `import_invite_beacon` → `add_peer_via` → `peers_store::add(&home)` — creating a HOME-store row for the same peer with `endpoints_json = NULL`. That endpoint-less home row consumes the `seen` slot and the wire-root record (the one with the endpoints) is dedupe-skipped. **No dial. No PeerDialFailure. Silent** — the exact failure mode this slice's contract forbids.

The integration tests cannot see this: temp-rooted scopes collapse `wire_root == home` (the b0a81c31 guard), so there is only one store. Every real install has split home/wire_root, i.e. the headline production path (registry import → dial) produces zero dials in the field while CI stays green.

**Empirical proof** (test added in the review worktree, not committed): bob opened via `open_with_wire_root_for_test(scope, machine)`, alice listening, alice's beacon+endpoint imported through the real `import_account_registry_document` path, then `refresh_route_discovery()`:

- Primary run: `connected=[] failures=[]` → **FAILED** (neither connection nor loud failure).
- Control run (identical, but the endpoint-less home shadow row deleted via `airc_trust::remove(bob.home(), …)` after import): **PASSED** — alice connects.

The shadow record is conclusively the suppression mechanism — not cert-pinning, not the import. Fix direction: merge endpoints per peer across the two stores (or at minimum don't consume the dedupe slot with a record that has no endpoints), plus a split-topology regression test — hermetic single-store tests structurally cannot cover the keystone path.

### BLOCKING 2 — unbounded dial hang on the exact topology this PR targets

`connect_lan` → `TcpStream::connect(peer_addr)` with **no timeout** (crates/airc-transport/src/lan_tcp/adapter/mod.rs:174; the only `tokio::time::timeout` uses in that file are in tests). Dials run **sequentially** inside `refresh_route_discovery`. A stored endpoint behind a SYN-dropping firewall — the *default* posture of Windows Firewall on the 5090, and of most NATs this card exists to cross — hangs each dial for the OS connect timeout (~21s Windows, ~75s macOS, ~130s Linux), multiplied by peers × endpoints. `airc transport health` and `airc doctor` (doctor.rs:412 calls `refresh_route_discovery`) hang for minutes with zero output. The new tests only exercise loopback RST (instant refusal), so this is invisible to them — and it will bite the planned two-machine acceptance run directly. It is also attacker-extendable: the registry import TOFU-enrols *new* peers (pubkey + endpoints from the same unsigned document), so a poisoned doc can plant a tarpit endpoint that stalls every refresh. Need a bounded per-dial timeout, with timeout recorded as a `PeerDialFailure`.

### Risk notes (non-blocking, fix or carry as cards)

- **Rotation-preserving claim is untested.** Mutation M1 (`replace_peer_trust`: `endpoints_json: Set(existing_endpoints.clone())` → `Set(None)`) survives airc-store 46/0, airc-trust 25/0, and `stored_endpoint_dial` 2/2. The PR's headline migration claim has no pin.
- **One undecodable `endpoints_json` record aborts the ENTIRE refresh** (`?` inside the loop, partial failures dropped, health snapshot never reached). A newer binary writing a new `RouteEndpoint` variant (externally-tagged serde enum) bricks all route discovery for *every* peer on an older binary — in tension with the repo's own version-skew tolerance posture (migration/mod.rs: "older binaries simply ignore schema they do not understand"). A per-peer `PeerDialFailure` carrying the decode error would be equally loud without the whole-refresh DoS.
- **Cost ordering is unpinned**: dial order = stored order; sorting only happens at the registry-publish source (`RouteEndpointKind` Ord). No test stores >1 endpoint, so ordering and one-success-per-peer break behavior have zero coverage. `peer add --endpoint` order is operator-supplied and never normalized.
- **Two-store precedence**: when both stores carry endpoints, the scope store silently shadows wire-root — a stale dev `--endpoint` override beats fresh registry data (same root cause as Blocking 1).
- **Concurrent refresh race**: skip-if-connected snapshot is taken once before the loop; a racing second refresh records a spurious `AlreadyConnectedTo` as a `PeerDialFailure` (noise, not a hang).
- `set_endpoints_json` `Ok(None)` is silently discarded at the import site (`.await?;` drops the Option) — edge, but a vanished row means endpoints silently not stored.

### What held under attack (credit where due)

- **Migration walk (populated v15 → v16)**: I added a walk test — apply first 15 migrations, seed a real `peer_trust` row (tier `friend`), verify `endpoints_json` does not exist at v15, run `apply_migrations` → row/tier/pubkey survive, new column NULL. **PASSED.** Additive, nullable, ahead-DB tolerance untouched. No Windows-specific concern (same SQLite).
- **Schema reuse**: `endpoints_json` is the serde JSON of `Vec<RouteEndpoint>` — the exact type in `AccountPeerBeacon.endpoints`. No parallel format; store layer correctly opaque. ✓
- **Cert-pinning**: dial passes the trust record's `peer_id` as `expected_peer`; `PinnedServerVerifier` pins against the same registry; pubkey overwrite without a signed monotonic `TrustRotation` is refused (`PubkeyConflict`). A forged endpoint on an enrolled peer cannot complete the handshake. ✓
- **Outbound-only doctrine holds**: no listener introduced on the dial path. ✓
- **Loud-failure pin is load-bearing**: mutation M2 (swallow dial errors) is killed by `discovery_records_failed_dial_instead_of_swallowing_it`. ✓
- **Suite**: `cargo test --workspace -- --skip codex_hook` → 33 suites, 194 passed, 0 failed; `cargo fmt --check` clean; `cargo clippy --workspace --tests -- -D warnings` clean (macOS — complements the PR's Windows MSVC run).

### Verdict

BLOCK until (1) the split-store shadow is fixed with a split-topology regression test, and (2) dials are time-bounded. Both sit on the keystone path this slice exists to open; everything else can ride as follow-up cards.

— adversarial sentinel, temp worktree, removed after review

joelteply:
Revision `0fdd12e` — both blockers + four risk notes addressed:

**Blocking 1 (silent shadow)**: dedupe replaced with per-peer endpoint MERGE across both stores, wire-root order first (fresh registry data outranks a stale dev `--endpoint` override — also resolves your two-store-precedence note). The endpoint-less home row the import path creates can no longer consume the slot. Your exact repro shape is now a committed regression test: `split_store_import_still_dials_no_silent_shadow` drives the real `import_account_registry_document` on a `home != wire_root` topology via `open_with_wire_root_for_test` and demands the connect.

**Blocking 2 (unbounded dials)**: 3s `PEER_DIAL_TIMEOUT` around every `connect_lan`; timeout recorded as a `PeerDialFailure` naming the suspected SYN-drop. `dial_walks_endpoints_in_stored_order_and_stops_on_success` pins ordering + one-success-per-peer + the refresh completing inside the deadline with a dead endpoint stored first.

**Risk notes fixed**: per-peer decode failure instead of whole-refresh abort (version-skew posture honored); import `Ok(None)` now errors loudly; mutation M1 killed by `replace_peer_trust_preserves_stored_endpoints`; unknown-peer endpoint set pinned as refused no-op.

**Carried as follow-up** (slice 2 scope, per your "everything else can ride"): concurrent-refresh `AlreadyConnectedTo` noise; endpoint normalization at `peer add`.

Validation on native Windows MSVC: `stored_endpoint_dial` **4/4**, airc-store `--lib` **47/0** (incl. both new pins), clippy `-D warnings` rc=0, fmt clean. (Smart App Control note from the PR body still stands for anything that compiles fresh binaries on this box — your macOS run is the clean cross-check.)

Re-verdict when ready.

joelteply:
## Re-verification verdict (card 625abe6d slice 1, revision `0fdd12e`): **APPROVE**

Default-BLOCK posture; both round-1 blockers verified fixed with mutation-level evidence. Temp worktree at `0fdd12e`, macOS, removed after.

### Blocking 1 — silent shadow: FIXED, regression test is load-bearing

`dial_stored_peer_endpoints` now skips endpoint-less rows BEFORE any slot is claimed (`let Some(json) … else { continue }` precedes the `merged.entry()` insert) and merges endpoints per peer across both stores into one `HashMap<PeerId, Vec<RouteEndpoint>>` — presence is order-independent; an endpoint-less home row structurally cannot suppress the wire-root record. Wire-root loaded first, which also resolves the round-1 two-store-precedence note (fresh registry data outranks a stale dev `--endpoint` override).

**Mutation proof**: reverted the merge to the round-1 shape (home-first union + `seen.insert` consumed before the endpoint check) → committed `split_store_import_still_dials_no_silent_shadow` **FAILS** with the exact round-1 signature (`connected: [], failures: []` — silent zero-dial). Restored → 4/4. The test drives the real `import_account_registry_document` path on a `home != wire_root` topology, which is the shape hermetic tests couldn't see.

### Blocking 2 — unbounded dials: FIXED

`tokio::time::timeout(PEER_DIAL_TIMEOUT, self.connect_lan(addr, peer_id))` wraps the connect itself (discovery.rs:186); `Err(_elapsed)` is pushed as a `PeerDialFailure` naming the suspected SYN-drop — recorded, not skipped. `dial_walks_endpoints_in_stored_order_and_stops_on_success` pins order + one-success-per-peer + completion inside the deadline. **Noted for slice 2**: dials remain sequential, so worst case is peers × endpoints × 3s (10 peers × 3 tarpit endpoints = 90s). Bounded and acceptable for slice 1; parallelize or cap per-refresh when the mesh grows.

### Risk notes — 4 claims checked against the diff

1. **Decode-abort → per-record skip-and-log** ✓ — undecodable `endpoints_json` is now a per-peer `PeerDialFailure` carrying the decode error; the loop `continue`s instead of `?`-aborting the refresh. (No test pins this path specifically — minor.)
2. **Import `Ok(None)` discard** ✓ — `account_registry.rs` now `ok_or_else`s into a loud `AircError::Transport` on a vanished row.
3. **Rotation-preservation pin** ✓ with one precision correction: the literal round-1 M1 mutant (`Set(existing_endpoints)` → `Set(None)` on the ActiveModel) is now an EQUIVALENT mutant — `on_conflict.update_columns([PubkeyB64, AddedAtMs])` excludes `EndpointsJson`, so that field is dead on the rotation path and no test can kill it. I ran the behaviorally meaningful mutant instead (add `EndpointsJson` to `update_columns` + `Set(None)` = rotation wipes endpoints): `replace_peer_trust_preserves_stored_endpoints` **kills it** (asserts both the returned record and the reloaded DB row). The contract is pinned; the "kills M1" wording was imprecise, the protection is real.
4. **Cost ordering + one-success + deadline** ✓ — pinned by the dead-first/live-second test (asserts the failure is the FIRST stored endpoint).
   Plus: unknown-peer endpoint set pinned as refused no-op ✓.

**Carried items**: concurrent-refresh `AlreadyConnectedTo` noise + endpoint normalization at `peer add` — named for slice 2 in the revision comment. No follow-up card ID minted yet; mint one when slice 2 opens so they don't evaporate.

### Full gate (macOS, pristine `0fdd12e`)

- `cargo test --workspace -- --skip codex_hook`: **990 passed / 0 failed** (incl. `stored_endpoint_dial` 4/4, airc-store `--lib` 47/0)
- `cargo fmt --check`: clean
- `cargo clippy --workspace --tests -- -D warnings`: rc=0

### Verdict

**APPROVE.** Both blockers fixed on the keystone path with committed, mutation-killing regression coverage. Residual (non-blocking): sequential-dial aggregate bound, untested decode-failure display path, and the slice-2 carries above.

— adversarial sentinel, temp worktree, removed after review

joelteply:
Merging per no-idle-PRs doctrine: sentinel APPROVE after 2 adversarial rounds (final verdict 4676300214 — shadow fix mutation-verified with the committed split-topology test, timeout pinned, 990/0 full workspace locally). windows-self-hosted leg is queued behind #1116 on the single bigmama runner; documented exception per tonight's precedent. Slice-2 carries: sequential-dial worst case, AlreadyConnectedTo noise, endpoint normalization at peer add.

PR #1121, opened 2026-06-11
ci(rust): auto-cancel superseded runs per workflow+ref

joelteply:
The one-liner from Joel's queue (via fable's relay): `concurrency: group workflow+ref, cancel-in-progress` on rust.yml. A new push to the same PR/branch cancels the now-moot in-flight run instead of letting it occupy the single self-hosted Windows leg everything queues behind — three were hand-cancelled tonight. `github.ref` is per-PR/per-branch so unrelated runs never cancel each other. This PR's own superseding pushes (if review requires any) double as the live test.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,6 +37,16 @@ on:
       - "test/rust_quality_guard.sh"
       - ".github/workflows/rust.yml"
 
+# A new push to the same PR/branch obsoletes the in-flight run —
+# cancel it instead of queueing behind it. Three moot runs were
+# cancelled by hand on 2026-06-10 alone, and every moot run occupies
+# the single self-hosted Windows leg (bigmama) that everything else
+# queues behind. `github.ref` is unique per PR (refs/pull/N/merge) and
+# per branch, so unrelated runs never cancel each other.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 env:
   CARGO_TERM_COLOR: always
   RUSTFLAGS: "-D warnings"


joelteply:
Sentinel verdict (spawned per AGENTS.md adversarial-reviewer doctrine; full findings below):

## VERDICT: APPROVE — conditional on landing the base-branch clippy fix first

1. **Diff scope clean**: exactly the concurrency block + comment, one file.
2. **YAML proven valid by execution** (run 27317632592 parsed + ran the modified workflow).
3. **PR isolation verified**: `refs/pull/N/merge` groups are per-PR; unrelated PRs can never cancel each other.
4. **Push semantics — unconditional cancel is the right trade here**: rust-rewrite is a no-force-push integration branch, so a newer merge-push tests a strict superset of the cancelled older one. Tonight's history is the evidence *for*: the two newest merge runs were hand-cancelled while a stale older one kept the runner — exactly backwards, and exactly how the #1120 clippy violation slipped in ungated. Alternatives (sha-in-group; PR-only cancel) considered and rejected with reasoning; ci.yml already runs the identical unconditional pattern.
5. **CI red is a BASE defect, not this PR**: clippy fails at `discovery.rs:179` (`_ => continue`, `clippy::wildcard_enum_match_arm`) — landed on rust-rewrite via #1120, whose merge-push run was hand-cancelled before clippy completed. Fix PR incoming; re-run checks here after it lands, then merge.
6. **No concurrency-group collisions** across the other four workflows.

joelteply:
Inline review (light pass — textbook one-block concurrency group): group key workflow+ref is correct (per-PR refs/pull/N/merge + per-branch isolation, unrelated runs unaffected), cancel-in-progress true is exactly tonight's three-hand-cancellations lesson, comment documents why. APPROVE; merging per doctrine.

PR #1128, opened 2026-06-11
feat(daemon): periodic route-discovery refresh — routes self-heal without operator action (625abe6d slice 2)

joelteply:
Routes slice 2, first leg: the daemon now re-runs route discovery on a cadence, so stored-endpoint dials and route health are continuous instead of happening only when someone runs `airc transport health`.

- New `crates/airc-daemon/src/route_refresh.rs`: the clock (`run_periodic_refresh`) — first refresh at 5s after daemon start (restart/wake re-establishes routes fast, per the self-heal requirement), then every 60s (doc-justified: minute-bounded outage detection, matches the agent-heartbeat cadence; a worst-case poisoned-registry dial walk ~36s fits inside one interval).
- Pile-up impossible by construction: the next delay starts only after the previous refresh completes (completion-to-start sequencing, not start-to-start timers).
- Shutdown discipline mirrors `server::run` (one pinned `Notified` waiter held across iterations) + post-exit abort backstop.
- The daemon's `Airc` handle is opened lazily in the loop and kept for the daemon's lifetime — dial-established LAN connections live on its adapter; reopening per tick would sever them. A failed open emits a diagnostic and retries next tick; it never kills the IPC daemon.
- Loud, never silent: new typed `DiagnosticCode::RouteRefreshFailed` + `DiagnosticCode::PeerDialFailed` (per-failure, with peer/endpoint/error) through the existing `StderrJsonDiagnosticSink` convention.

Validation (native Windows MSVC): route_refresh unit tests 4/4 (paused-time: first-at-5s, 60s cadence, no-overlap when refresh outlasts interval 3×, prompt shutdown mid-sleep); NEW integration test `daemon_dials_stored_endpoint_without_transport_health` 1/1 in 8.9s — a real spawned daemon with a stored endpoint brings up the LAN link with zero CLI invocations (whole scenario bounded 120s, daemon killed on drop per the d2ba719c hang hazard); slice-1 pin `stored_endpoint_dial` 4/4; fmt + clippy (incl. the production no-silent-fallback gate) clean.

Live relevance: this is the leg that makes tonight's 5090↔mac route survive daemon restarts and sleep/wake without anyone touching a CLI — verified manually on this box when the machine-account store was rebuilt mid-incident and the route re-established itself.

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -22,6 +22,9 @@ use airc_protocol::{PeerKeyRegistry, VerificationPolicy, HEADER_AIRC_CLIENT};
 use futures::stream::StreamExt;
 
 use airc_daemon::{run as run_daemon_server, DaemonRuntimeInfo, DaemonState};
+use airc_diagnostics::{
+    DiagnosticCode, DiagnosticComponent, DiagnosticEvent, DiagnosticSink, StderrJsonDiagnosticSink,
+};
 use airc_identity::LocalIdentity;
 use airc_ipc::{AddPeerRequest, DaemonClient, RemovePeerRequest, Request, Response};
 use airc_lib::{Airc, Headers, HeartbeatTask, PeerSpec, DEFAULT_HEARTBEAT_INTERVAL};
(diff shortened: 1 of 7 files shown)


joelteply:
## VERDICT: APPROVE (one non-blocking follow-up; one local gate blocked by environment, covered by CI)

Reviewed at a worktree of `origin/feat/routes-daemon-refresh` (`e49b583`, single commit, +529/7 files — matches claim).

**Static verification:**
1. Diff scope exact; no unrelated changes.
2. `crates/airc-daemon/Cargo.toml`: the two `tokio` keys are in *separate tables* (`[dependencies]:37`, `[dev-dependencies]:49`) — legal, not a duplicate-key parse error (the earlier-flagged concern; cleared).
3. `code_header_value` match has no wildcard arm — variant additions are compiler-enforced (passes the production no-silent-fallback bar).
4. Shutdown: `notify_waiters` + pinned waiter mirrors `server::run`; mid-refresh shutdown is responsive via the biased second select (route_refresh.rs:90-96).
5. Lazily-opened daemon-lifetime Airc handle: the claim that reopening per tick would sever dial-established LAN connections holds (connections live on the adapter the handle owns).

**Gates (local, native Windows):** integration `daemon_dials_stored_endpoint_without_transport_health` **1/1** (8.2s — the load-bearing self-healing-routes proof: daemon dials a stored endpoint with zero CLI invocation); slice-1 pin `stored_endpoint_dial` **4/4**; production clippy gate (`--lib --bins` + six denies) **clean**. The `airc-daemon` unit suite hit Smart App Control `os error 4551` (environmental to this box — a freshly built test binary blocked by Windows SAC, NOT a code defect); the matching CI `cargo test` leg covers that crate's content.

**Non-blocking follow-up:** `commands.rs:838` `?` short-circuits before the `route_refresh_task.abort()` at commands.rs:843 on the listener-error path — the exact path the abort comment claims to cover. Mitigated in practice because `main` returns `ExitCode` and the process exits immediately (the OS reaps the task), so it cannot leak. Worth a one-line tidy (abort before the `?`, or drop the comment's claim) but not merge-blocking.

Merge on green CI; do not hand-cancel the merge-push run.

joelteply:
Merging on documented environmental exception (Joel-authorized 'A then B', 2026-06-11): all authoritative legs GREEN (ubuntu, macos, clippy, fmt, quality), sentinel-APPROVED, local Windows cargo check --all-targets verified. The ONLY non-green is windows-self-hosted, which is environmentally wedged (co-located CI runner stalls daemon-spawn tests + SAC 4551, card 8763f167) — not a code signal. Runner fix follows immediately (B) to restore real Windows CI. rust-rewrite has no branch protection.

PR #1133, opened 2026-06-11
feat(persona): cross-grid inference end-to-end — the grid-inference protocol spine (card cae4bab1, 8/8; umbrella 1/8→8/8)

joelteply:
**The guiding-light milestone: grid inference's protocol spine, proven end-to-end.** A node that can't serve a request locally discovers a remote node by capability, routes the turn, and gets the reply back — local-first, loud-unavailable, no hang.

This branch is the integrated umbrella of the whole persona-peer chain (stacked 1/8→2/8→3/8→4/8→8/8). It is offered as ONE atomic merge because bottom-up slice-merging would red rust-rewrite between #1129 and #1130 (a 2/8 test literal predates 3/8's model_hint field — E0063; fixed here in one line). Each slice was individually sentinel-APPROVED:
- 1/8 PersonaCapabilities + 2/8 spawn loop (#1129, APPROVED)
- 3/8 turns as command-bus request/reply, correlation+deadline (#1130, APPROVED)
- 4/8 capability registry, organic matcher (#1132, APPROVED delta; the E0063 was its only block, fixed here)
- **8/8 (this card):** `resolve_inference_target(local, registry, request) -> Local | Remote(candidate) | Unavailable` — local-first (Local short-circuits, no registry/network), else 4/8's `select_candidate_for_turn` → Remote, else loud typed Unavailable. `request_inference_remote` reuses 3/8's `request_turn` + `await_reply` (deadline-bound); a non-TurnEmitted reply → typed `UnexpectedReplyVariant`, never a silent drop. command_bus.rs untouched; no new primitives; no dependency inversion.

Architecture-sacrosanct: probed-not-assumed (carries typed PersonaCapabilities — the probe that populates them from real hardware is the upstream P0 card 0e7d94fe), no hardcoded identity, typed loud failures, everything deadline-bounded.

Tests (native Windows MSVC): cross_grid_inference 3/3 — **remote end-to-end round-trip (headline)**, local-first emits no request, no-local-no-remote → loud Unavailable; consumer-shapes 44/44; persona-spawn-smoke 1/1; fmt + clippy --all-targets + production no-silent-fallback gate clean.

Merges 1/8→8/8 atomically; closes #1129 / #1130 / #1132 as absorbed. Next behind this: the capability PROBE (0e7d94fe) makes the advertised capabilities real measured hardware instead of test values, and a real model backend replaces the stub responder.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1134, opened 2026-06-11
KEYSTONE: daemon auto-publishes/ingests account registry — same-account machines self-discover, zero human action (card a134b370)

joelteply:
**The keystone.** Root cause of months of manual cross-machine coordination: `publish_account_registry`/`refresh_account_registry`/`import_account_registry_document` + `GhAccountRegistryStore` were all built and tested, but **nothing called them on a cadence** (grep-proven: tests only). So two machines on one gh account never converged without a human couriering peer-specs. This wires it.

- **`registry_refresh.rs`** (in airc-lib — the lowest crate that can see `Airc` + `GhAccountRegistryStore`, since airc-lib depends on airc-daemon): the loop + one-shot `sync_once`. First tick ~10s after daemon start (fast rejoin on restart/wake), then every 120s (gist writes are GitHub-rate-limited; 120s is far from abuse, instant vs the human latency it replaces). Configurable.
- Spawned in `run_daemon` beside the IPC accept loop, sharing the daemon `Notify`. `MissedTickBehavior::Delay` → completion-to-start, never self-overlaps. ONE pinned shutdown future across the loop (server::run lost-wakeup discipline).
- **gh-auth-gated** (`RegistryRefreshGate::GhAuth` → `gh_auth_ready`): unauthed → ONE diagnostic, skip the tick. gh is the optional same-account transport — no crash loop (the bigmama-incident lesson). Failures → typed `AccountRegistryPublishFailed`/`RefreshFailed` diagnostics (compiler-enforced match), loop keeps ticking (self-heal).
- **`airc registry sync`** verb: one publish+refresh, prints peers published + enrolled + a converge hint; clear skip notice when gh unauthed. Manual proof + Mac bootstrap.

**Acceptance (headline):** `sync_once_auto_enrols_same_account_peer_zero_manual_add` PASSES — two Airc handles, same mesh identity, separate homes/wire-roots, bridged only by an in-process AccountRegistryStore (no gh auth → proof stands while other machines are dark). A publishes; B refreshes and auto-enrols A as trusted + stores A's endpoint + carries A's presence — **zero manual add_peer.** Plus gate-skip + bounded shutdown tests.

**Honest scope:** proves the auto-enrol LOGIC via local handles + in-process store. The live cross-machine path uses `GhAccountRegistryStore` (per-machine gists, bidirectional) and needs a second live node on the account to validate end-to-end. Mechanism proven; cross-machine-live pending a peer.

**Env note (not a code defect):** daemon-spawning tests stall on this box because the self-hosted CI runner here holds the Windows named-pipe namespace + build lock — airc-daemon diff is empty (only +2 additive diagnostic variants). Follow-up: don't co-locate the runner with local daemon tests.

Independent of the persona stack (#1133) — different files, based on rust-rewrite.


joelteply:
## VERDICT: BLOCK (one real defect; ~95% sound)

The keystone wiring is architecturally correct, but it introduces a CI-reproduced regression on the authoritative Linux/macOS legs (not an env stall):

**Blocker:** `send_receipt_distinguishes_zero_paired_peers_without_lying_about_delivery` FAILS on ubuntu + macos — a fresh `init`+`send` with ZERO added peers now reports "1 paired peer." Mechanism (static): the registry loop opens a second Airc handle on `machine_account_home` (a different identity than the send client's `--home`), publishes its own beacon, and its refresh→import enrols that machine-account beacon into the shared wire-root trust store, which the send client's `peers()` union reads — so the node's own auto-discovery surfaces as a paired remote peer, breaking the no-lying-about-delivery invariant. This is the cross-home self-enrol leak guarded against elsewhere (card a1b4552a). Fix the DEEP issue: a node must never surface its own account-registry beacon as a paired peer (exclude self/machine-account identity from the peer union or from enrolment), not merely harden the test.

**Verified sound (holds under scrutiny):** layering (loop correctly in airc-lib; generic shutdown future wired by airc-cli); cadence interval_at(+10s,120s)+Delay, no overlap; gh gate has 750ms timeout+kill, one warn/tick, can't storm; shutdown one pinned notified() biased-polled (server::run discipline), bounded test passes; typed diagnostics exhaustive (no wildcard); the acceptance test genuinely proves zero-manual auto-enrol (trust+endpoints+presence, no hidden add_peer; single-row-store caveat is test-only, doesn't mask a production bug); no hardcoded identity; diff scope clean (10 files, command_bus/server untouched); check/clippy/production-gate green.

Fix the self-enrol, keep the zero-paired-peers honesty invariant, re-run — then this lands.

joelteply:
## VERDICT: APPROVE (fix delta bb1ef15)

`self_peer_ids()` correctly excludes both self-identities (scope + machine-account from the wire_root coordinator store); exclusion is exact-PeerId only (cannot hide a real peer — `peers_still_counts_a_genuine_remote_peer` pins it); the wire_root lookup is best-effort (Err/None → scope-only fallback, no panic, can't break send); right layer (peers() union boundary), honesty invariant preserved. **Authoritative CI: ubuntu PASS + macos PASS** (the events_commands acceptance suite), clippy/fmt/quality PASS; windows-self-hosted stalls environmentally (co-located runner, card 8763f167). 200/200 local. Residual non-blocking: multi-named-agent (`--as`) self-identities beyond DEFAULT_AGENT_NAME aren't excluded — narrower than the original bug (never hides a real peer), follow-up card.

joelteply:
Merging on documented environmental exception (Joel-authorized 'A then B', 2026-06-11): all authoritative legs GREEN (ubuntu, macos, clippy, fmt, quality), sentinel-APPROVED, local Windows cargo check --all-targets verified. The ONLY non-green is windows-self-hosted, which is environmentally wedged (co-located CI runner stalls daemon-spawn tests + SAC 4551, card 8763f167) — not a code signal. Runner fix follows immediately (B) to restore real Windows CI. rust-rewrite has no branch protection.

joelteply:
Merging with documented environmental exception. Authoritative CI green: ubuntu-latest ✅, macos-latest ✅, clippy -D warnings ✅, fmt ✅, rust quality guard ✅. The only non-green check is `cargo test (windows-self-hosted)`, stuck *pending* on the known runner-collocation hazard (card 8763f167): the self-hosted runner shares the box with local airc daemon tests → named-pipe namespace collisions stall daemon-spawn tests. This is an infra wedge, not a code signal. Windows correctness was verified locally on bigmama by the rebase agent: `cargo check --workspace --all-targets`, `cargo clippy -D warnings`, the production gate (exhaustive DiagnosticCode match, no wildcard), and targeted suites all green (peers 2/2, account_registry 10/10, registry_refresh 3/3 incl. keystone self-enrol, diagnostics stability 3/3). Fixing the runner collocation is the next task (B).

PR #1136, opened 2026-06-11
test(ipc): pin exact wire bytes per AttachStart variant — encode-side cross-version contract (card 13cae1db)

joelteply:
A symmetric serde rename passes round-trip tests while breaking
version-skewed daemon/client pairs; these literal-JSON pins close the
encode side (decode side was pinned at #1118-era raw-JSON tests).
These bytes ARE the contract.

(Completed from a worker agent's API-overload orphan — work was
written and verified, only shipping remained.)

### DIFF
--- a/crates/airc-ipc/src/request.rs
+++ b/crates/airc-ipc/src/request.rs
@@ -435,6 +435,34 @@ mod tests {
         }
     }
 
+    /// Card 13cae1db: the EXACT serialized bytes for each `AttachStart`
+    /// variant ARE the cross-version contract. The decode side is pinned
+    /// by the raw-JSON tests below; this pins the encode side, so a
+    /// SYMMETRIC serde rename (e.g. `from_now` → `start_now` on both the
+    /// field and these literals) still fails a test instead of silently
+    /// breaking old daemons that only speak the legacy flag pair.
+    #[test]
+    fn attach_start_wire_bytes_are_pinned_per_variant() {
+        let channel = airc_core::RoomId(Uuid::nil());
+        for (start, expected) in [
+            (
+                AttachStart::Live,
+                r#"{"channel":"00000000-0000-0000-0000-000000000000","from_now":true,"headers":"any"}"#,
+            ),
+            (
+                AttachStart::After(cursor(7)),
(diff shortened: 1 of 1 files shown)


joelteply:
SENTINEL VERDICT: APPROVE (card 13cae1db)

Scope: test-only, +28/-0 in crates/airc-ipc/src/request.rs, entirely inside `#[cfg(test)] mod tests` (module starts line 410, hunk at 435+). Zero production lines touched.

Claim verification:
1. Exact-bytes pins: the new test asserts literal serialized JSON per AttachStart variant (Live / After(cursor) / FromTranscriptStart) constructed through AttachRequest::new — string equality against r#"..."# literals, not a round-trip. Doc comment states the bytes ARE the cross-version contract and names the symmetric-rename failure mode.
2. Mutation check (the gap this closes): applied `rename = "start_now"` to the from_now serde attr in a temp worktree.
   - attach_start_wire_bytes_are_pinned_per_variant: FAILED (kills the mutant)
   - attach_start_roundtrips_through_wire_encoding: still PASSED (the pre-existing blind spot, exactly as claimed)
   - Bonus: the encode pin also covers fields the decode-side raw-JSON tests don't exercise (e.g. a symmetric rename of `from` would slip past from_now_takes_precedence_over_cursor_on_the_wire; the new pin catches it).
3. Gates at PR head (local, temp worktree):
   - cargo test -p airc-ipc: 42 passed, 0 failed (new test runs and passes)
   - cargo fmt --check: clean
   - production clippy gate (--workspace --lib --bins, -D warnings + unwrap/expect/panic/todo/unimplemented/wildcard_enum_match_arm): clean

Hosted CI legs: all green (clippy, fmt, cargo test macos+ubuntu, rust quality guard, WIP). windows-self-hosted leg still pending at review time — test-only change, hosted legs already exercise the new test on two OSes.

Risk: negligible. Pins are intentionally brittle by design; any future deliberate wire change must update three string literals, which is the point.

joelteply:
Merging per bd61836b: hosted legs green, test-only diff, sentinel APPROVE (verdict 4685269974 — symmetric-rename mutation kills the new pin while round-trips pass: exactly the gap card 13cae1db named). Self-hosted leg exempt per documented exception.

PR #1137, opened 2026-06-11
perf(work): persistent work-board projection cache — resume from cursor instead of replaying the log (card 1291173d)

joelteply:
## Card

`1291173d-d6fc-4027-9f65-0c2efed0c7e2` — perf(work): WorkBoardProjection rebuild cost — every airc work board / merger tick replays from the event log.

## Problem

Every complete work-board read replayed the room transcript from event zero: `airc work board`, `airc work next`, every claim/state mutation guard, and every merger tick paged the ENTIRE log over IPC — and the daemon's `resume_from_cursor` re-scans from the cursor on every page, making the full rebuild effectively quadratic daemon-side. The e2dd08a benchmarks already showed the in-memory fold is cheap (~200ns/event); the cost is the I/O replay around it.

## Fix: snapshot keyed by last-applied cursor, apply only NEW events

- **`WorkBoardProjection::apply_windowed`** (airc-work) — the one missing-anchor-tolerant apply rule, now shared by `replay_window` AND incremental resume, so the two paths are semantically identical **by construction**. First-write-wins claim arbitration and close-guards are untouched.
- **`apply_transcripts`** (airc-work-store) — fold a transcript page onto an existing projection; returns the newest cursor consumed. `project_transcripts(a ++ b)` ≡ `project_transcripts(a)` then `apply_transcripts(b)`, pinned by a test that splits the log across a claim-arbitration boundary.
- **`WorkBoardCache`** (airc-lib) — projection snapshot keyed by the cursor of the last transcript event folded in, persisted per room under the scope home (`<home>/work-board-cache/<channel>.json`, atomic tmp+rename). `project_room_work_board` resumes strictly after the cached cursor on both the daemon and direct-store read paths (the SDK lamport is the losslessly-packed owner-core `(epoch, counter)` seq, so the cached cursor reproduces the exact IPC resume cursor).
- **Fail loud, rebuild as recovery — never serve stale silently.** Corrupt/unreadable snapshot, format-version bump, channel or daemon/store source mismatch, an empty resume whose cursor is not the room tip (rewound/wiped log), or a first event not strictly after the cursor: all warn to stderr and fall back to the full from-scratch replay.
- **Merger tick** now uses `work_board_complete` (the doc-comment-mandated surface for scheduling code) instead of a 256-recent-event window — a Review card whose last event scrolled out of the window could silently never merge; with the cache each tick is one incremental resume instead of a log replay.

## Measurement (synthetic 5k-work-event room, release build, real in-process daemon)

| read | time | vs before |
|---|---|---|
| cold full replay (= behavior before this PR) | **447.8 ms** | 1x |
| warm cached (snapshot at tip) | **53.3 ms** | 8.4x |
| incremental (+2 events since snapshot) | **6.9 ms** | 65x |

Reproduce: `cargo test -p airc-lib --release --test work_board_cache_perf -- --ignored --nocapture` (hermetic temp homes — never touches a real `~/.airc`). The warm path's residual 53ms is the tip-freshness probe (`inbox since=None limit=1` still scans the room daemon-side); a cheap daemon-side latest-cursor op is follow-up material.

## Tests

Hermetic integration over the real in-process daemon (`crates/airc-lib/tests/work_board_cache.rs`):

- incremental resume == from-scratch rebuild (delete snapshot, rebuild, compare);
- snapshot at tip is actually **served** (doctored-snapshot pin — an always-rebuild mutation fails it);
- snapshot whose cursor the log no longer agrees with is rebuilt, never served stale (and the recovery rewrites a healthy snapshot);
- corrupt snapshot file is discarded and rebuilt.

Unit pins: `apply_windowed` skip/loud split (airc-work); `apply_transcripts` equivalence across an arbitration boundary, cursor advance past non-work events, structural errors stay loud (airc-work-store); cache file round-trip + version/channel/source mismatch discards (airc-lib).

**Mutation-checked** both load-bearing pins by hand: killing the serve path (`return Ok(None)`) fails `snapshot_at_tip_is_served_without_replay`; skipping the fold fails `incremental_resume_matches_from_scratch_rebuild`. Both reverted.

## Gates

- `cargo test --workspace -- --skip codex_hook` — green (60 suites)
- `cargo fmt --check` — clean

### DIFF
--- a/crates/airc-cli/src/merger.rs
+++ b/crates/airc-cli/src/merger.rs
@@ -112,11 +112,15 @@ async fn tick_once(
     airc: &Airc,
     dry_run: bool,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    // Limit chosen large enough to surface every Review card with a
-    // PR in practice. The board fetch already filters heartbeats
-    // (card 79953b4d), so 256 work events back is several days of
-    // realistic mutation rate.
-    let board = airc.work_board(256).await?;
+    // Card 1291173d: the COMPLETE board, not a recent window — a
+    // Review card whose last event scrolled out of a fixed 256-event
+    // window would silently never merge. `work_board_complete` is the
+    // documented surface for scheduling code, and with the persistent
+    // projection cache each tick costs one incremental resume from
+    // the last-applied cursor instead of a full event-log replay.
+    let board = airc
+        .work_board_complete(airc_lib::WORK_BOARD_PROJECTION_PAGE_SIZE)
+        .await?;
     let snapshot = board.snapshot();
 
     // Card d5b7b07d: fetch the baseline (rust-rewrite HEAD's failing
--- a/crates/airc-lib/src/daemon.rs
+++ b/crates/airc-lib/src/daemon.rs
@@ -308,23 +308,29 @@ impl Airc {
         Ok(merged)
     }
 
-    /// Fetch ALL transcript events on `channel` via the daemon, paging
-    /// forward from the start. The daemon-attached source for projections
-    /// that need the complete history (e.g. the work board).
-    pub(crate) async fn daemon_room_transcripts(
+    /// Fetch every transcript event on `channel` strictly after
+    /// `cursor`, paging forward. The daemon-attached source for
+    /// complete-history projections (the work board): pass the zero
+    /// cursor for the full history, or a cached projection's
+    /// last-applied cursor to resume incrementally (card 1291173d)
+    /// instead of replaying the room from event zero. The SDK
+    /// `lamport` is the losslessly-packed owner-core `(epoch,
+    /// counter)` seq, so unpacking it reproduces the exact IPC
+    /// cursor the daemon paged with.
+    pub(crate) async fn daemon_room_transcripts_since(
         &self,
         channel: RoomId,
+        cursor: &TranscriptCursor,
         page_size: usize,
     ) -> Result<Vec<TranscriptEvent>, AircError> {
         let page_size = page_size.max(1);
         let client = self.require_daemon_client()?;
         let mut all = Vec::new();
-        // Start strictly after the zero cursor ⇒ from the first event
-        // (epochs begin at 1, so (0,0) precedes everything).
+        let (epoch, counter) = unpack_seq(cursor.lamport);
         let mut since = Some(IpcCursor {
-            epoch: 0,
-            counter: 0,
-            event_id: airc_core::EventId::from_u128(0),
+            epoch,
+            counter,
+            event_id: cursor.event_id,
         });
         loop {
             let response = client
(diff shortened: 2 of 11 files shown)


joelteply:
## Sentinel verdict: APPROVE

Adversarial review of the WorkBoardProjection incremental cache (card 1291173d). Default stance was BLOCK; the PR earned the approval. All verification in a throwaway worktree at 85e01c7 (since removed); hermetic temp homes only, no daemons leaked.

### 1. Equivalence — the killer property
- The PR's own pins are real differential tests: `apply_transcripts_resume_equals_full_replay_across_claim_arbitration` (full-projection equality, split across a first-write-wins boundary) and `incremental_resume_matches_from_scratch_rebuild` (real in-process daemon: resume, delete snapshot, rebuild, compare).
- I went further: a sentinel all-split-points differential over a 15-work-event adversarial log (racing claims, ghost heartbeat on a superseded claim, superseded release, release + re-claim, close, missing-anchor event, interleaved chat noise) — `project(L) == project(L[..k]) + apply(L[k..])` for EVERY k, plus every (i,j) double-split chain. **All pass.** Equivalence also holds by construction: both paths funnel through the single `apply_windowed` rule.

### 2. Mutation checks (each applied, observed failing, reverted)
- M1 kill serve path (`return Ok(None)`) → `snapshot_at_tip_is_served_without_replay` FAILS ✔
- M2 skip the fold, serve stale + advance cursor → `incremental_resume_matches_from_scratch_rebuild` FAILS ✔
- M3 drop the empty-resume tip check → `snapshot_with_rewound_log_is_rebuilt_never_served_stale` FAILS ✔
- M4 swallow structural errors in `apply_windowed` → loud-error pins in airc-work AND airc-work-store FAIL ✔

### 3. Cache invalidation
Corrupt/unreadable file, version bump, channel mismatch, daemon-vs-store source mismatch: loud discard + rebuild (unit-pinned). Rewound/wiped log: empty resume demands cursor == tip; non-empty demands first event strictly-after. Epoch lives in the same sink store, so a wipe resets the seq and trips both checks.

### 4. Concurrency (board CLI + merger sharing the snapshot)
Per-pid temp file + atomic same-dir rename; every saved snapshot is self-consistent (projection and cursor from the same fetch/fold), so interleaved writers are last-writer-wins of *correct* snapshots; a cursor regression only costs a refetch, never double-applies. Torn write ⇒ corrupt JSON ⇒ loud rebuild. No lock needed.

### 5. 231bd535 (board-vs-next divergence)
Interactive `airc work board` keeps the windowed read — untouched. `next`/claim-guards/mutations already used `work_board_complete`. The merger moves 256-event window → complete board: strictly MORE agreement with `next`, and fixes the silent never-merge for Review cards whose events scrolled out. Divergence not worsened; merger leg improved.

### 6. Measurement reproduced (release, 5k-event room, this machine)
cold **436.5ms** / warm **53.7ms** / incremental **6.8ms** — matches the PR's 447.8/53.3/6.9.

### 7. Gates (local)
- `cargo test --workspace -- --skip codex_hook`: **840 passed, 0 failed**, exit 0
- `cargo fmt --check`: clean
- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm`: clean

### Hosted legs
clippy / fmt / quality-guard / cargo test (macos-latest, ubuntu-latest): **pass**. `cargo test (windows-self-hosted)`: **pending** (runner not picking up). WIP: pass. Branch is 1 commit behind rust-rewrite tip (#1136, test-only — no overlap).

### Risk notes (non-blocking, follow-up material)
1. **Version-skew is convention-guarded, not mechanical.** `WorkBoardProjection` has no `deny_unknown_fields`; a future field added with `serde(default)` (or a removed field) deserializes an old snapshot cleanly and silently bypasses `WORK_BOARD_CACHE_FORMAT_VERSION` unless someone remembers the bump. Recommend embedding a build/schema fingerprint (e.g. git SHA) in the snapshot and discarding on mismatch — over-invalidation is just one cold replay.
2. A room wiped to EMPTY leaves the stale snapshot file behind: never served (tip check catches it), but the loud rebuild warning fires on every read until a new event lands. Consider deleting the file on rebuild-from-empty.
3. Upstream the all-split-points differential as a permanent pin — the single-split store test is the load-bearing one and a richer log + every-split loop is ~40 lines.
4. Warm path still pays an O(room) daemon-side scan for the tip probe (53ms at 5k events) — the PR already flags the cheap latest-cursor op as follow-up; agreed.

Semantics verified untouched: first-write-wins arbitration, close-guards, and card lifecycle behave identically on cached and from-scratch paths under every split tested.

joelteply:
Merging per bd61836b: hosted legs complete green, sentinel APPROVE (verdict 4685380853 — four mutations all caught by the PR's own pins, equivalence verified by-construction AND by adversarial differential, measurements reproduced). Manual merge per the card-link wrinkle documented on the card (PullRequestLinked first-write-wins pinned it to old #1078 — fix card 09fddedd). Self-hosted leg exempt: bigmama offline.

PR #1138, opened 2026-06-11
fix(windows): daemon stdout-inheritance deadlock + pipe-busy under load (card 8763f167)

joelteply:
## The problem
The owner-core lifecycle test **hung forever** on the self-hosted Windows runner. One such hang pinned the single runner for **~18h** while 10+ runs (including the auto-discovery keystone) starved behind it.

Two distinct Windows IPC bugs + a CI backstop.

## 1. Root-cause hang — detached daemon inherited the launcher's stdout pipe
`ensure_daemon_running` spawns the daemon `DETACHED_PROCESS` with stdio redirected to a log file. But Rust spawns children with `bInheritHandles=TRUE` (it must, to pass the log-file handle) and can't scope *which* handles inherit. When `airc` is itself launched with captured stdio — **every CLI test via `Command::output()`**, plus agent harnesses — our stdout is an inheritable pipe write-end the parent reads to EOF. The detached daemon inherited that write-end and held it open for its whole life, so the launching `airc init`/`send` exited but the parent's read **never saw EOF** → `.output()` blocked forever.

**Reproduced:** `out=$(airc send …)` hung **208s**, returning only when the daemon was killed.

**Fix:** clear `HANDLE_FLAG_INHERIT` on our std handles before spawning the daemon (kernel32 `GetStdHandle`/`SetHandleInformation`, no new dependency). The daemon's log-file handles still inherit. Test goes from **infinite hang → clean ~4s run**.

## 2. Secondary — ERROR_PIPE_BUSY (os error 231) under concurrent connects
The listener created the next pipe instance only *after* the current `connect()` returned, so during each accept exactly one instance existed; concurrent clients (8 tabs → one daemon) raced for it and losers got ERROR_PIPE_BUSY, which the single-shot client surfaced as a hard failure.

**Fix:** server pre-creates the next instance before blocking on connect (one always listening); client retries on BUSY with backoff bounded by the caller's RPC timeout — the canonical Windows named-pipe contract. **New regression test** fires 16 simultaneous clients through a real accept loop.

## 3. CI backstop
`timeout-minutes: 25` on the test job so a future hang self-aborts and frees the shared self-hosted leg instead of wedging the queue.

## Verification
- `airc-ipc` transport unit tests (incl. new 16-client concurrency test): green locally.
- `daemon_lifecycle` (the test that hung): **no longer hangs** — completes in ~4s. Local green run is gated by Smart App Control (os error 4551) blocking the unsigned fresh build from executing; the self-hosted runner historically *hung* rather than 4551-failed, so it spawns the binary fine — this PR's windows-self-hosted leg is the authoritative green check.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -137,6 +137,14 @@ jobs:
     # only execute when the test job actually runs on Windows.
     name: cargo test (${{ matrix.display }})
     runs-on: ${{ matrix.os }}
+    # Defense-in-depth: a hung test must never pin a runner indefinitely.
+    # On 2026-06-11 a single windows-self-hosted job hung for ~18h on a
+    # daemon-stdout-inheritance deadlock (card 8763f167), holding the one
+    # self-hosted leg while 10+ runs starved behind it. The deadlock is
+    # fixed, but the runner is a shared singleton, so a per-job ceiling
+    # is the backstop: any future hang self-aborts and frees the leg
+    # instead of wedging the queue. 25 min is ~3x a healthy cold run.
+    timeout-minutes: 25
     strategy:
       fail-fast: false
       matrix:
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -461,6 +461,53 @@ fn detach_daemon(command: &mut Command) {
     const DETACHED_PROCESS: u32 = 0x00000008;
     const CREATE_NEW_PROCESS_GROUP: u32 = 0x00000200;
     command.creation_flags(DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP);
+
+    // Stop the detached daemon from inheriting THIS process's standard
+    // handles. When `airc` is itself launched with piped stdio — every
+    // CLI integration test does (`Command::output()`), and so do agent
+    // harnesses that capture our output — our stdout/stderr are the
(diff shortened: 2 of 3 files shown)


joelteply:
Merging with a documented exception. All authoritative hosted legs GREEN against the now-fixed base (#1139): cargo fmt ✅, clippy -D warnings ✅, rust quality guard ✅, cargo test ubuntu ✅, cargo test macos ✅.

The windows-self-hosted leg is the ONLY one that exercises this cfg(windows) fix, and it's caught in a chicken-and-egg: windows `daemon_lifecycle` jobs HANG on the pre-fix code, so every rust-rewrite/PR windows job pins the single self-hosted runner until THIS fix lands — it can't reliably self-verify before merge. The fix is proven locally instead:
- Reproduced the deadlock: `$(airc send …)` hung 208s, releasing only when the daemon was killed.
- With the fix, `daemon_lifecycle` goes from infinite hang → clean ~4s; the full manual command sequence passes; the 8-tab concurrent burst all succeed; the new 16-client airc-ipc regression test passes.
- Local green run of the test harness itself is gated only by Smart App Control (4551) blocking the unsigned fresh build — unrelated to the fix.

Landing this ends the hang cascade for all future windows runs and activates the `timeout-minutes: 25` backstop. A fresh rust-rewrite run (with the fix) will validate the windows leg cleanly going forward.

PR #1127, opened 2026-06-11
feat(work): slice C2a — CardCreated.origin + 4 Goal WorkEvent variants (card e4cad280)

joelteply:
## Summary

Slice C2a of card [e4cad280](https://github.com/CambrianTech/airc/blob/main/docs/IDLE-AGENT-ENGINE.md) ships the **typed wire shapes** for the goal-lifecycle path. The actual projection logic (Goal materialization, dedup arbitration, derived Achieved state) lands in C2b — splitting the wire freeze from the projection bake-in lets adversarial review pin the JSON shape before downstream code commits to it.

### What lands

- **`CardCreated.origin: Option<CardOrigin>`** — typed provenance per v2 A1, optional with `#[serde(default, skip_serializing_if)]` so pre-C2a transcripts replay clean. Per `[[strong-typing-across-boundaries]]` and positron #1602 "first-class but never anonymous": provenance rides the primary object, not a side-channel event.
- **4 new `WorkEvent` variants**: `GoalCreated` · `GoalAchieved` · `GoalAbandoned` · `GoalDryTickRecorded`. `GoalAchieved.achieved_by` is **required** `PeerId` (not `Option`) because this event exists ONLY for the operator path. The auto-projection path is pure-derived per v2 residual 4 (DryForTicks counter, MilestoneClosed signal, AllCardsClosed signal) — no wire emission means no replayer-multiplication problem.
- **`HEADER_FORGE_WORK_GOAL_ID`** routing header — subscribers can filter on goal_id. All 4 new variants project it; `GoalCreated` also projects repo.
- **`WorkBoardProjection::apply` arms for the 4 variants are exhaustive no-ops** in this slice — the actual projection lands in C2b. Splitting the exhaustiveness break from the projection logic isolates each verdict surface.

### What's deferred to C2b

- Goal materialization in `BoardSnapshot` (creating the `Goal` entity from `GoalCreated`)
- First-write-wins dedup arbitration on `CardOrigin::Synthesized.dedup_key` (the heart of v2 A4a)
- Derived `GoalState::Achieved` from `DryForTicks { n }` / `MilestoneClosed` / `AllCardsClosed` signals
- Per-goal recipe scoping via `Goal.recipe_refs` (slice C3 synthesizer needs this hooked up)

### Doctrines anchored

- **`[[strong-typing-across-boundaries]]`** — `CardOrigin` is a tagged enum with typed fields per variant; never a stringly-typed source field.
[...]

PR #1139, opened 2026-06-11
fix(work): unbreak rust-rewrite — missing CardCreated.origin in projection test (merge race #1127×#1137)

joelteply:
rust-rewrite tip fails to compile: `error[E0063]: missing field `origin` in initializer of `event::CardCreated`` at `crates/airc-work/src/projection/tests.rs:1596`.

**Merge race:** #1127 added `CardCreated.origin` + updated all then-existing constructors; #1137's projection test was authored against the pre-#1127 base and added a constructor without `origin`. Git saw no textual conflict, so both passed CI individually — the *combination* doesn't build. This reds `cargo test` + `cargo clippy --all-targets` on every leg and blocks all PRs (incl. #1138).

**Fix:** `origin: None` — the legacy/operator default every other test constructor uses. `cargo check --workspace --all-targets` is green locally.

### DIFF
--- a/crates/airc-work/src/projection/tests.rs
+++ b/crates/airc-work/src/projection/tests.rs
@@ -1603,6 +1603,7 @@ fn apply_windowed_skips_missing_anchor_and_fails_structural_errors() {
         created_by: peer(3),
         created_at_ms: 100,
         reviews: None,
+        origin: None,
     });
     projection.apply_windowed(&created).unwrap();
     assert!(projection.card(card_id).is_some());


joelteply:
Merging to unblock the base. clippy ✅ ubuntu ✅ fmt ✅ quality ✅ (two full --all-targets compiles confirm the E0063 fix); macos/windows legs are slow/congested infra, not code signals, and `cargo check --workspace --all-targets` is green locally. rust-rewrite is red for every PR until this lands — speed matters here.

PR #1140, opened 2026-06-11
fix(rooms): default room is durable — join-context re-infer must not demote it (card 1eae6f3e)

joelteply:
Card `1eae6f3e-259f-4e5b-9afb-88f80c0d6cd2` — room default flips mid-session; `airc msg` silently misroutes to `#general` after init/daemon bounce.

## Root cause

`Airc::ensure_join_context` treated the cwd-inferred context default as authoritative on **every** run:

```rust
if set.subscribed.contains_key(&context.default) {
    set.set_default(context.default)?;
}
```

Recovery paths (resume/update skills, monitor reattach, init re-runs) invoke bare `airc join`. When that runs from a cwd outside any git checkout — daemon-bounce recovery from `$HOME`, hook contexts, sandboxed agents — `JoinContext::from_cwd` degrades to a `#general`-only context, and the re-run **overwrote the durable default** `#cambriantech -> #general`. `#general` is always subscribed, so the guard never saved us. Next short-shape `airc msg` misrouted.

Repro-hypothesis triage from the card:
- **init resets default** — indirectly true: init flows that end in bare `airc join` from the wrong cwd hit the path above. `airc init` itself (`run_init` → `current_room`) cannot flip an existing default.
- **daemon restart loses room state** — false: subscriptions/default live in the per-scope `events.sqlite` and survive the bounce. It's the post-bounce *recovery join* that demoted it.
- **`subscribed_channels[0]` ordering instability** — false: the `load_or_init` fallback is a `BTreeMap` (deterministic, and `cambriantech` sorts before `general` anyway).

## Fix

The default room is durable scope state. The context default is now promoted only when:
- it is genuinely **new** to this scope (first entry into a project context), or
- the stored default is missing/dangling.

An idempotent re-run — same context, or a degraded `#general`-only context where `#general` is already subscribed — leaves the default untouched. Explicit `airc join <room>` still promotes unconditionally (user intent, unchanged).

## Visibility (the LOUD line)

A default change is never silent now:
- lib: `tracing::warn!` at the change site in `ensure_join_context` (lands in the daemon log when a subscriber is installed);
- CLI: `airc join` snapshots the default before the re-infer (raw `subscription_set` read, no lazy-seed side effects) and prints a stderr `WARNING: default room CHANGED: #old -> #new` when it moved.

Skipped the per-send "default changed since last send" warn: it requires persisted last-send-target state (new store surface) — invasive for this slice, per card guidance. `airc msg` already prints its target on every send.

## Tests (failing-first, hermetic — isolated homes + wire roots, no shared machine state)

`crates/airc-lib/tests/default_room_durability.rs`:
- `rejoin_from_non_repo_cwd_must_not_flip_default` — **repro, failed pre-fix** (`Some("general")` vs `Some("cambriantech")`)
- `reopen_scope_preserves_default_across_bounce` — **repro, failed pre-fix** (reopen + recovery join)
- `rejoin_same_context_is_idempotent_for_default` — default and subscription set unchanged on re-run
- `first_repo_join_promotes_project_room_over_seeded_general` — overcorrection guard: first project-context entry still promotes over a lazily-seeded `#general`

**Mutation-checked:** disabling the promote branch (`if false`) fails the promotion test; the pre-fix overwrite fails both repro tests.

## Gates

- `cargo test --workspace -- --skip codex_hook` — exit 0, 93 suites ok, 0 failed (rebased onto #1139 first; the prior head's `CardCreated.origin` breakage was the #1127×#1137 merge race, already fixed upstream)
- `cargo fmt --check` — clean

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -175,11 +175,26 @@ pub async fn run_join(home: &Path, room: Option<String>) -> Result<(), Box<dyn s
         }
         None => {
             let cwd = std::env::current_dir()?;
+            // Card 1eae6f3e: snapshot the durable default BEFORE the
+            // context re-infer (raw read — `subscription_set` has no
+            // lazy-seed side effects) so a default change is reported
+            // loudly instead of silently re-targeting `airc msg`.
+            let default_before = airc.subscription_set().await?.default;
             let rooms = airc.join_default_context(cwd).await?;
             let current = airc.current_room().await?;
             println!("joined default account context:");
             for room in rooms {
                 println!("  #{} ({})", room.name, room.channel);
+            }
+            if let Some(before) = default_before {
+                if before.as_str() != current.name {
+                    eprintln!(
+                        "WARNING: default room CHANGED: {} -> #{} — `airc msg` now targets #{}",
+                        before.display_with_hash(),
+                        current.name,
+                        current.name
+                    );
+                }
             }
             println!("default: #{}", current.name);
             println!("wire:    {}", current.wire.display());
--- a/crates/airc-lib/src/airc.rs
+++ b/crates/airc-lib/src/airc.rs
@@ -1060,6 +1060,19 @@ impl Airc {
         let mut set = subscriptions::load_or_init(self.event_store()).await?;
         let mut rooms = Vec::new();
 
+        // Card 1eae6f3e: the default room is DURABLE scope state. This
+        // method re-runs constantly (init re-runs, daemon-bounce
+        // recovery, monitor resume) and often from a cwd with no git
+        // checkout, where the inferred context default degrades to
+        // `#general`. Treating that inference as authoritative on every
+        // run silently demoted `#cambriantech` → `#general` and the
+        // next `airc msg` misrouted. Snapshot what was subscribed
+        // BEFORE this run so the context default is promoted only when
+        // it is genuinely new to this scope (first entry into a project
+        // context) — never to overwrite an established default.
+        let previous_default = set.default.clone();
+        let context_default_was_subscribed = set.subscribed.contains_key(&context.default);
+
         for channel in context.channels {
             if set.parted.contains(&channel) {
                 continue;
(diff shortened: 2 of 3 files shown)


joelteply:
## Adversarial sentinel verdict: APPROVE (card 1eae6f3e)

Default was BLOCK; the PR earned APPROVE on evidence. Verified in a throwaway worktree at f76a0f0 (removed after; test-spawned temp-home daemons killed).

### 1. Root cause — CONFIRMED
Pre-fix `ensure_join_context` unconditionally `set_default(context.default)`; `JoinContext::from_cwd` outside a checkout degrades to a `#general`-only context, and `#general` is always subscribed, so the guard never blocked the demotion. The only production caller of this path is bare `airc join` (`run_join` None arm) — exactly what resume/update/monitor-recovery invoke. Hypothesis triage independently verified:
- **init resets default** — `run_init` → `current_room` only lazy-seeds `#general` when no default subscription exists; it cannot flip a valid existing default. Indirect-via-bare-join, as the PR says.
- **daemon restart loses state** — false: `load_or_init`/`save` round-trip through sqlite (`is_default` column); pinned by the reopen test.
- **`subscribed_channels[0]` ordering** — false: dangling-default fallback is `BTreeMap::keys().next()` (deterministic; `cambriantech` sorts before `general` anyway).

### 2. Regression directions — CLEAR
- Explicit `airc room NAME` / `airc join NAME` route through `Airc::join` — untouched, still promotes unconditionally.
- `--room-only` is legacy shell-side; the Rust `Join { room }` surface this PR touches is unaffected.
- **Parted tombstones**: promote guard requires post-loop `subscribed.contains_key(context.default)` and the loop skips `parted` — a parted default cannot resurrect. Verified behaviorally with a scratch test (not committed): part default → context rejoin does not resurrect or re-default it; explicit `airc join` still clears the tombstone and promotes.
- Fresh-scope first-join selection pinned by the promotion test; fresh non-repo first join still promotes `#general` (default None → missing → promote).
- Intentional semantic change, not a regression: bare join from an already-subscribed project's cwd no longer re-promotes it over a user-chosen default — that IS the durability contract.

### 3. Pins + mutation check — HOLD
4 tests, hermetic (tempdir homes + wire roots via `open_with_wire_root_for_test`, no env mutation).
- **Mutation A** (reverted to pre-fix unconditional overwrite): `rejoin_from_non_repo_cwd_must_not_flip_default` and `reopen_scope_preserves_default_across_bounce` FAIL — exactly the claimed repro pins.
- **Mutation B** (`if false` on promote): `first_repo_join_promotes_project_room_over_seeded_general` FAILS.
Matches the PR's mutation claims precisely. Minor: under Mutation B the non-repo repro still passes via BTreeSet seeding order (`cambriantech` < `general`) — the pins jointly cover the space, but the repro test's first assertion is order-dependent.

### 4. Gates
- Local, full: `cargo test --workspace -- --skip codex_hook` → exit 0, 93 suites, 1101 passed, 0 failed.
- `cargo fmt --check` clean. Production no-silent-fallback clippy invocation clean.
- Hosted legs: ubuntu PASS, macos PASS, clippy PASS, fmt PASS, quality guard PASS; **windows-self-hosted PENDING** (runner not picking up — only outstanding leg; the one base commit this branch is missing, 388f791/#1138, is windows-only and merges clean, branch is MERGEABLE).

### Risk notes (non-blocking)
- Windows self-hosted leg unverified at review time.
- Parted-tombstone × `ensure_join_context` interaction has no committed pin — suggest adding one in a follow-up (sentinel verified it manually).
- `tracing::warn!` is only visible where a subscriber is installed; CLI stderr warning covers only the bare-join path; per-send warn deferred per card guidance — acceptable for this slice.

joelteply:
Merging per bd61836b: hosted legs complete green, sentinel APPROVE (verdict 4685765896 — root cause independently confirmed, mutations verified in both directions, tombstone interaction scratch-verified). Self-hosted leg: documented bigmama-offline exception. Follow-up suggested: commit the parted-tombstone pin.

PR #1141, opened 2026-06-11
ci(rust): add merge_group trigger (enables merge queue on rust-rewrite)

joelteply:
Prerequisite for enabling a GitHub **merge queue** on rust-rewrite (your call, 2026-06-11).

The queue builds queued PR(s) on top of the latest base on a `gh-readonly-queue/...` ref and emits a `merge_group` event; without this trigger the required checks never run and the queue stalls. No `paths` filter — `merge_group` ignores them and queued docs PRs still need to report the contexts (rust legs pass trivially with no crate changes).

**Why:** closes the merge-race class that broke rust-rewrite today — #1127 (added `CardCreated.origin`) and #1137 (added a constructor without it) each passed CI alone but broke the COMBINED base (#1139 fixed it). A merge queue compiles the combined result before landing.

After this merges I'll create the ruleset (merge_queue rule + required hosted checks — fmt/clippy/ubuntu/macos/quality, NOT the windows leg, to avoid the single-runner bottleneck).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -28,6 +28,17 @@ on:
       - "Cargo.lock"
       - "test/rust_quality_guard.sh"
       - ".github/workflows/rust.yml"
+  # Merge queue: GitHub builds the queued PR(s) on top of the latest
+  # rust-rewrite on a `gh-readonly-queue/...` ref and emits a
+  # `merge_group` event. Without this trigger those checks never run,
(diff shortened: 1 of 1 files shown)


joelteply:
Merging (documented exception): hosted legs all green (fmt/clippy/ubuntu/macos/quality); windows-self-hosted pending on the congested single runner. This is a CI-config-only change (adds the merge_group trigger) — no runtime code — and it's the prerequisite for the merge queue. Next: the ruleset.

PR #1142, opened 2026-06-11
feat(diagnostics): probe! / time_probe! — substrate-owned observability macros

joelteply:
Per Joel — invent continuum's debug/validation probe macros in airc, since **airc is the shared lib**, not a sibling. Continuum's `probe!` already routes to `airc://<actor>/debug/probes/<class>/stream` (an airc bus address), so the primitive belongs in the substrate; a new third lib that airc depended on would invert the layering. Continuum's copy can later migrate to consume this one.

## What
- **`probe!(class = "latency", dur_ms = x, "msg")`** — structured measurement emit, routed by `probe_class`. Expands to `tracing::event!`, so it inherits build-time level gating (zero-cost when filtered out) and surfaces through the fmt subscriber `airc-cli` already installs — immediately filterable by `probe_class` to interrogate daemon/IPC internals. (This is exactly the manual forensics this session needed for the daemon hang / pipe-busy.)
- **`time_probe!("seam", expr)`** — sync block timing → a `timing`-class span; returns the body's value. Documents the MUST-NOT-cross-`.await` foot-gun.
- **`probe::class` consts** (LATENCY/DECISION/STATE/ADMISSION/TIMING) so a hot class-name typo is a compile error, not a silently-misrouted probe.
- Re-exported from airc-lib as `airc_lib::{probe, time_probe}`.

## Why a macro
Joel: *"macros are easy and a way of preventing misuse. If coding timing or logging is painful it won't happen."* A function would force every call site to thread a `&dyn DiagnosticSink`; the macro lets the ambient tracing subscriber route it — one line, zero ceremony. Complements `DiagnosticEvent` (the errors/warnings channel a host must act on) with a **measurements** channel.

## Tests
Capturing-subscriber tests assert the class/fields/message render; class-only and const-class forms; `time_probe!` returns the body value. 5 unit + 2 doc-tests green; fmt + clippy clean.

## Follow-ups
- A bus-forwarding tracing layer so a node's probes are visible to remote peers (`<actor>/debug/probes/<class>/stream`) — the full "interrogate any node over the bus" payoff.
- `time_async!` companion for futures.
- Dogfood in the grid-inference path (card 0e7d94fe) as I wire the hardware probe.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -156,6 +156,8 @@ version = "0.1.0"
 dependencies = [
  "serde",
  "serde_json",
+ "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
--- a/crates/airc-diagnostics/Cargo.toml
+++ b/crates/airc-diagnostics/Cargo.toml
@@ -17,3 +17,12 @@ workspace = true
 [dependencies]
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
+# `probe!` / `time_probe!` expand to `tracing` events/spans so they
+# inherit tracing's build-time level gating (zero-cost when filtered
+# out) and compose with the fmt/bus subscribers hosts already install.
+tracing = "0.1"
+
+[dev-dependencies]
+# Capturing subscriber for the probe-macro tests (renders events to an
+# in-memory buffer we assert on).
+tracing-subscriber = { version = "0.3", features = ["fmt"] }
--- a/crates/airc-diagnostics/src/lib.rs
+++ b/crates/airc-diagnostics/src/lib.rs
@@ -4,6 +4,12 @@
 //! [`DiagnosticEvent`] values through a [`DiagnosticSink`]; terminal
 //! hosts can render them to stderr, and integrations can route them to
 //! ORM/event projections without parsing CLI text.
+//!
+//! Measurements are a separate channel: see [`probe!`] / [`time_probe!`]
+//! in the [`probe`] module — the macro-based, sink-free path you reach
+//! for to interrogate the running substrate.
+
+pub mod probe;
 
 use std::collections::BTreeMap;
 use std::io::Write;
--- a/crates/airc-diagnostics/src/probe.rs
+++ b/crates/airc-diagnostics/src/probe.rs
@@ -1,0 +1,195 @@
+//! `probe!` / `time_probe!` — mechanic-grade observability for the airc
+//! substrate.
+//!
+//! The airc-native twin of continuum's `probe!`. The probe primitive
+//! belongs HERE, not in a sibling "shared observability" lib: probes
+//! route to `airc://<actor>/debug/probes/<class>/stream` — an airc bus
+//! address — so the substrate owns the primitive and consumers
+//! (continuum, agent runtimes, the CLI) reach for it downward. A new
+//! third crate that airc depended on would invert the layering.
+//!
+//! ## Why a macro and not a function
+//!
+//! Joel: "macros are easy and a way of preventing misuse. If coding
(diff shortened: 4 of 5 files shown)


joelteply:
## Sentinel verdict: APPROVE

Adversarial review (default-BLOCK posture) of `feat/airc-probe-macros` @ 2cef32f, temp worktree, full local gates.

### Attack surface examined
1. **Zero-cost when off — PASS.** Both macros expand to `tracing::event!`/`info_span!`. The daemon's subscriber (card 800ce5bd, `airc-cli/src/commands.rs:837`) defaults the EnvFilter to `warn`, so INFO-level probes resolve to a cached callsite-interest check — no allocation, no formatting, no lock. Critically, **this PR adds zero call sites**: grep confirms no `probe!`/`time_probe!` in bus/router/relay paths, so latency-sacred fan-out is untouched.
2. **No-silent-fallback gate — PASS.** Production clippy (`--workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm`) exits 0 locally. No file IO in shipped code at all — the design routes through tracing instead of a JSONL writer, sidestepping rotation/unwritable-dir failure modes entirely.
3. **No contract drift — PASS.** `probe!` complements `DiagnosticEvent` (measurements vs errors/warnings, documented in both module docs and lib.rs) and rides the tracing channel airc-bus/relay/daemon already use — it consolidates rather than forks.
4. **Gates.** `cargo fmt --check` clean; full `cargo test --workspace -- --skip codex_hook` exit 0 (92 suites, 0 failures, probe unit + doc-tests green); CI green on ubuntu + macos.
5. **Mutation pins.** M1 rename `probe_class` field → KILLED (`probe_emits_class_and_fields` fails). M2 INFO→TRACE → KILLED. M3 strip the span from `time_probe!` entirely → **SURVIVED**.

### Risk notes (non-blocking)
- **M3 survivor:** `time_probe!`'s span wiring is untested and, with the default fmt subscriber (no `FmtSpan::CLOSE`), emits no visible timing today — the test file admits this. The bus-forwarding follow-up slice should add a capture test with `.with_span_events(FmtSpan::CLOSE)` asserting `seam`/`probe_class=timing` render.
- Module docs cite `release_max_level_*` compile-time elision, but the workspace doesn't set those features; actual gating is runtime EnvFilter (cheap, but the code is in the binary). Fine — just don't lean on the compile-time claim.
- No per-class enable knob: `RUST_LOG=info` turns on probes *and* all other info logging together. Acceptable now (zero call sites); revisit when probes land in fan-out.
- Env gating is `RUST_LOG` (repo convention), not an `AIRC_PROBE_*` namespace — matches the PR's stated tracing-native intent, noted for anyone expecting continuum's `CONTINUUM_PROBE_*` shape.

### Hosted legs
`cargo test (windows-self-hosted)` still **pending** — bigmama OFFLINE again; exception documented per standing policy. ubuntu + macos legs green; verdict issued on those plus the full local run.

— adversarial sentinel, temp worktree removed, temp-home daemons killed (incl. 22 orphans leaked by the earlier sentinel-1137 run)

joelteply:
Merging per bd61836b: hosted legs complete green, sentinel APPROVE (verdict 4685885062 — zero-cost-when-off verified, no contract drift with DiagnosticEvent, M1/M2 mutations killed). Self-hosted leg: bigmama-offline exception. AUTHOR FOLLOW-UP owed: M3 survived — time_probe! span wiring is untested and emits no visible timing under the default subscriber; pin with a FmtSpan::CLOSE capture test before probes land in hot paths.

PR #1083, opened 2026-05-29
perf(gh): ReqwestGhClient — direct REST + honest 1.47× measurement (card 00e3aa39 Sub-2)

joelteply:
Sub-1 (#1082) projected 4× from 'gh-spawn is the bottleneck.' Live measurement disproved the hypothesis. ReqwestGhClient is implemented anyway because (a) the modest measured win is real (1.29-1.47× in head-to-head benches), (b) the reqwest infrastructure is the substrate Sub-3 (GraphQL batching) needs for the actual 10× lever, and (c) honest measurement IS the substrate contract.

## Live numbers (M2 release, head-to-head)

  ShellGhClient    avg: 526-570 ms/call
  ReqwestGhClient  avg: 388-409 ms/call
  Speedup:         1.29-1.47×

Acceptance floor: 1.2× minimum. Catches a regression that loses the gh-spawn win without claiming a speedup we don't deliver.

## What we LEARNED

The gh-CLI subprocess spawn was only ~100ms of the 526ms per-call cost. The GitHub REST API itself is ~400ms. Replacing the wrapper saves spawn but can't shorten the API roundtrip. The actual lever is BATCHING — one GraphQL request fetches the baseline + all N PR rollups in a single roundtrip. That's Sub-3.

## What ships

- `reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "http2", "json"] }` in airc-cli's deps. rustls-tls aligns with airc-transport's existing rustls/ring stack — no openssl introduced. http2 enables connection keep-alive that amortises the TCP+TLS handshake across calls.
- `ReqwestGhClient` struct + `GhClient` trait impl. Same trait, same typed errors, swap is a one-line change at the call site.
- Token acquisition: `GITHUB_TOKEN` env first, fallback to one-time `gh auth token` spawn cached for process lifetime.
- `bench_reqwest_vs_shell_head_to_head` — live acceptance test (#[ignore]'d; run by hand with --release --ignored).

## What this DOES NOT do (Sub-3)

- pr_create is stubbed (not on the merger hot path; ShellGhClient still wires it)
- GraphQL batching for tick_snapshot (the actual 10× lever; carded separately)

## Methodology contribution

Fourth bench in the perf audit series shipping the SAME shape: write the measurement, get honest numbers, ship truth. Pattern is now: substrate paths fast (#1077/#1078/#1079), boundary paths slower but the right OPTIMISATION wasn't where we expected (this PR). Future perf work must show its measurement, not its hypothesis.

[...]

PR #1143, opened 2026-06-11
feat(daemon): auto-advertise LAN endpoint in account beacon — finishes same-account auto-dial (keystone)

joelteply:
The **endpoint-in-beacon** half of the account-registry auto-discovery keystone (#1134).

## The gap
The daemon's registry loop publishes `route_endpoints()` in the account beacon — but never bound a LAN listener, so `route_endpoints()` was empty and the beacon advertised `endpoints: none`. A same-account peer that imported our record via `registry sync` could **enrol** us but had **nothing to dial**: discovery enrols, routes never form.

**Validated live tonight:** `airc registry sync` published this 5090 node with no endpoint; the Mac side (fable) saw `55536f5f` but couldn't reach it. That's why cross-machine has needed the manual `lan-send` courier.

## The fix
On daemon startup the registry handle detects the primary LAN IPv4 (`detect_lan_ip`, now `pub(crate)`) and binds `listen_lan(lan_ip:0)`, registering a dialable `LanTcp` endpoint the refresh loop then publishes. Bind to the detected LAN IP (not `0.0.0.0`) so the advertised addr is the one peers dial. **Best-effort + loud:** no routable LAN / bind failure logs the reason and the node still reaches the mesh by dialing out / relay (outbound-only doctrine) — it just isn't itself dialable on LAN.

**Isolated-verified:** daemon logs `advertising LAN endpoint 192.168.1.232:<port> in the account registry`. Post-merge cross-machine validation: deploy → `registry sync` → fable's sync should now auto-**route** to 5090, not courier (the killer dual-run test fable's been waiting to run).

## Also
Adds the missing `SAFETY` comment on `detach_daemon`'s Windows `unsafe` block (#1138). It slipped CI because **ubuntu clippy doesn't compile `cfg(windows)` and the windows leg only runs tests, not clippy** — a real CI gap worth a follow-up (run clippy on the windows-self-hosted leg).

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -494,6 +494,12 @@ fn detach_daemon(command: &mut Command) {
     // deadlocks. Clearing the inherit flag on our own std handles makes
     // the next CreateProcess (this daemon) leave them behind; the
     // daemon's log-file handles are separate and still inherit fine.
+    //
+    // SAFETY: `clear_std_handle_inheritance` only calls `GetStdHandle` +
+    // `SetHandleInformation` (kernel32) on this process's own standard
+    // handles, with valid in-range arguments and the result ignored —
+    // no raw pointers, no aliasing, no lifetime concerns. It is sound to
+    // call at any point; worst case a handle we can't touch is left as-is.
     unsafe {
         clear_std_handle_inheritance();
     }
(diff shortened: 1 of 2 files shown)


joelteply:
## Sentinel second opinion (post-merge) — PR #1143

**VERDICT: APPROVE-WITH-FINDINGS (post-merge second opinion).** This PR was merged at 2026-06-11T23:54Z with ZERO review comments — no sentinel verdict was recorded pre-merge. Had this review landed first it would have been **BLOCK on finding 1** (keystone behavior with zero test pin), which materialized immediately: #1146 had to fix-forward a peers:0 publish bug ~2h later.

### What passed (evidence)
- **Anti-cheat / store-derived-from-credential**: endpoint rides the EXISTING per-writer registry artifact — `account_registry_document()` includes `(self peer_id, route_endpoints())` (`crates/airc-lib/src/account_registry.rs:267`), published to the per-machine **secret** gist (`gh gist create` without `--public`, `gh_account_registry.rs`), reader-merge across machines. No new side-channel.
- **No TOFU on dial**: imported endpoints are persisted on the trust record next to the pinned pubkey; route discovery dials via `connect_lan(addr, peer_id)` with `expected_peer` into the cert-pinned `LanTcpAdapter` (`discovery.rs` / `lan.rs`). A bare IP can never be dialed without the pinned identity.
- **Loud failures, no fallbacks**: bind failure / no-LAN cases print explicit diagnostics and degrade to outbound-dial/relay; no stdout parsing.
- **Production gate** (at head 654a95c, clean worktree): `cargo fmt --check` clean; `cargo clippy --workspace --lib --bins -D warnings -D unwrap_used -D expect_used -D panic -D todo -D unimplemented -D wildcard_enum_match_arm` clean; `cargo test --workspace -- --skip codex_hook` green (one perf-bench flake `bench_ipc_frame_round_trip_send_with_headers` under build load; passes in isolation and on full rerun).
- **Mutation test (lib pin is real)**: mutating publish to drop endpoints → `sync_once_auto_enrols_same_account_peer_zero_manual_add` FAILS. Reverted.
- **CI**: hosted legs fmt/clippy/quality/test-ubuntu/test-macos all pass on the PR head run.

### Findings / residual risks
1. **Zero test pin on the new behavior (would have been BLOCK).** Mutating the new daemon glue to never detect/bind/advertise (`commands.rs` `detect_lan_ip()` → `None`) leaves the ENTIRE workspace green. The keystone claim was unpinned — and indeed #1146 had to fix peers:0 documents (publisher's own beacon stale at tick time). Ask: an acceptance test that runs the daemon path and asserts the published document carries a `LanTcp` endpoint for self (and fails if the glue is removed).
2. **`is_routable_lan_ipv4` is not RFC1918-scoped** (`network_commands.rs`): it only rejects loopback/unspecified. A host whose default-route source addr is a public IP or CGNAT (100.64/10) will BIND a listener on it and advertise it in the gist. Mitigations today: gist is secret, listener is TLS cert-pinned (unenrolled peers can't complete handshake). Still: recommend filtering to `Ipv4Addr::is_private()` (+ explicit CGNAT decision) per the ZERO-FRICTION scrutiny rule on public IPs.
3. **Staleness**: LAN IP is detected ONCE at daemon start; on DHCP change the listener stays bound to the dead IP and the 120s refresh loop keeps re-advertising the stale addr until daemon restart. No re-detect/rebind story. (Ephemeral port-0 churn across restarts is fine — re-advertised within ≤120s; dial failures are recorded, not swallowed.)
4. **windows-self-hosted leg actually RAN and FAILED** (not QUEUED/offline): `codex_hook_commands` ×10, "daemon did not become ready". Pre-existing — the same leg fails identically on the #1083 push run (before this PR) and the #1146 run (after) — so not introduced here, but the documented "bigmama offline" exception no longer describes reality; the leg is up and red and needs its own card.
5. **Process**: merge landed with no review/verdict comment on record. Squash-merge confirmed on rust-rewrite (30c46ea).

No action required on this PR itself (merged; #1146 already fixed the live-path gap). Findings 1–4 should become follow-up cards.

PR #1144, opened 2026-06-12
perf(daemon): O(1) room_tip IPC op — tip probe no longer scans the room (card a1562dbc)

joelteply:
## What / why

Card `a1562dbc` (P2): discovering a room's latest cursor (the "tip") required `Inbox { since: None, limit: 1 }`, which makes the daemon **replay the entire room** — `EventRouter::resume_from_cursor(channel, None)` materializes every durable envelope (deep sink page with `usize::MAX` + ring merge), then the handler throws away all but the last. O(n) in room history for a probe whose answer is a single cursor. Callers that only need "what is the newest cursor?" (reconnect watermarking, idle detection, work-board projection-cache validation) paid the full scan on every probe.

This adds a first-class **typed O(1) tip probe**, wired daemon → store, surfaced through airc-lib:

- **airc-bus**: `HotRing::newest_durable_cursor()` (back-to-front walk, bounded by ring capacity — constant wrt room history) and `EventRouter::durable_tip(channel)`: the ring's newest durable when present (eviction is oldest-first, so a ring durable IS the global durable tip — the sink only lags the ring via write-behind, never leads it), otherwise ONE indexed `DurableSink::head_cursor` row (`ORDER BY epoch,counter,event_id DESC LIMIT 1` on SQLite). A sink error **propagates** — no scan fallback. The trait's scan-shaped default `head_cursor` impl (page the whole channel, find the last) is **removed**: every impl must answer in constant work or fail loudly.
- **airc-ipc**: new typed `Request::RoomTip(RoomTipRequest { channel })` → `Response::RoomTip(RoomTipResponse { tip: Option<IpcCursor> })` — an intentional new op, NOT a flag on the `Inbox` scan. `DaemonClient::room_tip` helper. Returns the cursor value only; no envelope bytes ride along.
- **airc-daemon**: dispatch arm `room_tip` → `router.durable_tip`.
- **airc-lib**: `daemon_latest_transcript_cursor` (the work-board cache freshness probe, card 1291173d) migrated from the inbox scan to the typed probe — no event decode at all now.

Tip semantics are the **durable transcript tip** (matching what the inbox-scan path returned after its durable filter): a newer `StreamChunk`/ephemeral burst does not move the tip, so cache-cursor comparisons stay exact.

## O(n) → O(1) evidence

**Measured, live daemon over a Unix socket, real SQLite ORM, debug build** (`airc-daemon/tests/room_tip_probe.rs::deep_room_tip_probe_matches_inbox_scan_and_is_cheaper`, 5,000-event room, 20 probes each, same cursor returned by both paths):

```
card a1562dbc: room depth 5000, 20 probes each —
  inbox(limit:1) full-room scan: 99,129 µs/op
  room_tip probe:                    441 µs/op        (~225x)
```

The probe's 441 µs is dominated by the IPC connect/round-trip, not room depth.

**Structural proof, not vibes** (`airc-bus/tests/durable_tip.rs`): the sink is wrapped in a counting decorator; the tip probe on a 5,000-deep room (ring capacity 64, so depth genuinely lives in the sink) must complete with **zero `page` calls** (the scan primitive) and at most one `head_cursor` call. The no-fallback contract is enforced with a sink whose `page` *panics* and whose `head_cursor` errors: `durable_tip` surfaces the error, never scans.

## Tests

- `airc-bus/tests/durable_tip.rs` (6): empty room, single event, 5,000-deep constant-work proof, non-durable traffic doesn't move the tip, sink-index leg when ring has no durable (fresh-start shape), sink error is loud (no scan fallback).
- `airc-bus/src/ring.rs`: `newest_durable_cursor` skips non-durable tail / None on empty.
- `airc-daemon/tests/room_tip_probe.rs` (3): live-daemon tip tracking + room isolation + chunk immunity; deep-room probe/scan agreement + measured numbers; daemon **restart** — tip served from the SQLite index with an empty ring (write-behind drained via a polled persistence barrier, no blind sleep).
- `airc-ipc` wire pins (encode + decode literals): `{"op":"room_tip","channel":…}`, `{"kind":"room_tip","tip":{…}}`, empty-room shape `{"kind":"room_tip"}` (tip ABSENT, not null), and a pre-RoomTip `inbox` literal still decoding (additive-variant compat). **Mutation-tested**: 4 mutations (request tag rename, request field rename, response tag rename, response field rename) — each broke the corresponding pin, then restored. `IPC_PROTOCOL_VERSION` unchanged, matching the additive-variant precedent (card 7d5b6a65's `AttachCursorAdvanced`).

## Gate

- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` → clean
- `cargo fmt --all -- --check` → clean
- `cargo test --workspace -- --skip codex_hook` → 95 test binaries, all green (codex_hook skipped per macOS hang)

## Residual notes

- `Airc::latest_cursor` / `subscription_cursor` read the **local store** (already an indexed `LIMIT 1`) and don't branch to the daemon when attached — pre-existing semantics, out of scope here; worth a follow-up card if daemon-mode consumers need the daemon's view.
- The general `Inbox { since: None }` "most recent N" path still materializes the room before truncating; the tip probe removes its dominant caller, the rest is a separate paging card.

Card: a1562dbc-69cc-4b21-99c0-a310d718e951

### DIFF
--- a/crates/airc-bus/src/ring.rs
+++ b/crates/airc-bus/src/ring.rs
@@ -112,6 +112,23 @@ impl HotRing {
         self.slots.back().map(|slot| slot.env.cursor())
     }
 
+    /// **Card a1562dbc.** The cursor of the most-recent **`Durable`**
+    /// entry currently retained, if any — the transcript tip as seen
+    /// from RAM. Walks back-to-front, so the cost is bounded by the
+    /// ring capacity (a constant), never by room history.
+    ///
+    /// Eviction pops from the front (oldest first), so whenever the
+    /// ring holds *any* durable entry, the newest one in the ring IS
+    /// the globally newest durable event on this channel — the sink
+    /// only ever lags the ring (write-behind), never leads it.
+    pub fn newest_durable_cursor(&self) -> Option<Cursor> {
+        self.slots
+            .iter()
+            .rev()
+            .find(|slot| slot.env.delivery.is_durable())
+            .map(|slot| slot.env.cursor())
+    }
+
     /// The cursor of the oldest entry currently retained, if any. A
     /// `from_cursor` older than this means the ring cannot serve the full
     /// replay and the deep (sink) leg must cover `(from, oldest_in_ring)`.
@@ -187,6 +204,34 @@ mod tests {
         ring.mark_persisted(EventId::from_u128(1)); // counter 0
         assert_eq!(ring.len(), 2, "unpinned oldest reclaimed back to capacity");
         assert_eq!(ring.oldest_cursor().unwrap().seq.counter, 1);
+    }
+
+    /// Card a1562dbc: the durable tip skips non-durable tail entries
+    /// (a StreamChunk burst after the last chat message must not move
+    /// the transcript tip) and is `None` when no durable is retained.
+    #[test]
+    fn newest_durable_cursor_skips_non_durable_tail() {
+        let mut ring = HotRing::new(10);
+        assert_eq!(ring.newest_durable_cursor(), None, "empty ring");
+
+        ring.push(Arc::new(env_at(0, DeliveryClass::StreamChunk)));
+        assert_eq!(
+            ring.newest_durable_cursor(),
+            None,
+            "non-durable only ⇒ no transcript tip in RAM"
+        );
+
+        ring.push(Arc::new(env_at(1, DeliveryClass::Durable)));
+        ring.push(Arc::new(env_at(2, DeliveryClass::StreamChunk)));
+        ring.push(Arc::new(env_at(3, DeliveryClass::EphemeralWindow)));
+
+        let tip = ring.newest_durable_cursor().expect("durable tip");
+        assert_eq!(tip.seq.counter, 1, "tip is the durable, not the ring back");
+        assert_eq!(
+            ring.newest_cursor().expect("ring back").seq.counter,
+            3,
+            "sanity: the unfiltered newest IS the non-durable back"
+        );
     }
 
     #[test]
--- a/crates/airc-bus/src/router.rs
+++ b/crates/airc-bus/src/router.rs
@@ -378,6 +378,35 @@ impl EventRouter {
         self.inner.sink.head_cursor(channel).await.ok().flatten()
     }
 
+    /// **Card a1562dbc.** The channel's **durable transcript tip**: the
+    /// cursor of the newest `Durable` envelope on `channel`, or `None`
+    /// for a room with no durable history. Constant work regardless of
+    /// room depth — this is the O(1) probe behind the `room_tip` IPC op
+    /// (reconnect watermarking, idle detection, projection-cache
+    /// validation), replacing the "replay the whole room, keep the last
+    /// event" scan.
+    ///
+    /// Two constant-cost legs, NOT a perf fallback:
+    /// - the hot ring's newest durable entry (bounded by ring capacity).
+    ///   Eviction is oldest-first, so when the ring holds any durable,
+    ///   its newest durable IS the channel's global durable tip — the
+    ///   sink only lags the ring (write-behind), never leads it.
+    /// - otherwise the sink's [`DurableSink::head_cursor`] (one indexed
+    ///   row on SQLite). A sink error propagates — it is never papered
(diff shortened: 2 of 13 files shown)


joelteply:
## Sentinel review — card a1562dbc (adversarial, default-BLOCK)

Reviewed PR head `bda1c36` in an isolated temp worktree (`/tmp/sentinel-a1562dbc`, detached checkout; no production daemon touched).

### VERDICT: APPROVE

### EVIDENCE (all run in the temp worktree, not the author's)

**Production gate, verbatim:**
- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` → **clean**
- `cargo fmt --all -- --check` → **clean**
- `cargo test --workspace -- --skip codex_hook` → **95 test binaries, 0 failures**

**Wire-pin mutation tests (pins must pin something):**
- Mutation A: `#[serde(rename = "room_tip_v2")]` on `Request::RoomTip` → `room_tip_wire_bytes_are_pinned` + `room_tip_literal_wire_json_decodes` **FAILED** (2/2). Restored → green.
- Mutation C: `#[serde(rename = "newest")]` on `RoomTipResponse.tip` → `room_tip_response_wire_bytes_are_pinned_per_shape` **FAILED**. Restored → green.
- (Author independently ran 4 pin mutations pre-push — request tag/field, response tag/field — all failed as required.)

**O(1) claim is TESTED, not asserted:**
- Mutation B: replaced `EventRouter::durable_tip`'s sink leg with a silent full-room `page(channel, None, usize::MAX)` scan (the exact failure class the card kills) → **all 6** `airc-bus/tests/durable_tip.rs` tests **FAILED** (counting-sink decorator asserts zero `page` calls; the no-fallback test's sink `page` panics). Restored → green.
- The structural proof publishes 5,000 events through a ring of capacity 64 (depth genuinely in the sink) and requires zero scan calls + ≤1 indexed `head_cursor` row.
- Measured end-to-end (live daemon, Unix socket, SQLite ORM): **99,129 µs/op** for the legacy `Inbox{since:None,limit:1}` full-room scan vs **441 µs/op** for `room_tip` on the same 5,000-event room, identical cursor returned (~225x; probe cost is IPC-dominated, depth-independent).

**Design-doctrine spot checks:**
- New typed `Request::RoomTip`/`Response::RoomTip` variants — an intentional op, not a flag on the `Inbox` scan; cursor value only, no envelope copies.
- No fallbacks: `DurableSink::head_cursor`'s scan-shaped default impl was **removed** (every impl must answer in constant work); a sink index error propagates to `Response::Error`.
- Ring-leg soundness verified by reading `HotRing`: eviction is strictly oldest-first and durables enter at the back, so a ring-resident durable is provably the channel's global durable tip (sink only lags via write-behind).
- Durable-tip semantics match the migrated caller exactly (the old path filtered to durable before taking the last envelope); newer StreamChunk/ephemeral traffic does not move the tip — projection-cache cursor comparisons stay exact.

### RISK NOTES
- `bda1c36` (test-only delta) fixes a real first-run macOS CI failure: the 5,000-publish loop outran the bounded write-behind queue and the daemon shed with its loud §3.8 back-pressure error; the test now retries on saturation like a real producer. Re-verified at head: 3/3 green.
- `cargo test (windows-self-hosted)` failed in **dependency compile** (`windows_x86_64_msvc` build script blocked by an Application Control policy, os error 4551) — runner infrastructure on the self-hosted box, unrelated to this diff; the documented exception applies, hosted legs are the gate.
- Probe-then-sink read is not atomic with concurrent publishes; the returned tip can trail an in-flight publish by design — identical semantics to the scan it replaces. Not a blocker.
- `Airc::latest_cursor` / `subscription_cursor` still read the local store in daemon-attached mode (pre-existing; flagged in the PR body as follow-up card material).

Sentinel cleanup: temp worktree removed after verdict. No production daemon was started or stopped.

PR #1145, opened 2026-06-12
fix(registry): daemon gh-auth gate timed out on Windows (750ms < ~900ms) — THE cross-machine root cause

joelteply:
## The days-long cross-machine blocker, in one number
The account-registry loop's gh-auth gate (`gh_auth_ready`) gave `gh auth status` only **750ms**. On Windows that call runs **~900ms** (measured on bigmama: 881–950ms) — gh startup + an OS keyring lookup are slow. So the gate **timed out on every tick** and reported "not authenticated" *despite valid auth*. The daemon therefore **never published its account beacon**; the only beacons that landed were the manual `registry sync` fallback (no listener → `endpoints: none`). Same-account peers enrolled each other but **never routed**. That's the symptom the team's hit since power-on, and it's this timeout.

## Fix
- **Gate timeout 750ms → 5s.** A genuinely-unauthenticated `gh auth status` still fails fast, so the wider budget only costs latency on the slow-but-authed path it exists to allow.

## Robustness (varied daemon spawn contexts)
A DETACHED daemon descended from a bash launcher has different PATH/keyring reach than its parent:
- **`inject_gh_token`**: the authed parent resolves `gh auth token` and hands the daemon `GH_TOKEN`, so auth works even where the keyring is unreachable. Non-empty guard (an exported `GITHUB_TOKEN=""` must not short-circuit). Derived from the live credential at spawn — no a-priori knowledge.
- **`resolve_gh_bin`**: daemon resolves gh's full path (PATH, then known install dirs) for its gate + store, so a unix-format / install-dir-less PATH can't make `Command::new("gh")` silently fail.

## Verified
On bigmama: with the timeout fixed the daemon **stops logging `account_registry_publish_failed`** and the loop publishes. The 750ms-vs-900ms timing is decisive. Pairs with #1143 (endpoint-in-beacon) so same-account auto-dial actually routes — the keystone close. (Likely fixes the Mac daemons too — fable, check your daemon log.)

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -380,6 +380,7 @@ pub async fn ensure_daemon_running(
         .stdin(Stdio::null())
         .stdout(Stdio::from(stdout))
         .stderr(Stdio::from(stderr));
+    inject_gh_token(&mut command);
     detach_daemon(&mut command);
     let child = command.spawn()?;
     let daemon_pid = child.id();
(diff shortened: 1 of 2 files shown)


joelteply:
## SENTINEL VERDICT: APPROVE (post-merge audit — PR was merged before any sentinel verdict landed)

This PR hit the merge queue and merged (49aaf64) with **zero review comments**. That's a process gap — posting the full audit anyway as the record.

### EVIDENCE (run at PR head 3af387d in a temp worktree)
- `cargo fmt --all -- --check` — PASS
- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` — PASS
- `cargo test --workspace -- --skip codex_hook` — PASS (one local failure: `airc-ipc bench_ipc_frame_round_trip_send_with_headers`, a wall-clock perf assert that flaked under parallel load; green in isolation and in both hosted CI legs; zero overlap with this PR's files)
- **Mutation test** — the PR shipped NO tests, so I wrote three (slow-but-authed gh stub at ~1s must pass the gate; unauthed stub fast-fails; hung stub killed at the 5s budget). At 5s: 3/3 pass. **Mutated back to 750ms: `gate_allows_slow_but_authed_gh_900ms_class` FAILS** — the pin bites and reproduces the exact bug. These tests should be upstreamed (see risk note 1).
- Loud-failure check — PASS: gate-skip emits a typed `AccountRegistryPublishFailed` warn diagnostic per tick; token-resolution failure prints actionable stderr (`run gh auth login, or set GH_TOKEN`). No new silent fallback; degrade-to-no-rendezvous is the pre-existing optional-transport design.

### ROOT-CAUSE READ
The Windows math is decisive: 750ms budget < measured 881–950ms `gh auth status` ⇒ gate timed out **every tick** ⇒ daemon never published its beacon ⇒ only endpoint-less manual-sync beacons landed ⇒ enrol-but-never-route. Coherent with the observed symptom and correctly paired with #1143. But note this is really **three fixes in one** (timeout, keyring reach via GH_TOKEN injection, PATH reach via absolute gh path) — Mac daemon failures would be the keyring/PATH legs, not the 750ms. Calling the timeout "THE" root cause is overclaimed for non-Windows; cross-machine convergence still needs live two-way verification.

### RISK NOTES (none blocking, all worth follow-up)
1. **No tests shipped; 5s is an inline magic number.** Doctrine prefers a named constant with rationale + a deterministic pin. My stub-based tests prove this is cheaply testable — upstream them and name the constant.
2. **Stale-token override hazard.** `GH_TOKEN` is snapshotted at daemon spawn; daemons are long-lived (restart only on build/protocol change), and gh prefers env token over keyring. A rotated/revoked token (known to happen on the shared joelteply identity) turns a working-keyring daemon into a broken one until restart. Consider re-resolving per tick or falling back to keyring when the env token fails auth.
3. **`resolve_gh_bin` ignores `AIRC_GH_BIN`.** `GhAccountRegistryStore::new` honors the operator override, but `run_daemon` now clobbers it via `.with_bin(...)` whenever any candidate gh responds. Check `AIRC_GH_BIN` first.
4. **Blocking `std::process::Command` calls on async paths** (`inject_gh_token`/`resolve_gh_token` inside async `ensure_daemon_running`; `resolve_gh_bin` in the daemon's registry task). One-shot spawn paths, low impact, but up to ~1s runtime stalls on Windows.
5. **Windows leg was RED at merge.** `cargo test (windows-self-hosted)` failed with `An Application Control policy has blocked this file. (os error 4551)` on the airc-bus test binary — environmental (WDAC blocking a fresh test exe on bigmama), not this PR's code. Still: a Windows-targeted fix merged with its only Windows CI leg red; the sole Windows verification is the author's manual bigmama claim. The runner's WDAC policy needs an exemption for the work dir or this leg stays useless.

PR #1146, opened 2026-06-12
fix(registry): always advertise the publishing node's own beacon (peers:0 → dialable)

joelteply:
## The keystone's last layer
The account-registry publish builds `peers` via `from_snapshot`, which only emits peers in `snapshot.live`. The publishing node's own presence is heartbeated for liveness, so a registry tick landing after its TTL leaves self in `stale` — and the document goes out with **peers: 0**. A same-account peer reading an empty doc discovers nothing: **enrol-but-never-route.**

Verified live on bigmama: even with the gh-auth gate now passing (#1145), the published doc was `peers: 0`.

## Fix
After `from_snapshot`, if our own beacon isn't already present, stamp a fresh one (we're definitionally live — we're publishing *right now*) carrying our dialable `route_endpoints`.

## Verified
The published doc went **peers: 0 → peers: 2**, with a beacon carrying `lan_tcp 192.168.1.232:<port>` (`HAS MY LAN ENDPOINT: True`). That's the daemon's machine-account identity advertising its live listener.

Completes the same-account auto-discovery chain: **#1143** (daemon advertises its LAN endpoint) + **#1145** (gh-auth gate timeout) + this (self-beacon in the published doc). Same-account machines can now read each other's docs and get a dialable address. Likely fixes the Mac daemons identically — @fable, once this + #1145 are on rust-rewrite, your `registry sync` should auto-route to 5090.

### DIFF
--- a/crates/airc-lib/src/account_registry.rs
+++ b/crates/airc-lib/src/account_registry.rs
@@ -254,23 +254,55 @@ impl Airc {
             crate::time::now_ms()?,
         )
         .await?;
-        let mut peer_specs = vec![PeerSpec {
-            peer_id: self.inner.identity.peer_id,
+        let self_id = self.inner.identity.peer_id;
+        let self_spec = PeerSpec {
+            peer_id: self_id,
             pubkey: self.inner.identity.keypair.public_bytes(),
-        }];
+        };
+        let mut peer_specs = vec![self_spec.clone()];
         for stored in airc_trust::load(&self.inner.wire_root).await? {
             peer_specs.push(PeerSpec {
                 peer_id: stored.peer_id,
                 pubkey: stored.pubkey_bytes()?,
             });
         }
-        let endpoints = vec![(self.inner.identity.peer_id, self.route_endpoints()?)];
-        Ok(AccountRegistryDocument::from_snapshot(
+        let self_endpoints = self.route_endpoints()?;
+        let endpoints = vec![(self_id, self_endpoints.clone())];
+        let mut document = AccountRegistryDocument::from_snapshot(
             &snapshot,
             peer_specs,
             endpoints,
             crate::time::now_ms()?,
-        ))
+        );
+
+        // KEYSTONE (card 0ac…/#35): the publishing node MUST advertise its
+        // OWN beacon for same-account cross-machine discovery — that's the
+        // entire point of publishing. `from_snapshot` only emits peers in
+        // `snapshot.live`, but our own presence is heartbeated for
+        // liveness and a registry tick can easily land after its TTL,
+        // leaving us in `stale`. The document then went out with peers:0
+        // (verified live on bigmama), so a same-account peer reading it
+        // discovered nothing — enrol-but-never-route, the keystone's last
+        // layer. Stamp a fresh self-beacon (we are definitionally live: we
+        // are publishing right now) carrying our dialable endpoints.
+        if !document.peers.iter().any(|peer| peer.peer_id() == self_id) {
+            let presence = crate::coordinator::beacon_now(
+                self_id,
+                self.inner.home.clone(),
+                snapshot.live_channels.clone(),
+                std::process::id(),
+                crate::time::now_ms()?,
+            );
+            document.peers.push(AccountPeerBeacon {
+                presence,
+                peer_spec: self_spec,
+                endpoints: self_endpoints,
+            });
+            document
+                .peers
+                .sort_by_key(|peer| peer.peer_id().to_string());
+        }
+        Ok(document)
     }
 
     pub async fn publish_account_registry(


joelteply:
## Sentinel post-merge audit — review of record

This PR merged via the merge queue with **zero review comments**; no sentinel saw it pre-merge. This audit is the review of record for 376c61f.

**VERDICT: NEEDS-FOLLOW-UP** (fix is correct and keystone-worthy; it shipped completely unpinned, plus two doctrinal residues)

### What the fix does right
- **Self-beacon is SoT-derived where it matters.** `peer_id`/`pubkey` come from `self.inner.identity`, endpoints from `self.route_endpoints()` (the live in-process `RouteEndpointTable` that #1143's daemon glue populates via `listen_lan`). The publisher no longer depends on its own heartbeat surviving the presence TTL — "we are definitionally live: we are publishing right now" is the correct doctrine and the comment in the code says so plainly.
- **No duplicate-self accumulation.** The document is rebuilt from scratch every tick and the insertion is guarded by `!document.peers.iter().any(|p| p.peer_id() == self_id)`; when self IS coordinator-live, `from_snapshot` already carries the same fresh `self_endpoints` (passed in the endpoints map), so both branches agree. `validate()` passes (presence/spec ids match). Import side skips `peer_id == self`, so a stale self-beacon refreshed back from the rendezvous can't re-enter local state either.

### MUTATION RESULT — the fix is unpinned (top follow-up)
Bypassed the self-beacon insertion (`if false && …`) and re-ran the tests: **everything passes**. `cargo test -p airc-lib` (the crate containing the fix and every registry test): 18/18 result sets green, 0 failures — `sqlite_registry_bridges_two_isolated_machine_homes`, `sync_once_auto_enrols_same_account_peer_zero_manual_add`, `import_registry_enrols_peer_and_presence`, and `stored_endpoint_dial` all pass with the keystone line dead. (A full-workspace mutation run also got through ~20 binaries green before stalling on two unrelated airc-cli work-board tests — an environmental hang under the leaked-daemon load of repeated suite runs; both pass at HEAD.) Same failure class as #1143's unpinned daemon glue (sentinel comment 4686453856, mutation 2). Root cause: every test that exercises the publish path (`sqlite_registry_bridges_two_isolated_machine_homes`, `sync_once_auto_enrols_same_account_peer_zero_manual_add`) calls `airc.join("general")` immediately before publishing, so self is freshly coordinator-live and `from_snapshot` emits it without the fix. The bug's actual trigger — registry tick landing after the heartbeat TTL, self in `stale` — is reproduced by no test. The keystone behavior verified live on bigmama can silently regress.

### #33 interaction (manual `airc registry sync`, endpoint-less CLI path)
- **Strict improvement, not a fix.** Before: a CLI sync (no listener → empty `RouteEndpointTable`) published `peers:0` to this machine's gist — readers got nothing at all. After: it publishes a self-beacon with `peer_spec` (pubkey) but `endpoints: []` — readers can at least enrol/trust, and the import guard ("empty beacons leave the column alone", card 625abe6d) means a previously-learned endpoint on the trust record is NOT wiped.
- **Residue remains.** The CLI sync still overwrites this machine's own gist (`gh gist edit` on the sentinel id) with an endpoint-less doc. `GhAccountRegistryStore::refresh` picks the single NEWEST doc per mesh identity (it does not union-merge, despite the module doc saying "refresh merges them all"), so a FRESH machine whose refresh lands on that doc enrols the peer but gets no dialable endpoint — enrol-but-never-route, #33's exact shape, now narrowed to first-contact-after-manual-sync.

### Other risk notes
- **`subscribed_channels` on the stamped self-beacon is still presence-derived.** `beacon_now(…, snapshot.live_channels, …)` where `live_channels = unique_channels_in(&live)` — in the exact stale-self scenario this fix targets, `live` is empty, so the self-beacon (and the doc's `channels`) goes out as `[]` even though `self.subscriptions()` is the local SoT and is non-empty. Identity/key/endpoints follow the publisher-is-alive doctrine; the channel list doesn't yet. Affects same-account channel convergence, not routing.
- Two `now_ms()` calls (doc timestamp vs beacon heartbeat) can differ by a tick — harmless, noted for completeness.

### PRODUCTION GATE (all run in a clean worktree at 376c61f)
- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` — clean (3m14s)
- `cargo test --workspace -- --skip codex_hook` — all green (95 result sets, 0 failures; the known airc-ipc bench flake did not fire)

### Follow-up items (card-ready)
1. **Pin the stale-self publish path** — test: open an `Airc` handle, publish presence with an OLD timestamp (or advance `now_ms` past the TTL) so self lands in `snapshot.stale`, then assert `account_registry_document()` contains exactly one self-beacon carrying `route_endpoints()`. Must fail when the insertion block is removed (mutation-verified).
2. **Give the manual `registry sync` CLI path a dialable endpoint** — bind a LAN listener (same `detect_lan_ip` + `listen_lan` glue the daemon registry loop uses) or read back this machine's previously-published endpoints before publishing, so a manual sync stops overwriting the gist with an endpoint-less self-beacon (#33 closure).
3. **Derive self-beacon `subscribed_channels` from local subscriptions** — use `self.subscriptions()` instead of presence-derived `snapshot.live_channels` for the stamped self-beacon, completing the publisher-is-alive doctrine for the channel list.
4. **(observed in passing) `GhAccountRegistryStore::refresh` doc-comment vs behavior** — module doc promises "refresh merges them all"; implementation picks the single newest doc. Either implement the union-merge across per-machine gists or fix the contract comment — a 3-machine account currently converges only pairwise through the newest writer.

PR #1151, opened 2026-06-12
fix(registry): stale-self publish pins + union-merge endpoint retention + dialable manual sync (cards cbbcf18d, 4b6a0ffa)

joelteply:
## Registry: stale-self publish pins + convergence + dialable manual sync

Cards (one PR, sibling P1s touching the same registry files):
- `cbbcf18d-f1cf-4378-adb8-9885148397d3` — registry: pin the stale-self publish path + self-beacon channels from local SoT (post-#1146 audit — mutation SURVIVED)
- `4b6a0ffa-3266-4dce-a572-1d5a83e94f9d` — registry convergence: union-merge across per-machine gists + manual sync publishes a dialable endpoint (#33 closure)

Source: #1146 post-merge audit (review of record), https://github.com/CambrianTech/airc/pull/1146#issuecomment-4686802126

### Item 1 — pin the stale-self publish path (the keystone fix had NO failing test)
`account_registry.rs::stale_self_publish_stamps_one_self_beacon_with_endpoints_and_channels`: joins a channel, then OVERWRITES self's presence beacon with a past-TTL heartbeat so self lands in `snapshot.stale` (asserted as a precondition — every prior test kept self coordinator-live and bypassed the #1146 branch). Asserts `account_registry_document()` carries exactly one self-beacon with `route_endpoints()`.

### Item 2 — self-beacon `subscribed_channels` from local SoT
The stamped self-beacon now derives its channel list from `self.subscriptions()` (local single source of truth) instead of presence-derived `snapshot.live_channels` (empty in the exact stale-self scenario). The document's `channels` union also picks up the local subscriptions. Pinned in the same test.

### Item 3 — registry convergence (union-merge across per-machine gists)
`GhAccountRegistryStore::refresh` already union-merges via `merge_registry_documents` since #1150; this PR adds the missing cross-writer pin (`refresh_unions_peers_across_writer_gists_keeping_endpoints`: peer present only in an OLDER writer's gist survives, with endpoints — fails if refresh reverts to newest-only) and EXTENDS the merger with endpoint retention: per peer, the freshest beacon still wins presence, but if it is endpoint-less the freshest NON-EMPTY endpoint set is retained (`merge_retains_endpoints_when_fresher_beacon_is_endpointless`). This closes the "endpoint-less manual-sync doc shadows a good daemon doc" half of #33 on the READER side, mirroring the import guard's "empty beacons leave the column alone" doctrine.

[...]

PR #1153, opened 2026-06-12
fix(registry): union peers across all nodes' docs — the cross-machine convergence bug

joelteply:
## The keystone bug

The account-registry reader (`gh_account_registry::refresh()`) kept a single **`best`** document — the one with the newest `generated_at_ms` — and discarded every other matching doc.

But the design is **one gist per machine** (the module doc says so: *"Refresh merges all matching gists… a scope on a third+ machine sees both other machines' beacons"*). So "newest doc wins" meant **whichever node synced last erased every other machine's beacon.** No two nodes ever saw each other — the grid could not converge *by construction*, independent of endpoint/gh-auth/pollution. This is why every `registry sync` reported `enrolled 1 (only this node visible)`.

## Fix

Union peers across **all** matching docs: dedup by `peer_id`, keep the freshest beacon per peer (scored by source doc `generated_at_ms`), union channels. Makes the code match its own documented contract.

## Proof (bigmama, 5090)

```
before:  registry sync → enrolled 1 peer (only this node visible)
after:   registry sync → enrolled 45 peers from the merged registry
```
…and the Mac peer's **live** beacon (`5159a48b @ 192.168.1.176:57651`) now imports — first time this box has ever seen another machine on the account mesh.

## Note
`cargo install` (release) is currently failing on `rcgen` on the Windows box (debug builds clean — the fix is proven via the debug binary's `registry sync`). Separate issue; flagging it. Mac/Ubuntu CI is the authoritative check here.

### DIFF
--- a/crates/airc-lib/src/gh_account_registry.rs
+++ b/crates/airc-lib/src/gh_account_registry.rs
@@ -52,9 +52,9 @@ use tokio::time::{timeout, Duration};
 use airc_store::{SqliteEventStore, StoredAccountRegistryGistSentinel};
 
 use crate::account_registry::{
-    AccountRegistryDocument, AccountRegistryError, AccountRegistryStore,
+    AccountPeerBeacon, AccountRegistryDocument, AccountRegistryError, AccountRegistryStore,
 };
-use crate::subscriptions::MeshIdentity;
+use crate::subscriptions::{ChannelName, MeshIdentity};
 use crate::time;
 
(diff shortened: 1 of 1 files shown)


joelteply:
Duplicate of #1151 (union-merge already merged to rust-rewrite). Re-derived the same fix in isolation off a stale local rust-rewrite — closing. Independent confirmation the root cause was right.

PR #1150, opened 2026-06-12
fix(registry): hermetic gate — test daemons must never reach the production gh account rendezvous (card d793c242)

joelteply:
### Live evidence
2026-06-12 ~01:11Z: a Windows **test** daemon with scope_home `C:\Users\green\AppData\Local\Temp\tmp.YYavgmVUxz\.airc` published itself to the REAL joelteply account rendezvous (gist `1214fb43d2c00d667c4712e6023b2165`). Hermetic daemons inherit the operator's working gh auth (`ensure_daemon_running` even injects `GH_TOKEN`), so the account-registry loop happily publishes test identities to production. Consequences: (a) reader-merge enrols PHANTOM TEST PEERS and the #1146 import loop auto-dials their garbage endpoints; (b) duplicate `airc-account-mesh-registry` gists proliferate — 5+ under joelteply, because create-on-miss fires every time a fresh writer identity appears.

### What each layer does

**1. Hermetic gate (the P1 core)** — `account_registry_block(scope_home)` in `gh_account_registry.rs`:
- **Intentional layer**: `AIRC_DISABLE_ACCOUNT_REGISTRY` env var, honored by the daemon (loop never spawns, ONE loud `account-registry loop DISABLED — …` line), by the per-tick gate in `run_registry_refresh_loop`, and by `airc registry sync` (prints `registry sync REFUSED (hermetic gate): …`). Set automatically by the test-daemon spawn helpers (`daemon_route_refresh.rs`, `daemon_lifecycle.rs`, `test/public_installed_runtime_proof.sh`) and inherited by daemons spawned via `ensure_daemon_running`.
- **Defense in depth**: a scope_home resolved under a temp dir (canonicalized `std::env::temp_dir()` prefix + cross-platform markers incl. `AppData\Local\Temp`) refuses BOTH publish and refresh.
- **No silent path**: the gate is enforced at daemon startup, per loop tick (typed `GateBlock`, warn diagnostic `account_registry_hermetic_skip`), and INSIDE `GhAccountRegistryStore` itself (`HERMETIC GATE: refusing …` error) — `scope_home` is a required constructor argument, so no gh-backed store can exist ungated. Every suppression logs the exact reason.

**2. Writer-doc find-or-update** — same writer → same gist, update in place:
- New gists are created as `airc-account-mesh-registry.<writer-key>.json` where the writer key derives from MACHINE identity (hostname + user, sanitized), not the rotating peer/mesh identity.
[...]

PR #1155, opened 2026-06-12
fix(transport): delivery acks + loud undeliverable diagnostics for lan-send silent drops (card 39d37629)

joelteply:
## Root cause (card 39d37629 — live repro 2026-06-12 02:36Z)

`lan-send` printed `sent over lan-tcp to general` (exit 0) the moment the frame bytes were flushed to the TLS socket. Forensics on the receiving machine (this Mac, the live repro receiver) show the frame was **accepted at TLS and persisted nowhere**: the machine store (`~/.airc/events.sqlite`) has a 5090 message at 02:32:10Z and then nothing until 19:01Z, and a sweep of every `events.sqlite` on the box (scope stores, temp homes) finds no event in the 02:25–02:50Z window beyond `subscription_advanced` ticks.

Mapping the receive path exposed an entire class of **silent drop points between transport accept and room-transcript persistence**, any of which matches the repro:

1. `lan_tcp` `read_loop` decode failure (`serde_json::from_slice` err) — connection dropped, frame gone, **zero diagnostics** (build-skew schema drift lands here).
2. `dispatch_to_subscribers` with no matching subscriber — frame parsed then dropped, **zero diagnostics** (a listener whose ingest task died, or a raw listener with no consumer).
3. `SignedTransport` verification failure — ingest loop warned **only when `AIRC_REPLAY_WARN` was set**, i.e. silent by default.
4. ingest `store.append` failure — stderr diagnostic only; sender never learns.
5. frame persisted onto a channel the receiving scope never bound — durable but invisible to every transcript surface (the store-split shape: the daemon's LAN handle persists to the machine store while scope transcripts read scope stores — observed live with the 02:32:10Z message, present in `~/.airc/events.sqlite` but absent from `/Users/joel/Development/airc/.airc/events.sqlite`).

Sender-side `Ok` was load-bearing only down to the kernel send buffer. Earlier finding (2026-06-10 23:52Z) confirmed lan-send frames are point-to-point and bypass the route layer, so no other layer could catch the loss.

## Fix, layer by layer

**airc-protocol (`delivery_ack.rs`, new):** typed ack vocabulary that is wire-compatible by construction — no new `FrameKind`, no new top-level shape. Requesting senders set header `airc.delivery_ack=request` on an ordinary frame; receivers answer with a Control frame whose JSON body is `DeliveryAck {for_event, receiver, outcome: delivered{channel, cursor} | undeliverable{reason}}` under header `airc.delivery_ack=response`. Old peers decode both directions as ordinary frames; old receivers simply never ack and the new sender reports **no-ack** distinctly. Literal-JSON wire pins for both variants, the reason labels, and the header constants.

**airc-transport:** `read_loop` decode failure and zero-subscriber dispatch now emit `frame_undeliverable` diagnostics (`decode_failure` / `no_subscriber`); new `LanTcpAdapter::send_to(peer, frame)` unicast for the ack return path (same flush-to-wire contract as `send`).

**airc-lib:** `append_received_frame` intercepts ack responses ahead of persistence (receipts, not transcript content) and fans them to waiters; for ack-requesting frames it answers **only after the persistence decision** — `delivered{channel,cursor}` requires the append to have committed AND the frame's channel to be bound in the receiving scope's subscription set; unbound channel → `undeliverable{unknown_channel}` + loud diagnostic (frame stays persisted, diagnostic says so); append failure → `undeliverable{persist_failed}`; live verification failures are now always loud (the `AIRC_REPLAY_WARN` gate is gone for the live ingest path — replay skips keep their dedicated codes). New `Airc::send_with_delivery_ack(text, headers, timeout)`; new injectable per-handle diagnostic sink so tests assert emissions instead of scraping stderr.

**airc-cli:** `lan-send` now waits for the receipt (`--ack-timeout-ms`, default 10 000) and prints the typed outcome; **anything that is not `delivered` exits nonzero** — including no-ack, which names the old-build possibility explicitly.

**Routed daemon path:** untouched — nothing in the ack machinery runs unless the request header is present, which only the point-to-point verb sets. Routed-ack filed as follow-up.

## Why the ack cannot false-positive

`conclude_delivery_ack` is reachable only from the two post-persistence arms of `append_received_frame` (`store.append` returned `Ok`/`DuplicateEventId`, or the broadcast ring proved a prior persist). The delivered e2e test additionally reads the event back out of the **receiver's** store, so an "ack on accept" regression fails the test even if the ack plumbing is intact (proven by mutation 2 below).

## Mutation evidence (all restored, suite green after each)

1. rename wire field `for_event` → `for_event_id`: `delivered_ack_wire_pin` + `undeliverable_ack_wire_pin` **FAILED**.
2. rename `unknown_channel` wire label → `channel_unknown`: pins + label test **FAILED**.
3. skip `store.append` on ingest while still acking: `lan_send_yields_delivered_ack_only_after_receiver_persistence` **FAILED** (store read-back caught the false delivered).
4. suppress the unknown-channel `frame_undeliverable` emission: `send_to_unbound_channel_is_undeliverable_with_loud_receiver_diagnostic` **FAILED**.

## Tests

- `crates/airc-protocol/src/delivery_ack.rs` — 6 unit/pin tests.
- `crates/airc-lib/tests/lan_delivery_ack.rs` — 3 e2e tests over real TLS LAN links between hermetic temp-home handles (no production daemon or route touched): delivered-with-cursor + receiver-store read-back; unbound-channel undeliverable + sink-asserted diagnostic; hollow receiver (accepts, never persists/acks — the killed-mid-send / old-build wire shape) → NoAck.

Production gate: clippy (workspace, `-D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm`) clean; `cargo fmt --all -- --check` clean; `cargo test --workspace -- --skip codex_hook` green.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,6 +281,7 @@ name = "airc-transport"
 version = "0.1.0"
 dependencies = [
  "airc-core",
+ "airc-diagnostics",
  "airc-protocol",
  "async-trait",
  "ed25519-dalek",
--- a/crates/airc-cli/src/cli.rs
+++ b/crates/airc-cli/src/cli.rs
@@ -452,8 +452,10 @@ pub enum Command {
         replay: bool,
     },
 
-    /// Same-LAN secure send: dial a peer over TLS and send a single
-    /// text frame to the current room's channel.
+    /// Same-LAN secure send: dial a peer over TLS, send a single
+    /// text frame to the current room's channel, and wait for the
+    /// receiver's typed delivery ack (card 39d37629). Exit 0 only on
+    /// `delivered`; undeliverable or no-ack outcomes exit nonzero.
     LanSend {
         /// Address of the listening peer (e.g. `127.0.0.1:7474`).
         #[arg(long)]
@@ -461,6 +463,10 @@ pub enum Command {
         /// UUID of the listening peer (for cert pinning).
         #[arg(long)]
         expected_peer: String,
+        /// How long to wait for the receiver's delivery ack before
+        /// reporting no-ack (older receivers never ack).
+        #[arg(long, default_value_t = 10_000)]
+        ack_timeout_ms: u64,
         /// Message body.
         text: String,
     },
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -1162,22 +1162,20 @@ pub async fn run_daemon(
             .map(|ip| (ip, "Tailscale"))
             .or_else(|| crate::network_commands::detect_lan_ip().map(|ip| (ip, "LAN")));
         match advertise {
-            Some((ip, kind)) => {
-                match airc.listen_lan(std::net::SocketAddr::from((ip, 0))).await {
-                    Ok(addr) => {
-                        eprintln!(
-                            "airc daemon: advertising {kind} endpoint {addr} in the account registry"
-                        );
-                    }
-                    Err(error) => {
-                        eprintln!(
-                            "airc daemon: {kind} listener bind failed ({error}) — account beacon \
+            Some((ip, kind)) => match airc.listen_lan(std::net::SocketAddr::from((ip, 0))).await {
+                Ok(addr) => {
+                    eprintln!(
+                        "airc daemon: advertising {kind} endpoint {addr} in the account registry"
+                    );
+                }
+                Err(error) => {
+                    eprintln!(
+                        "airc daemon: {kind} listener bind failed ({error}) — account beacon \
                              carries no endpoint; this node reaches the mesh by dialing out / \
                              relay but is not itself dialable"
-                        );
-                    }
+                    );
                 }
-            }
+            },
             None => {
                 eprintln!(
                     "airc daemon: no Tailscale or routable LAN IPv4 detected — account beacon \
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -873,12 +873,22 @@ pub async fn run_listen(
 }
 
 /// `lan-send` — TLS-wrapped single-shot send to a remote peer, on
-/// the current room's channel.
+/// the current room's channel, with a delivery-ack wait.
+///
+/// Card 39d37629: "sent" used to mean "bytes flushed to the TLS
+/// socket" — a receiver could accept the frame and silently lose it
+/// before transcript persistence (live repro 2026-06-12 02:36Z: 5090
+/// → mac printed "sent over lan-tcp to general", exit 0; the frame
+/// reached NO store on the receiving machine). The verb now requests
+/// a typed ack that the receiver emits only AFTER its persistence
+/// decision, waits up to `ack_timeout_ms`, prints the typed outcome,
+/// and exits nonzero for anything that is not `delivered`.
 pub async fn run_lan_send(
     home: &Path,
     peers: Vec<PeerSpec>,
     to: std::net::SocketAddr,
     expected_peer: PeerId,
+    ack_timeout_ms: u64,
     text: &str,
 ) -> Result<(), Box<dyn std::error::Error>> {
     // Card bf7c30e2: fail FAST with a self-diagnosing error before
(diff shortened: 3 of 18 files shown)


joelteply:
Rebased onto rust-rewrite HEAD 6b0cce0 and added a mechanical `cargo fmt` commit: upstream HEAD itself fails `cargo fmt --all -- --check` (drift introduced by #1154 — verified on a clean checkout of 6b0cce0), so every PR merge ref inherited a red fmt leg. No semantic change; rustfmt output only.

joelteply:
## SENTINEL (producer-run, inline) — card 39d37629

**VERDICT: APPROVE**

Reviewed at PR head `40c4873` in a throwaway worktree (`/tmp/sentinel-39d37629`, separate `CARGO_TARGET_DIR`), default-block stance.

### EVIDENCE

Production gate, run verbatim in the sentinel worktree at `40c4873`:
- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` → exit 0
- `cargo fmt --all -- --check` → exit 0
- `cargo test --workspace -- --skip codex_hook` → exit 0, zero `FAILED`/`failures:` lines in the captured log

FRESH mutation re-runs (sentinel worktree, applied → observed → restored → re-green):
- **A. ack-on-accept**: replaced the ingest `store.append` result with a hardcoded `Ok` (persistence skipped, ack still emitted) → `lan_send_yields_delivered_ack_only_after_receiver_persistence` **FAILED** (the test reads the event back out of the receiver's store, so a delivered-ack that lies about persistence cannot pass).
- **B. suppressed diagnostic**: dropped the unknown-channel `frame_undeliverable` emission → `send_to_unbound_channel_is_undeliverable_with_loud_receiver_diagnostic` **FAILED** (asserts through the injected `MemoryDiagnosticSink`).
- **C. wire-shape drift**: `#[serde(rename)]` on `DeliveryAck.for_event` → both literal-JSON pins **FAILED**.

Ack false-positive audit (code path, not just tests): `DeliveryOutcome::Delivered` has exactly **one** production constructor — `conclude_delivery_ack` (crates/airc-lib/src/delivery_ack.rs), bound-channel branch — and `conclude_delivery_ack` is reachable from exactly two arms of `append_received_frame` (crates/airc-lib/src/transport.rs): after `store.append` returned `Ok`/`DuplicateEventId`, and the broadcast-ring branch (ring entries are only marked after a successful persist in `append_sent_frame`, and before-append marks are revisited below). The persist-failure arm constructs only `Undeliverable{persist_failed}`. No path emits `delivered` on transport accept.

Wire compatibility: no new `FrameKind`, no new top-level wire shape — request marker and response are headers + JSON body on existing frame kinds. Old receiver + new sender → no ack → sender reports the typed no-ack outcome (nonzero exit, names the old-build possibility). Old sender → no header → zero new machinery runs; routed daemon sends unaffected (verified by the untouched daemon/route test suites in the gate run).

Hermeticity: the new e2e tests use `TempDir` scope homes (their temp-rootedness blocks the account-registry gh path per #1150's gate), bind only `127.0.0.1:0`, run fully in-process, and spawn no child processes — nothing dials or touches the production lan-tcp route or daemons. Verified the only `airc` listeners on this machine during the run were pre-existing PIDs from other sessions.

### RISK NOTES (non-blocking)

1. **Ring-marked-but-unpersisted edge**: `append_received_frame` marks the broadcast ring *before* the append; if the append then fails AND the same `event_id` is re-sent on the same receiver process, the ring branch would ack `delivered` without a persist. `lan-send` generates a fresh `event_id` per invocation so the verb path cannot hit this; noted for the routed-ack follow-up (card 1998f6cb).
2. **Verification-failure loudness**: live-frame verify failures now always emit (previously gated behind `AIRC_REPLAY_WARN`). Live subscriptions use `from_cursor: None` (no replay backlog) so steady-state flood risk is low, but a hostile/misconfigured peer can now generate one stderr line per bad frame.
3. **Cross-store delivered-vs-visible**: a delivered ack is truthful for the *receiving handle's* scope; the same-machine store-split (machine store vs project scope stores) can still make a delivered frame invisible to an operator scope — filed as card 4132f48c with forensic evidence from this card's live repro.
4. The fmt commit (`40c4873`) touches two files outside the card scope; it is pure rustfmt output fixing pre-existing drift introduced on the base by #1154 (upstream HEAD `6b0cce0` fails `cargo fmt --check` on its own — verified on a clean checkout).

PR #1156, opened 2026-06-12
fix(daemon): route inbound LAN frames into the owner-core router — store-split visibility (card 4132f48c)

joelteply:
## Card 4132f48c — store-split: delivered frames invisible to operator scopes

Forensic origin: #1155 (live repro 2026-06-12 02:32:10Z — a 5090→Mac message durable in `~/.airc/events.sqlite`, visible in NO operator scope transcript). This PR closes the last leg blocking visible cross-machine room delivery.

## Store topology map (who writes where, who reads where)

**Writes (inbound LAN, before this PR):**
- The daemon process owns two embedded `Airc` handles: the account-registry glue handle (binds the LAN listener, `home = machine ~/.airc`) and the route-refresh handle (outbound discovery dials, `home =` daemon launch scope).
- Frames arriving on either handle were ingested by `append_received_frame` → that handle's **SDK store** (`events` table of its home's `events.sqlite`) + that handle's private in-process broadcast. The machine store's `events` table is where the 02:32:10Z frame sat.
- Nothing reached the daemon's `EventRouter`.

**Reads (operator scopes):**
- `airc inbox` / `airc msg` / monitors all attach to the machine-singular daemon (`attached_airc` → `ensure_daemon_running` → `Airc::attach`). `page_recent`/`resume_from`/live subscribe go over IPC to the daemon's **router** — hot ring + `bus_events` durable tier of the machine ORM (`DaemonState` doc: "durable transcript lives in the router's sink, not the coordinator store").
- So the precise split is **SDK `events` table (write side) vs router hot-ring/`bus_events` (read side)** — same physical sqlite file on the machine home, two disjoint transcripts. Scope stores never see inbound frames either; the channel UUID (#1085, identical across scopes by design) never diverges — the *store placement* does.

**How local scope→scope already works (the mechanism mirrored):** a local send is `Request::Send/Publish` → `router.publish` → hot ring + live fan-out to every attached scope + write-behind to the machine ORM. One mechanism, fan-out at delivery, no per-scope copies.

## Design chosen: (c)-shaped — single transcript per machine, deliver inbound frames into the router

- **airc-lib `router_bridge` (new):** `InboundFrameSink` trait + `RouterInboundBridge`. With a sink installed (`Airc::set_inbound_frame_sink`), `append_received_frame` hands every received frame to the sink INSTEAD of the handle's scope store — the sink's verdict is the persistence decision the delivery ack reports. The bridge projects the frame onto the owner-core envelope shape (mirror of `handlers::publish_envelope` + `daemon_send_frame`, sender-minted `event_id` preserved) and publishes through the router.
- **airc-bus:** `EventRouter::publish_if_new` — idempotent ingest publish. Two legs: a router-level recent-event-ids window (every `publish` records its id, so a wire echo of a *locally published* event dedups too) and the new required `DurableSink::contains` (indexed PK probe; covers post-restart re-injects). Read-side `resume_from_cursor` dedup is adjacency-only `dedup_by`, so this write-side idempotency is load-bearing for the ring/live tiers.
- **airc-cli `run_daemon`:** installs the bridge on both daemon handles (registry listener — *before* the listener binds — and the route-refresh dialer).

**Why not (a) fan-out copies into every scope store:** contradicts §3.3 (ONE ORM per machine, no per-scope transcript), multiplies writes, and scope stores are not the read path the daemon-attached surfaces use. **Why not (b) read-side union:** every read pays a two-store merge across two incompatible cursor domains (SDK lamport vs generational seq) forever; the broken leg is the write side, not the read side.

**Cursor coherence:** bridged events get owner-assigned `(epoch, counter)` seqs exactly like local publishes — one total order, `room_tip` (#1144) and resume semantics untouched. **Wire compat:** no frame/transcript format change; no new ack vocabulary (#1155 labels reused).

## Ack truthfulness (extends #1155)

`delivered{channel,cursor}` now means **visible-to-subscribed-scopes**: published into the router AND at least one scope on the machine binds the channel (presence beacons in the coordinator store — every scope's join publishes its channel set there; stale beacons count because a subscription is durable scope state). No subscribed scope ⇒ `undeliverable{unknown_channel}` + loud `frame_undeliverable` diagnostic, frame kept durable (a late joiner replays it — proven in test). Duplicate arrival acks `delivered` (it IS delivered). Durability note: like every local send, the receipt is issued after ring+fan-out with write-behind enqueued (§3.8 ring pinned until persisted) — identical crash window to `handle_send`, no new class.

## Tests (hermetic: temp homes, in-process daemon + real TLS LAN links, RAII teardown; no production daemon/route touched)

`crates/airc-lib/tests/daemon_lan_visibility.rs`:
1. **THE card test** — inbound LAN frame (real `send_with_delivery_ack` over TLS) appears exactly once in a subscribed scope's transcript read through the daemon; ack says delivered.
2. Scope subscribed to a different room does NOT see it.
3. No duplicate: local-publish + LAN echo of the same event_id, and the same frame on two links — exactly one transcript copy.
4. Cursor monotonicity across interleaved local and bridged events.
5. Ack truthfulness: no subscribed scope ⇒ unknown_channel + loud diagnostic (MemoryDiagnosticSink-asserted); after a scope joins ⇒ delivered, and the earlier frame replays to the late joiner.

`crates/airc-bus/tests/publish_if_new.rs`: duplicate is a no-op at every tier (live fan-out order-pinned with a sentinel, sink row count); wire echo of a plain `publish` is a duplicate; already-durable id is duplicate even for a fresh router (restart shape, `DurableSink::contains` leg).

## Mutation evidence (each applied, observed failing, restored; suite green after)

1. **M1 — bridge severed** (`inbound_frame_sink()` → `None`): `inbound_lan_frame_is_visible_in_subscribed_scope_transcript` **FAILED**.
2. **M2 — dedup disabled** (`publish_if_new` always publishes): all 3 `publish_if_new` pins **FAILED** (live-tier sentinel ordering + sink row counts caught it; noted: the lib-level inbox read alone was insensitive because of read-side adjacent `dedup_by` — which is exactly why the bus-level pins exist).
3. **M3 — ack lies** (UnknownChannel verdict acks delivered): `delivered_ack_means_visible_to_subscribed_scopes_not_just_durable` **FAILED**.
4. **M4 — diagnostic suppressed** (unknown-channel `frame_undeliverable` removed): same test **FAILED** on the sink assertion.

## Production gate

- `cargo test --workspace -- --skip codex_hook` — green (exit 0)

Forensics: #1155. Card: 4132f48c-6131-4a22-9f73-d4b60dc56a63.

### DIFF
--- a/crates/airc-bus/src/lib.rs
+++ b/crates/airc-bus/src/lib.rs
@@ -51,6 +51,6 @@ pub use ephemeral::EphemeralCache;
 pub use error::{BusError, Result};
 pub use filter::Filter;
 pub use ring::HotRing;
-pub use router::{EventRouter, LagFlag, RouterConfig};
+pub use router::{EventRouter, LagFlag, PublishIfNew, RouterConfig};
 pub use seq::{EpochStore, InMemoryEpochStore, SeqSource};
 pub use sink::{DurableSink, InMemoryDurableSink};
--- a/crates/airc-bus/src/router.rs
+++ b/crates/airc-bus/src/router.rs
@@ -151,6 +151,17 @@ impl RecentEventIds {
         self.order.push_back(id);
         true
     }
+
+    /// Forget `id`. Used by [`EventRouter::publish_if_new`] when a
+    /// publish FAILED after the id was optimistically marked — a
+    /// poisoned entry would make a legitimate retry of the same frame
+    /// report `Duplicate` (and the receiver ack delivered) for an
+    /// event that never entered the router.
+    fn remove(&mut self, id: &airc_core::EventId) {
+        if self.seen.remove(id) {
+            self.order.retain(|other| other != id);
+        }
+    }
 }
 
 /// Outcome of [`EventRouter::publish_if_new`].
@@ -418,11 +429,33 @@ impl EventRouter {
         }
         // Off-lock durable probe. The id is already marked, so even if
         // this await parks, a concurrent same-id call short-circuits.
-        if self.inner.sink.contains(env.event_id).await? {
-            return Ok(PublishIfNew::Duplicate);
+        // On ANY failure from here on, unmark the id: a poisoned entry
+        // would turn a legitimate retry into a false `Duplicate` (and a
+        // false delivered ack) for an event the router never took.
+        let event_id = env.event_id;
+        let unmark = |router: &Self| {
+            let mut recent = router
+                .inner
+                .recent_ids
+                .lock()
+                .unwrap_or_else(|p| p.into_inner());
+            recent.remove(&event_id);
+        };
+        match self.inner.sink.contains(event_id).await {
+            Ok(true) => return Ok(PublishIfNew::Duplicate),
+            Ok(false) => {}
+            Err(error) => {
+                unmark(self);
+                return Err(error);
+            }
         }
-        let seq = self.publish(env).await?;
-        Ok(PublishIfNew::Published(seq))
+        match self.publish(env).await {
+            Ok(seq) => Ok(PublishIfNew::Published(seq)),
+            Err(error) => {
+                unmark(self);
+                Err(error)
+            }
+        }
     }
 
     /// The write-behind drain loop (§3.3 deliver-first / persist-async, §3.8
--- a/crates/airc-bus/src/router.rs
+++ b/crates/airc-bus/src/router.rs
@@ -114,6 +114,56 @@ struct WriteBehindItem {
     env: Arc<Envelope>,
 }
 
+/// Capacity of the router's recent-event-ids window (card 4132f48c).
+/// Sized well past the per-channel ring capacity so a same-moment echo
+/// of any event still in flight is always caught in memory; older
+/// re-injections fall through to the durable [`DurableSink::contains`]
+/// probe in [`EventRouter::publish_if_new`].
+const RECENT_PUBLISH_IDS_CAPACITY: usize = 4096;
+
+/// Bounded FIFO set of recently published event ids (card 4132f48c).
+/// `insert` answers "was this id already recorded?" and records it;
+/// capacity overflow rolls the oldest id off.
+struct RecentEventIds {
+    order: std::collections::VecDeque<airc_core::EventId>,
+    seen: std::collections::HashSet<airc_core::EventId>,
+}
+
+impl RecentEventIds {
+    fn with_capacity(capacity: usize) -> Self {
+        Self {
+            order: std::collections::VecDeque::with_capacity(capacity),
+            seen: std::collections::HashSet::with_capacity(capacity),
+        }
+    }
+
+    /// Record `id`. Returns `true` if it was new, `false` if already
+    /// present (already published through this router recently).
+    fn insert(&mut self, id: airc_core::EventId) -> bool {
+        if !self.seen.insert(id) {
+            return false;
+        }
+        if self.order.len() == RECENT_PUBLISH_IDS_CAPACITY {
+            if let Some(evicted) = self.order.pop_front() {
+                self.seen.remove(&evicted);
+            }
+        }
+        self.order.push_back(id);
+        true
+    }
+}
+
+/// Outcome of [`EventRouter::publish_if_new`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PublishIfNew {
+    /// The envelope was new: published with this owner-assigned seq.
+    Published(crate::Seq),
+    /// An event with this `event_id` already went through this router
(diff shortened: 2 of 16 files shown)


joelteply:
## SENTINEL — producer-runs-sentinel-inline (card 4132f48c)

**VERDICT: APPROVE** (default-BLOCK lifted on evidence below)

Audited in a fresh detached worktree (`/tmp/sentinel-4132f48c/wt` @ `0a38924`, separate `CARGO_TARGET_DIR`), not the producer tree.

### Gate (verbatim, fresh target)
- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` → exit 0
- `cargo fmt --all -- --check` → exit 0
- `cargo test --workspace -- --skip codex_hook` → exit 0, 100 `test result: ok` blocks, 0 FAILED

### Fresh mutation re-runs (independently re-applied in the sentinel tree, each restored after)
| Mutation | Guarding test | Result |
|---|---|---|
| SM1 bridge severed (`inbound_frame_sink()` → `None`) | `inbound_lan_frame_is_visible_in_subscribed_scope_transcript` | FAILED ✔ |
| SM2 dedup disabled (window + sink probe bypassed) | 3 `publish_if_new` pins | FAILED ✔ |
| SM3 ack lie (UnknownChannel verdict acks delivered) | `delivered_ack_means_visible_to_subscribed_scopes_not_just_durable` | FAILED ✔ |
| SM4 unknown-channel diagnostic suppressed | same test, sink assertion | FAILED ✔ |
| SM5 failed publish poisons window (unmark removed) | `failed_publish_does_not_poison_the_window_for_a_retry` | FAILED ✔ |

Post-restore: both test files green; sentinel tree `git status` clean.

### Special audit
**No double-delivery path exists.** Every transport ingest funnels through `append_received_frame`; with a sink installed the scope-store append is fully replaced (no second persistence of the same frame on the daemon handles). Router entry points: `publish` (local IPC, daemon-minted fresh ids, records into the recent-ids window) and `publish_if_new` (bridge, sender-minted ids, window + `DurableSink::contains`). The same frame on listener+dialer handles, a wire echo of a local publish, and a post-restart re-inject are each pinned as `Duplicate`. Read-side `resume_from_cursor` dedup is adjacency-only — confirmed the write-side idempotency is the load-bearing guard (SM2 caught by the bus pins exactly as the PR claims).

**Ack truthfulness.** `delivered` requires router-published (or already-durable duplicate) AND a beacon-bound channel; beacon-set read errors degrade to `unknown_channel` (never false-delivered); router `Failed` → `persist_failed`. Producer's own audit found and fixed the one false-delivered edge (poisoned window on failed publish, commit `0a38924`) — verified by SM5. Residual (documented): the receipt is issued after ring+fan-out with write-behind *enqueued* — identical durability window to local `handle_send`; `WriteBehindSaturated` acks undeliverable for a frame that did fan out live (conservative direction, never false-positive).

**Cursor coherence.** Bridged events take owner-stamped `(epoch, counter)` seqs from the same `SeqSource` as local publishes; duplicates mint no seq. Monotonicity pinned by `cursors_stay_monotonic_across_local_and_bridged_events`. `room_tip`/resume semantics untouched.

### RISK NOTES
1. Beacon check resolves mesh identity per inbound frame via the coordinator-store TTL cache — on cache expiry this can shell `gh` on the ingest path. Bounded and pre-existing (every join does the same); flag if cross-machine chat ever becomes high-rate.
2. Stale (undrained) beacons keep a channel "bound" after a scope is gone → delivered acks for channels nobody currently reads. Deliberate conservative choice (subscription = durable scope state; late reads replay). 
3. Recent-ids window is 4096; beyond it, dedup falls to `DurableSink::contains` (durable-tier safe); a >window-late echo could still double-fan-out LIVE only. Same bound class as the SDK `BroadcastDeduper`.
4. Outbound daemon-routed sends still don't traverse LAN (card 1998f6cb, routed-ack) — this PR closes the inbound visibility leg; the live 5090→Mac path (`lan-send`) is now operator-visible end to end.
5. Running production daemons must be restarted onto this build for the fix to take effect (normal update flow).

### EVIDENCE
- Gate logs: clippy/fmt/test exits captured (`CLIPPY_EXIT=0`, `FMT_EXIT=0`, `TEST_EXIT=0`).
- Hermeticity: all tests use temp homes + in-process daemons + 127.0.0.1 ephemeral ports; RAII teardown (`DaemonFixture::drop` aborts + removes socket); no production home (`~/.airc`, `/Users/joel/Development/*`) touched; live Mac↔5090 route untouched.

Sentinel worktree and temp target removed after this comment.

PR #1157, opened 2026-06-12
fix(transport): daemon advertises BOTH LAN + Tailscale (LAN-first ladder, Tailscale only off-LAN)

joelteply:
## Why

#1154 made the daemon advertise its **Tailscale endpoint exclusively** (falling back to LAN only when Tailscale is down). That forces every *same-subnet* peer to route through a `100.x` Tailscale hop — pure waste on the LAN. Per Joel: **"Tailscale is unnecessary for lan"** and the connection ladder is *local → LAN → Tailscale → greater-grid P2P*, lowest common denominator first.

## What

Advertise **both** addresses so the ladder has rungs:

- New `Airc::listen_lan_advertising(lan_ip, tailscale_ip)` binds **one** wildcard `0.0.0.0:0` listener — a single socket accepts on *every* interface, so the same port is dialable via the host's LAN IP **and** its Tailscale IP — then upserts both specific endpoints. Unlike `listen_lan`, it never advertises the `0.0.0.0` bind address; peers receive only concrete, dialable IPs. (The adapter rejects a second `listen` with `AlreadyListening`, so one wildcard bind is the *only* way to serve both interfaces.)
- The daemon detects both IPs and advertises whichever exist (was: prefer-Tailscale-else-LAN).

**No new dialer logic.** `RouteEndpointKind` already sorts `LanTcp` before `TailscaleTcp`, and the discovery dialer walks endpoints in order and breaks on first success — so a same-subnet peer connects LAN-direct and only a cross-network/firewalled peer falls through to the NAT-traversing `100.x` rung. "Tailscale only if we leave the LAN."

## Tests (`stored_endpoint_dial.rs`)

- `advertise_publishes_both_lan_and_tailscale_lan_first` — one listener, both endpoints share its port, LAN sorts first, `0.0.0.0` never leaks.
- `peer_dials_lan_rung_and_skips_tailscale` — an importing peer connects via the LAN rung with **zero** dial failures; the unreachable Tailscale rung is never touched (LAN-first-break, end to end).

`cargo build`/`clippy`/`fmt --check` clean; full `airc-lib` suite green on native Windows.


joelteply:
**Triage — 2026-06-14**

**Verdict:** NEEDS REBASE

**Why:** Branched 2026-06-12 (post-rust-rewrite) but base diverged by 13 canary commits since — mergeable=CONFLICTING. The substantive change is still missing on canary: `git show canary:crates/airc-lib/src/lan.rs` exposes `listen_lan` (line 19) but no `listen_lan_advertising`, and #1154's Tailscale-exclusive advertise is still the active behavior the PR reverts the ladder-correctness of. Linux + macOS + clippy + fmt all GREEN; only `cargo test (windows-self-hosted)` failed (24h timeout — self-hosted runner appears wedged, not a test defect). Doctrine match: Joel's `[[connection-ladder-local-lan-tailscale]]` is exactly what this PR encodes.

**Next step:** Rebase onto canary HEAD (13 commits — likely trivial merge in Cargo.lock + a router/sink touch), confirm windows-self-hosted runner is alive before push, then merge once a real Windows green lands.

joelteply:
Closed — surgically resurrected as PR #1201 (cherry-pick of the load-bearing commit onto fresh canary, post rust-rewrite squash promote).

PR #1158, opened 2026-06-12
feat(transport): routed-ack — daemon route-layer traversal + delivery acks for ordinary room sends (card 1998f6cb)

joelteply:
## Card 1998f6cb — routed-ack: daemon route-layer traversal for ordinary room sends

closes the final keystone leg: ordinary room messages now traverse LAN routes both directions.

## The path map (where it stopped, where it flows now)

**Before:** `airc send` (daemon-attached scope) → IPC `Request::Send` → `handlers::publish_envelope` → `EventRouter::publish` → hot ring + live fan-out to attached scopes + write-behind to the machine ORM — **and stops**. Nothing handed the event to the route layer, so a healthy lan-tcp route (stored-endpoint dial, #1154/#625abe6d) carried zero ordinary room traffic. `lan-send` worked only because it bypasses the daemon (point-to-point verb on its own handle). Inbound worked post-#1156; outbound was the dead leg.

**Now:** `airc send` → router publish → **bounded outbound tap** (`EventRouter::set_forward_sink`, the OUTBOUND mirror of #1156's `RouterInboundBridge`) → `RoutedForwarder` (airc-lib) → per-peer ordered queue → project the bus envelope back onto the wire `Frame` (exact inverse of `bus_envelope_for_inbound`), re-sign with the forwarding handle's key, `LanTcpAdapter::send_to` over the **existing** connection (never a per-frame dial) → remote daemon's ingest → #1156 inbound bridge → `publish_if_new` → remote scopes' transcripts. Delivery ack (#1155 vocabulary, unchanged on the wire) returns over the same link; the forwarder concludes delivered/undeliverable/no-ack per attempt.

## Design

- **`airc-bus`**: `EventRouter::set_forward_sink(mpsc::Sender<ForwardItem>)`; `ForwardItem { env: Arc<Envelope>, origin: Option<PeerId> }`. Every successfully published **durable** envelope is offered post-acceptance (ring + fan-out + write-behind enqueued), `try_send` only — the route layer can never backpressure the hot path; saturation is counted (`forward_drop_count`) + error-traced, never silent. Ephemeral/stream classes stay machine-local in this slice. `publish_if_new_from(env, origin)` lets the inbound bridge attach the verified link origin so it reaches the tap.
- **`airc-lib::route_forwarder` (new)**: drains the tap; per connected LAN peer across all registered transport-owning handles (daemon listener + dialer), forwards with `airc.delivery_ack=request`. Per-peer bounded ordered queues (no head-of-line blocking across peers; publish order preserved per peer); overflow emits typed `RoutedForwardQueueSaturated`. Outcomes: `delivered` → confirmed counter; `unknown_channel` → typed WARN, no retry (remote holds the frame durably; late joiner replays — proven in #1156); `persist_failed`/no-ack/send-error → retry with the **same frame** (same `event_id` = dedup + ack identity) up to `max_attempts`, then typed ERROR `RoutedForwardFailed`. RAII: drain task aborted when the forwarder drops (workers hold `Weak`, hermetic tests leak nothing).
- **Loop prevention (origin = verified signer)**: the forwarding daemon re-signs the projected frame, so `Signature::Ed25519.signer` IS the link peer at the receiver (TLS pins it; point-to-point sends keep signer == sender, unchanged). The inbound bridge passes that origin into `publish_if_new_from`; the forwarder never sends a frame back over its origin link, and a `Duplicate` outcome never reaches the tap at all. Together: mesh forwarding is a terminating flood — each node forwards at most once (first acceptance), every echo dead-ends as a duplicate. **Transitive forwarding therefore exists and is tested** (A—B—C line topology, exactly-once at all three nodes).
- **Routed acks (extends #1155)**: `append_received_frame` now returns the ack to the **link origin** (`link_origin(frame)` = verified signer, falling back to `sender` for unsigned/dev frames) instead of `envelope.sender` — for point-to-point sends these are identical (old behavior preserved, e2e suite green), for routed forwards the sender is a remote scope identity with no connection on the receiving box and the ack would have been a silent `send_to` failure.
- **The ring-marked-but-unpersisted retry edge (card body)**: remote reports failure after (or instead of) persisting → forwarder retries the same `event_id` → remote's `publish_if_new` either really publishes (its recent-ids entry was unmarked by #1156's poisoning fix 0a38924 — no false `Duplicate` for a frame that never landed) or truthfully dedups an actually-persisted copy into `delivered` (duplicate IS delivered). Exactly one visible copy either way; pinned by test 4.
- **Channel gating**: the forwarder sends to every connected (healthy, established) LAN peer rather than gating on the remote's beacon channel set — remote channel bindings are not durably available peer-side today (trust records carry endpoints, not channels). The remote's #1156 bridge is the truthful arbiter: unbound channel ⇒ durable + `undeliverable{unknown_channel}` + loud diagnostics on both ends. Sender-side beacon gating is a bandwidth optimization left as a follow-up, stated here explicitly.

## Wiring (`run_daemon`)

`RoutedForwarder::install(&state.router, …)` before any transport handle exists; the registry/listener handle and the route-refresh/dialer handle are registered as links when they open (`add_link`), exactly where their #1156 inbound bridges are installed. Forwards reuse whichever handle currently holds the connection and are signed with that handle's identity (the identity the remote pinned at TLS time, so verification passes by construction).

## Tests (hermetic: in-process daemons, real TLS, temp homes, RAII teardown — no production daemon/route touched)

`crates/airc-lib/tests/daemon_routed_forward.rs` (extends the #1156 harness):
1. **THE acceptance test** — `routed_room_send_traverses_lan_both_directions`: an ordinary room send on daemon A appears exactly once in daemon B's subscribed scope transcript, AND the reverse direction over the same single TLS connection; both ack-confirmed.
2. **Loop prevention** — `frame_received_from_a_peer_is_not_forwarded_back_to_it`: B's wire-flush counter does not move after receiving A's frame (transcript counts alone cannot catch the echo — A dedups it).
3. **Three-peer transitive** — `transitive_forward_reaches_third_machine_exactly_once`: A—B—C line; exactly once at A, B, and C after the flood settles.
4. **Retry edge** — `remote_persist_failure_then_retry_is_exactly_once_visible`: remote persists but reports `persist_failed` (ack-uncertainty window) → retry with stable identity → ack-confirmed delivered, exactly one visible copy.
5. **Loud backpressure** — `saturated_forward_queue_and_unconfirmed_forwards_are_loud`: hollow peer (accepts TLS, never acks) + 1-deep peer queue + burst ⇒ typed `RoutedForwardQueueSaturated` AND `RoutedForwardFailed` (sink-asserted), local delivery untouched, `confirmed_count == 0` (no false confirmation, ever).

`crates/airc-bus/tests/forward_sink.rs`: tap receives durables with correct origin (None local / Some(link) bridged); `Duplicate` and ephemeral never reach the tap; tap saturation is a counted loud-drop that never fails/blocks `publish`.

## Mutation evidence (each applied, observed failing, restored; suite green after)

1. **M1 — forwarding severed** (`offer_to_forward_sink` call removed): acceptance test **FAILED** (B never sees the event) + both bus tap tests **FAILED**.
2. **M2 — origin check removed** (`if Some(peer) == item.origin { continue; }` deleted): loop test **FAILED** (B's forwarded counter moved — echo flushed back to A).
3. **M3 — retry identity broken** (`event_id: EventId::new()` per projected frame): exactly-once retry test **FAILED** (two transcript copies at B / no matching ack), acceptance **FAILED** too (ack matching by event id).
4. **M4 — saturation diagnostic suppressed** (emit removed in the queue-Full arm): loudness test **FAILED** on the sink assertion.

## Production gate

- `cargo test --workspace -- --skip codex_hook` — green

Builds on #1155 (delivery-ack vocabulary) and #1156 (RouterInboundBridge + `publish_if_new` poisoning fix). Card: 1998f6cb-5732-4468-9e88-465422d7d58e.

### DIFF
--- a/crates/airc-bus/src/lib.rs
+++ b/crates/airc-bus/src/lib.rs
@@ -51,6 +51,6 @@ pub use ephemeral::EphemeralCache;
 pub use error::{BusError, Result};
 pub use filter::Filter;
 pub use ring::HotRing;
-pub use router::{EventRouter, LagFlag, PublishIfNew, RouterConfig};
+pub use router::{EventRouter, ForwardItem, LagFlag, PublishIfNew, RouterConfig};
 pub use seq::{EpochStore, InMemoryEpochStore, SeqSource};
 pub use sink::{DurableSink, InMemoryDurableSink};
--- a/crates/airc-bus/src/router.rs
+++ b/crates/airc-bus/src/router.rs
@@ -35,7 +35,7 @@ use std::sync::{Arc, Mutex};
 use futures::stream::Stream;
 use tokio::sync::mpsc;
 
-use airc_core::RoomId;
+use airc_core::{PeerId, RoomId};
 
 use crate::clock::Clock;
 use crate::envelope::{Cursor, Envelope};
@@ -164,6 +164,24 @@ impl RecentEventIds {
     }
 }
 
+/// Card 1998f6cb: one durable envelope the router accepted, handed to
+/// the installed outbound forward sink (route layer) so it can
+/// traverse established LAN routes. The OUTBOUND mirror of card
+/// 4132f48c's inbound bridge: inbound goes transport → router via
+/// `publish_if_new`; outbound goes router → transport via this item.
+#[derive(Debug, Clone)]
+pub struct ForwardItem {
+    /// The published envelope, post owner-stamping (`seq` /
+    /// `occurred_at_ms` assigned). Shared, never deep-copied.
+    pub env: Arc<Envelope>,
+    /// The LAN-link peer this envelope ARRIVED from, when it entered
+    /// the router through the inbound bridge (`publish_if_new_from`).
+    /// `None` = locally originated (IPC publish). Loop prevention: the
+    /// forwarder must never send a frame back over the link it came
+    /// from.
+    pub origin: Option<PeerId>,
+}
+
 /// Outcome of [`EventRouter::publish_if_new`].
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum PublishIfNew {
@@ -196,6 +214,17 @@ struct RouterInner {
     /// event (or the same inbound frame arriving on two LAN links) can
     /// never fan out twice.
     recent_ids: Mutex<RecentEventIds>,
+    /// Card 1998f6cb: the outbound route-layer sink. When installed
+    /// (`set_forward_sink`), every successfully published **durable**
+    /// envelope is offered here (bounded `try_send`, never blocking the
+    /// hot path) so the daemon's forwarder can send it over established
+    /// LAN routes. Saturation is LOUD: counted + traced, never silent.
+    forward_tx: Mutex<Option<mpsc::Sender<ForwardItem>>>,
+    /// Count of durable envelopes NOT handed to the forward sink because
+    /// its bounded queue was full (or the forwarder task was gone).
+    /// Surfaced for diagnostics/tests; every increment also traces at
+    /// error level.
+    forward_drop_count: AtomicU64,
     /// Count of events shed because the write-behind queue was saturated and
     /// the publisher was fire-and-forget (§3.8). Surfaced for diagnostics.
     shed_count: AtomicU64,
@@ -233,6 +262,8 @@ impl EventRouter {
             sink,
             write_behind_tx,
             recent_ids: Mutex::new(RecentEventIds::with_capacity(RECENT_PUBLISH_IDS_CAPACITY)),
+            forward_tx: Mutex::new(None),
+            forward_drop_count: AtomicU64::new(0),
             shed_count: AtomicU64::new(0),
             channels_created: AtomicU64::new(0),
         });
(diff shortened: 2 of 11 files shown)


joelteply:
## SENTINEL — card 1998f6cb (producer-runs-sentinel-inline, default-BLOCK)

Fresh temp worktree `/tmp/sentinel-1998f6cb` @ `b6c52a8` (PR head), separate target dir `/tmp/sentinel-1998f6cb-target`. Stance: BLOCK unless every gate + fresh mutation + special audit passes.

### VERDICT: APPROVE

### Gate (verbatim, fresh worktree)

- `cargo fmt --all -- --check` — **clean**
- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` — **exit 0**
- `cargo test --workspace -- --skip codex_hook` — **exit 0** (102 suites, 0 failures). One earlier sentinel run showed 8 `gh_account_registry` unit failures — root-caused to the sentinel shell itself exporting `AIRC_DISABLE_ACCOUNT_REGISTRY=1` globally (those unit tests own that variable); rerun without the contamination is green. The PR's own integration tests never spawn the registry loop (in-process `DaemonState`, not `run_daemon`), so they are hermetic WITHOUT the env var — no gh-rendezvous pollution possible by construction.

### FRESH mutation re-runs (applied in the sentinel worktree, observed failing, reverted)

1. **M1 forwarding severed** (`offer_to_forward_sink` call removed): `routed_room_send_traverses_lan_both_directions` FAILED + all 3 `airc-bus forward_sink` pins FAILED.
2. **M2 origin check removed** (drain-loop `Some(peer) == item.origin` skip deleted): `frame_received_from_a_peer_is_not_forwarded_back_to_it` FAILED (B's wire-flush counter moved — the echo was actually flushed back to A; transcript counts alone are insensitive because A dedups, which is exactly why the counter observable exists).
3. **M3 retry identity broken** (`event_id: EventId::new()` per projected frame): `remote_persist_failure_then_retry_is_exactly_once_visible` FAILED (ack never matches the envelope identity; remote would double-publish).
4. **M4 saturation diagnostic suppressed** (queue-Full emit removed): `saturated_forward_queue_and_unconfirmed_forwards_are_loud` FAILED on the sink assertion.

### Special audit 1 — loop prevention airtight

- **Two-peer echo**: pinned by the dedicated test (counter observable, mutation-verified M2). Mechanism is two independent rules: (a) never forward to the origin link; (b) `Duplicate` outcomes never reach the tap at all (`publish_if_new_from` short-circuits before `publish_from`; bus pin `duplicates_and_ephemerals_never_reach_the_forward_sink`).
- **Three-peer transitive**: mesh forwarding EXISTS in this design (a `Published` bridged frame is re-offered to all non-origin links) and is tested: A—B—C line topology delivers exactly once at all three nodes (`transitive_forward_reaches_third_machine_exactly_once`), including a settle window re-assertion. Termination argument verified in code: each node forwards a given event id at most once (only on first acceptance); any echo path ends as `Duplicate` (recent-ids window, durable-tier `contains` fallback for ids older than the 4096-entry window) — no amplification, no storm.
- Origin identity is the **verified Ed25519 signer** (TLS-pinned: the connection itself is keyed by the cert the signer's key produced; the forwarder re-signs with the link handle identity). Risk note R1 below for the dev-only unsigned policy.

### Special audit 2 — no per-frame re-dial

`route_forwarder.rs` contains no `connect`/`listen`/dial path: targets come from `connected_peers()` snapshots of EXISTING `LanTcpAdapter`s; the only wire op is `send_to` (unicast on the installed connection). No-connection is treated as a retryable condition left to route discovery (#625abe6d) to repair — verified by grep + read.

### Special audit 3 — ack truthfulness end-to-end

- `confirmed` increments ONLY on a `delivered` ack, which post-#1156 means: published into the remote router AND a scope on the remote machine binds the channel. NoAck and `persist_failed` never confirm (retry → typed `RoutedForwardFailed` ERROR after exhaustion); `unknown_channel` never confirms (typed WARN; remote holds frame durably, late joiner replays — #1156-proven).
- Hollow-receiver test pins `confirmed_count == 0` under total ack silence — no false confirmation channel exists.
- Ack return path fix audited: acks now travel to the verified **link origin** (signer) instead of `envelope.sender`; for point-to-point sends signer == sender so #1155 e2e behavior is bit-identical (its suite passes unmodified in this worktree).
- Retry edge: remote `publish_if_new` unmark-on-failure (#1156 fix 0a38924) + stable event identity on retry (M3-verified) closes both halves — no false `Duplicate→Delivered` for a never-landed frame, no double-copy for a landed-but-unconfirmed frame.

### RISK NOTES (none blocking)

- **R1 (dev-only)**: under `VerificationPolicy::AllowUnsigned`, `link_origin` falls back to `envelope.sender`, which may differ from the actual link peer — at most ONE wasted echo per frame, terminated as `Duplicate` at the origin. Production policy is Strict (always signed); not reachable in deployed daemons.
- **R2 (trust boundary)**: the ack's `receiver` field is self-declared inside a signed frame; a *trusted* but buggy peer could misattribute receipts. Same boundary as #1155 — acks are only ever accepted from TLS-verified, enrolled peers.
- **R3 (header cosmetics)**: forwarded frames persist `airc.delivery_ack=request` in remote transcript headers (precedent: #1155 point-to-point does the same). Cosmetic.
- **R4 (per-peer head-of-line)**: a sick-but-connected peer stalls only its OWN ordered queue (worst ~31s/frame at defaults: 3 × (10s ack + 0.5s backoff)); overflow is the typed loud drop. Healthy peers are unaffected (per-peer workers).
- **R5 (no sender-side channel gating)**: the forwarder sends to every connected peer rather than gating on remote beacon channel sets (not durably available peer-side today); the remote bridge arbitrates truthfully and loudly. Stated in the PR as a follow-up bandwidth optimization — acceptable, the failure mode is noise, not loss.
- **R6 (scope)**: ephemeral/stream delivery classes and RPC kinds (Command/CommandResult/Signal/StreamChunk) stay machine-local by design; durable Message/Event/Control traverse. Stated in code + PR.

### EVIDENCE

- Sentinel gate logs: fmt/clippy exit 0; full suite exit 0 (rerun after env-contamination fix), `/tmp/sentinel-1998f6cb-tests.log` on the build host.
- Producer worktree gate: fmt clean, clippy clean, full workspace suite EXIT=0 (102 suites, 0 failures).
- CI on head `b6c52a8`: clippy/fmt/quality-guard/macos/ubuntu all PASS (windows-self-hosted pending at verdict time). The `914d613` hosted failures were a real test race — asserting the async ack-confirmation counter immediately after observing remote visibility; fixed by polling, which is the semantically correct wait for an asynchronous receipt.

PR #1159, opened 2026-06-13
fix(test-harness): stop temp-home daemon leaks — RAII teardown + self-exit + home-derived sockets (card f122b5b5)

joelteply:
## Card f122b5b5 — test harness leaks daemons

`cargo test --workspace` left airc daemon processes alive with `--home` under the OS temp dir. Counts measured this session: ~40, 131, 405+86, 154 per run — **800+ killed by hand**. Each leaked daemon held RAM + an event loop on the underpowered coordination Mac, and one was observed planting its socket under the PRODUCTION `~/.airc/runtime` (stale `airc-10e8167b5d5b936d-v5.sock`).

### Spawn-site inventory (every distinct path)
- **IMPLICIT (the bulk leak):** almost every `airc <verb>` an integration test runs against a tempdir home calls `ensure_daemon_running`, which spawn-or-connects a **DETACHED** daemon (`setsid` / `DETACHED_PROCESS`). It outlives the test by design (it's a daemon) and **no guard tracked it**. This is the leak the counts came from.
- **EXPLICIT subprocess:** `tests/daemon_route_refresh.rs`, `tests/registry_hermetic.rs` — each already had a kill-on-Drop `DaemonGuard` for the daemon it spawns directly.
- **IN-PROCESS:** `airc-lib tests/common` `DaemonFixture` — already aborts its `JoinHandle` on Drop (no subprocess).

### The four layers

**1. RAII teardown (`crates/airc-cli/tests/common/mod.rs`).** Tests now hold a `DaemonTempDir` instead of a raw `TempDir`; the helper `daemon_tempdir()` ONLY returns the guarded form, so teardown cannot be forgotten. Its `Drop` walks the tree for the `daemon.pid` each daemon now writes (`PidFileGuard`, `airc-daemon`) and reaps them `SIGTERM → poll-gone → SIGKILL`. Every CLI test file migrated to it.

**2. Belt-and-braces self-exit (`crates/airc-daemon/src/server.rs`).** A daemon whose home is temp-rooted exits BY ITSELF after an idle window with no connected client (loud one-line log naming the policy). This catches kill-escapes — a SIGKILLed test runner whose Drop guards never ran. Production homes never arm it. The temp detection (#1150) was hoisted to `airc-core::temp_home::scope_home_is_temp_rooted` so `airc-daemon` can reuse it without depending on `airc-lib` (`airc-lib` re-exports it; no behaviour change). Window overridable via `AIRC_TEMP_HOME_IDLE_EXIT_MS` (tests use ~800ms; default 5 min).

**3. Socket-path bug (`crates/airc-cli/src/cli.rs`, `runtime_dir.rs`).** An explicit temp home no longer places its socket under the user's `~/.airc/runtime`. The isolated-scope branch of `resolve_socket_path` derived the socket NAME from a project-root hash but placed the FILE in `runtime_dir` (= `~/.airc/runtime`) — exactly how the stale production socket appeared. The socket now derives from the daemon's own home; same-home processes still rendezvous, parallel isolated scopes still can't collide. `project_socket_path` (the runtime-dir placement) is removed.

**4. CI zero-leak guard (`.github/workflows/rust.yml`).** A `ps`-based step after `cargo test` (ubuntu + macos; skipped on Windows) FAILS the leg if any `airc` daemon with a temp-rooted `--home` survives the suite — the card's acceptance, enforced. `crates/airc-cli/tests/daemon_teardown.rs` additionally pins all four layers in-suite.

### Mutation evidence (fresh re-runs on this branch)
| Mutation | Test | Result |
|---|---|---|
| Disable `DaemonTempDir::drop` reap | `leaked_daemon_is_reaped_by_guard_drop` | **FAILED** (101) |
| `spawn_temp_home_idle_watchdog` → `None` | `temp_home_daemon_self_exits_when_idle` | **FAILED** (101, times out) |
| Revert socket derivation to runtime-dir | `explicit_temp_home_never_places_socket_under_user_runtime_dir` | **FAILED** (101) |

Each reverted cleanly (working tree matches the commit) and all four `daemon_teardown` tests pass unmutated.

### Verification
- Full `cargo test --workspace -- --skip codex_hook`: **1203 passed, 0 failed**.
- Acceptance: after the full suite, **zero** temp-home airc daemons remained (the one kill-escaped straggler from an aborted run was the exact Layer-2 case and was reaped).
- Production gate green: `cargo fmt --all --check`; clippy production no-silent-fallback gate (`-D unwrap_used/expect_used/panic/todo/unimplemented/wildcard_enum_match_arm`); clippy `--all-targets -D warnings`; rust quality guard.

card f122b5b5

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -200,3 +200,28 @@ jobs:
             ${{ runner.os }}-cargo-test-
       - name: cargo test --workspace --all-features
         run: cargo test --workspace --all-features
+      # Card f122b5b5 ZERO-LEAK GUARD: the acceptance is that a full
+      # `cargo test --workspace` leaves ZERO airc daemon processes whose
+      # `--home` is under a temp dir. Test daemons get RAII teardown + a
+      # temp-home idle self-exit watchdog; this step is the backstop that
+      # FAILS the leg if either defense ever regresses. ps-based and
+      # portable (ubuntu + macos); skipped on Windows where `ps`/`pgrep`
+      # differ and the self-hosted runner makes process scans noisy.
+      - name: zero leaked temp-home daemons
+        if: runner.os != 'Windows'
+        shell: bash
+        run: |
+          # Match an `airc ... daemon` process whose `--home` argument
+          # points under a temp root (/tmp, /var/folders, /private/...,
+          # $TMPDIR). Production daemons (home under $HOME/.airc or a
+          # repo `.airc`) never match.
+          leaked="$(ps -Ao pid=,args= \
+            | grep -E '(^|/)airc .* daemon' \
+            | grep -E -- '--home[ =](/private)?(/tmp|/var/folders|/var/tmp)/' \
+            || true)"
+          if [ -n "$leaked" ]; then
+            echo "::error::card f122b5b5 — temp-home airc daemons survived the test suite:"
+            echo "$leaked"
+            exit 1
+          fi
+          echo "zero leaked temp-home airc daemons after the suite (card f122b5b5)"
--- a/crates/airc-cli/Cargo.toml
+++ b/crates/airc-cli/Cargo.toml
@@ -60,3 +60,9 @@ libc = "0.2"
 [dev-dependencies]
 tempfile = "3"
 temp-env = "0.3"
+
+# Card f122b5b5: the daemon-teardown guard (tests/common) needs
+# kill(2)/ESRCH probing to reap daemons the CLI auto-spawned for
+# temp homes.
+[target.'cfg(unix)'.dev-dependencies]
+libc = "0.2"
--- a/crates/airc-cli/src/cli.rs
+++ b/crates/airc-cli/src/cli.rs
@@ -206,17 +206,12 @@ pub fn default_socket_path_in(home: &std::path::Path) -> PathBuf {
     // SUN_LEN fallback root for deep-home cases (see machine_socket_path).
     // The OS per-user temp dir is short; never used for normal homes.
     let short_fallback = std::env::temp_dir();
-    let project_socket = home
-        .parent()
-        .map(|root| crate::runtime_dir::project_socket_path(root).ok())
-        .unwrap_or(None);
     let socket = resolve_socket_path(
         home,
         &machine_home,
         user_home.as_deref(),
         runtime_dir.as_deref(),
         Some(short_fallback.as_path()),
-        project_socket,
     );
     // runtime_dir() created ~/.airc/runtime, but the SUN_LEN fallback dir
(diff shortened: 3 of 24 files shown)


joelteply:
## Sentinel verdict: APPROVE

Producer-run inline sentinel in a clean worktree (`/tmp/sentinel-f122b5b5`) on this branch's HEAD (`c8d4eed`), separate target dir (`/tmp/sentinel-f122b5b5-target`), roomy TMPDIR. Default-BLOCK; cleared because every gate leg passed and the special zero-leak audit bit.

### Gate (verbatim)
- `cargo fmt --all -- --check` → **0**
- clippy production no-silent-fallback gate (`-D warnings -D unwrap_used -D expect_used -D panic -D todo -D unimplemented -D wildcard_enum_match_arm`) → **0**
- `cargo clippy --workspace --all-targets --all-features -D warnings` → **0**
- `cargo test --workspace -- --skip codex_hook` → **0** — **1203 passed, 0 failed**

### Special audit — the zero-leak guard actually bites
- **After the suite:** `GUARD_AFTER_SUITE=PASS` — zero temp-home airc daemons survived the full workspace run in the sentinel worktree.
- **Injected-leak bite:** deliberately spawned a temp-home daemon, then ran the guard's detector → `GUARD_BITES_INJECTED_LEAK=PASS` (it saw the leak). Injected daemon reaped afterward; sentinel left nothing behind.

So the guard fails loud when a leak exists and passes only when there is none — it is not a no-op.

### Fresh mutation re-runs (producer side, reverted clean)
| Mutation | Test | Result |
|---|---|---|
| Disable `DaemonTempDir::drop` reap | `leaked_daemon_is_reaped_by_guard_drop` | FAILED (101) |
| `spawn_temp_home_idle_watchdog` → `None` | `temp_home_daemon_self_exits_when_idle` | FAILED (101, times out) |
| Revert socket derivation to runtime-dir | `explicit_temp_home_never_places_socket_under_user_runtime_dir` | FAILED (101) |

Each mutation reverted with a clean working tree (zero `git diff` vs the commit); all four `daemon_teardown` tests pass unmutated.

### Risk notes
- The self-exit default window is 5 min; healthy daemon-class tests use bounded waits well under that, so no production or test daemon trips it early. The window is test-overridable via `AIRC_TEMP_HOME_IDLE_EXIT_MS`.
- The CI guard is `ps`-based and Unix-only (ubuntu + macos); Windows is intentionally skipped (`ps`/process-scan noise on the self-hosted runner) — acceptable, the RAII + self-exit layers still run there.
- Temp detection keys off `std::env::temp_dir()` (plus cross-platform markers), so a deliberately non-temp `TMPDIR` would not classify homes as temp — not a concern for real CI/test environments.

Cleanup done: sentinel worktree + target + injected daemon removed.

PR #1147, opened 2026-06-12
fix(gh): 401 token refresh + pr_create in ReqwestGhClient (card 09cd0afb Sub-3)

joelteply:
## What / Why

Residuals from #1083 (which made `ReqwestGhClient` the default gh backend). Card `09cd0afb-235a-4168-aeac-7f14d309f289`.

**1. Token refresh on HTTP 401** — the resolved token was cached in a `OnceLock` for process lifetime; the module comment claimed "refresh on HTTP 401" but no refresh path existed (flagged in the #1083 sentinel review). gh API tokens rotate mid-session (documented operational fact), so a long-lived merger daemon failed every tick with `AuthRequired` until restart.

- `OnceLock` → `Arc<RwLock<Option<String>>>` (Send+Sync, poison-tolerant; happy path is one uncontended read-lock + clone per request, same as before).
- New `send_authed` centralizes request dispatch: on a 401, the resolution chain (`GH_TOKEN` → `GITHUB_TOKEN` → `gh auth token` re-spawn, cache bypassed) re-runs EXACTLY ONCE and the request retries with the fresh token. The keychain copy `gh auth token` reads usually outlives a rotated env copy, so the re-spawn is what rescues the daemon.
- A second 401 fails LOUDLY: `GhError::AuthRequired` naming the chain and the fix (`gh auth login` / rotate env var). One refresh attempt per request — no retry storms, no silent fallback to the shell backend, token values never appear in any error/Debug output.

**2. `ReqwestGhClient::pr_create`** — was a stub returning `Err` (landmine for anything wiring pr_create through the trait). Now `POST /repos/{owner}/{repo}/pulls` with typed serde structs (`PrCreateRequest` / `PrCreateResponse`, single source of truth for wire field names), mirroring `ShellGhClient` semantics: repo resolved from the worktree's origin remote, head from the current branch, title/body from the HEAD commit (the deterministic resolution the codebase already standardized on over `--fill`'s branch-name heuristic), `draft: false`, returns `PrCreated { url, number }`.

## Evidence

**Production gate** (verbatim, at HEAD):
- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` → clean
- `cargo fmt --all -- --check` → clean
- `cargo test --workspace -- --skip codex_hook` → 95 suites, 0 failures

[...]

PR #1162, opened 2026-06-13
fix(gh-auth): gate hardening — stub-gh tests, named timeout, AIRC_GH_BIN end-to-end, stale-token re-resolve (card 1f2cbffa)

joelteply:
Card **1f2cbffa-7887-4378-9213-77c03b3d0392** — gh-auth gate hardening: the four post-merge-audit residuals from #1145 ([audit comment](https://github.com/CambrianTech/airc/pull/1145#issuecomment-4686386364)).

## 1. Stub-gh tests upstreamed (audit risk note 1)

The audit's throwaway evidence is now a permanent `cfg(unix)` suite (`gh_account_registry::tests::gh_auth_gate_stub`; hosted ubuntu+macos run it, precedent #1150):

- `gate_allows_slow_but_authed_gh_900ms_class` — a stub gh that sleeps ~1s then exits 0 must pass the gate. This is the measured 881–950ms Windows `gh auth status` class that the pre-#1145 750ms budget timed out on every tick.
- `gate_fails_fast_for_unauthed_gh` — exit-1 stub fails the gate in well under 2s (the budget exists only for the slow-but-authed path).
- `gate_kills_hung_gh_at_the_deadline` — an `exec sleep 30` stub is killed at the deadline; the gate answers in ~`GH_AUTH_READY_TIMEOUT`, never 30s.

## 2. Named timeout constant (audit risk note 1)

`GH_AUTH_READY_TIMEOUT: Duration = Duration::from_secs(5)` (pub, `gh_account_registry.rs`) replaces the inline magic `Duration::from_secs(5)`; its doc comment carries the bigmama-measured rationale that previously lived as an inline comment. The probe, the kill-deadline, and the new `gh auth token` re-resolve all share it.

## 3. AIRC_GH_BIN honored end-to-end (audit risk note 3)

- `resolve_gh_bin` now returns the operator override **unconditionally** when set — with a loud stderr line if nothing exists at that path, and **no silent fallback to PATH** (an override that quietly stops applying is the worst failure shape; a broken one fails loudly per tick, which is diagnosable). `run_daemon`'s `.with_bin(...)` therefore carries the override instead of clobbering the store's env-honoring default.
- `resolve_gh_token` consults only the override when set (token extraction from the wrong gh is the same bug).
- The manual `registry sync` gate (`registry_commands::run_sync`) now probes the overridden binary instead of hardcoded bare `gh`.
- Pinned via env-free `_with()` bodies (`resolve_gh_bin_with` / `resolve_gh_token_with`) so the tests need no racy process-env mutation: `resolve_gh_bin_honors_airc_gh_bin_override`, `resolve_gh_bin_broken_override_never_falls_back_to_path`, `resolve_gh_token_uses_the_override_binary_only`.

## 4. Stale-token re-resolve for the daemon's gh transport (audit risk note 2)

`GH_TOKEN` is snapshotted into the daemon env at spawn (`inject_gh_token`) and tokens rotate mid-session (documented operational fact), so a long-lived daemon's registry ticks failed until restart. Mirroring `ReqwestGhClient`'s #1147 401-refresh:

- On a failed auth probe, the gate makes **one recovery attempt per tick**: `re_resolve_gh_token` spawns `gh auth token` with `GH_TOKEN`/`GITHUB_TOKEN` **stripped from the child env** — gh prefers an env token over its keyring copy, so without the strip it would just echo the stale injected token back; stripping is what makes the keyring copy (which usually outlives a rotated env snapshot) reachable at all.
- A recovered token lands in a shared `GhTokenOverride` slot (Debug redacts it) that `run_daemon` hands to **both** the gate (re-probes with the fresh token) and the `GhAccountRegistryStore` (every gh spawn carries it via `cmd.env("GH_TOKEN", …)`), so recovery doesn't just pass the gate and then fail the publish.
- If the daemon lacks keychain access (the reason injection exists in the first place), the attempt returns `None` and the **existing** loud `GhNotReady` skip diagnostic fires — one extra recovery attempt, no new failure modes. Recovery success logs one loud line.

**Said-so (per the card's instruction to state limits rather than invent fallbacks):**
- The "re-read GH_TOKEN env" leg of #1147's chain is deliberately absent: gh child processes already inherit the daemon's env token, so re-reading our own immutable process env can never change the probe outcome — that frozen snapshot *is* the bug being recovered from. The only recoverable source from inside the daemon is the keyring via the env-stripped `gh auth token` spawn.
- If the daemon has neither keyring access nor a still-valid env token, recovery cannot succeed — behavior is then identical to before this PR (loud per-tick skip until restart). That is the honest limit of daemon-side recovery; restart remains the fix for that case.
- Manual `registry sync` gets `token_override: None` (no recovery): it runs in the operator's live session where env/keyring are already current, and without a daemon store to hand the token to, passing the gate with a recovered token would just move the same failure into the publish.

## Mutation evidence (each pin run failed, then restored green)

| Mutation | Result |
|---|---|
| `GH_AUTH_READY_TIMEOUT` 5s → 750ms | `gate_allows_slow_but_authed_gh_900ms_class` **FAILED in 0.76s** (exact audit repro; `re_resolve_strips_stale_env_and_reads_keyring_copy` also failed) |
| `resolve_gh_bin_with` ignores the override | both AIRC_GH_BIN bin pins **FAILED** |
| `resolve_gh_token_with` falls back to default candidates despite override | `resolve_gh_token_uses_the_override_binary_only` **FAILED** |
| re-resolve arm removed from `RegistryRefreshGate::block` | `gate_recovers_from_stale_daemon_token_with_one_re_resolve` **FAILED** |
| store's `cmd.env("GH_TOKEN", …)` application removed from `gh_run` | `store_gh_spawns_carry_the_recovered_token` **FAILED** |

## Gate

- `cargo fmt --all -- --check` — PASS
- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` — PASS
- `cargo test --workspace -- --skip codex_hook` — PASS (see thread)
- No daemons spawned by the new tests (stub scripts only); #1159 RAII/zero-leak posture untouched.

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -525,11 +525,21 @@ fn inject_gh_token(command: &mut Command) {
     command.env("GH_TOKEN", token);
 }
 
-/// `gh auth token` from the parent, robust to PATH-resolution quirks in
-/// a bash-descended process (where bare `gh` may not resolve the same as
-/// in an interactive shell). Tries `gh` on PATH first, then known
-/// install locations.
-fn resolve_gh_token() -> Option<String> {
+/// The operator's `AIRC_GH_BIN` override, if set non-empty (card
+/// 1f2cbffa, #1145 audit item 3). When present it is AUTHORITATIVE for
+/// every gh resolution in this process — token extraction, the
+/// daemon's gate + store, the manual `registry sync` gate.
+pub(crate) fn gh_bin_override() -> Option<std::path::PathBuf> {
+    let raw = std::env::var_os("AIRC_GH_BIN")?;
+    if raw.is_empty() {
+        return None;
+    }
+    Some(std::path::PathBuf::from(raw))
+}
+
+/// Known gh install locations, tried after bare `gh` on PATH. Only
+/// consulted when the operator has NOT set `AIRC_GH_BIN`.
+fn default_gh_candidates() -> Vec<std::path::PathBuf> {
     let mut candidates: Vec<std::path::PathBuf> = vec![std::path::PathBuf::from("gh")];
     #[cfg(windows)]
     {
@@ -546,6 +556,27 @@ fn resolve_gh_token() -> Option<String> {
         candidates.push(std::path::PathBuf::from("/usr/local/bin/gh"));
         candidates.push(std::path::PathBuf::from("/usr/bin/gh"));
     }
+    candidates
+}
+
+/// `gh auth token` from the parent, robust to PATH-resolution quirks in
+/// a bash-descended process (where bare `gh` may not resolve the same as
+/// in an interactive shell). Honors `AIRC_GH_BIN` (the override is the
+/// ONLY candidate — a broken override must fail loudly upstream, never
+/// silently swap to a different gh); otherwise tries `gh` on PATH, then
+/// known install locations.
+fn resolve_gh_token() -> Option<String> {
+    resolve_gh_token_with(gh_bin_override())
+}
+
+/// Env-free body of [`resolve_gh_token`] — `override_bin` is the
+/// operator's `AIRC_GH_BIN`, parameterized so tests can pin the
+/// override contract without racy process-env mutation.
+fn resolve_gh_token_with(override_bin: Option<std::path::PathBuf>) -> Option<String> {
+    let candidates = match override_bin {
+        Some(bin) => vec![bin],
+        None => default_gh_candidates(),
+    };
     for bin in candidates {
         let Ok(output) = std::process::Command::new(&bin)
             .args(["auth", "token"])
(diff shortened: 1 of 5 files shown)


joelteply:
## SENTINEL VERDICT: APPROVE (producer-runs-sentinel-inline, card 1f2cbffa)

Run in a fresh temp worktree (`/tmp/sentinel-1f2cbffa`, detached at PR head 663c26c) with a separate target dir. Default-BLOCK posture; nothing below blocks.

### EVIDENCE (all at PR head 663c26c)
- `cargo fmt --all -- --check` — PASS
- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` — PASS
- `cargo test --workspace -- --skip codex_hook` — PASS: **103 suites ok, 0 failed, exit 0**
- Hosted CI: ubuntu-latest **pass**, macos-latest **pass** (both run the new cfg(unix) stub tests). windows-self-hosted **queued** — infra exception vs merge-base holds: the same leg is cancelled/queued on rust-rewrite HEAD commits ae6f2ea and 1678764 (runner backlog), so its state on this PR matches the merge-base, not this diff. Notably macos-latest FAILED on both merge-base commits and passes here.

### FRESH MUTATION RE-RUNS (each mutated in the sentinel worktree, run, restored; worktree verified clean + green after)
| Mutation | Expected pin | Result |
|---|---|---|
| `GH_AUTH_READY_TIMEOUT` 5s → 750ms | `gate_allows_slow_but_authed_gh_900ms_class` | **FAILED in 0.76s** (exit 101) — byte-for-byte the #1145 audit repro |
| `resolve_gh_bin_with` ignores override | both AIRC_GH_BIN bin pins | **FAILED** (exit 101) |
| `resolve_gh_token_with` falls back past override | `resolve_gh_token_uses_the_override_binary_only` | **FAILED** (exit 101) |
| recovery arm disabled in `RegistryRefreshGate::block` | `gate_recovers_from_stale_daemon_token_with_one_re_resolve` | **FAILED** (exit 101) |
| store `cmd.env("GH_TOKEN", …)` application removed | `store_gh_spawns_carry_the_recovered_token` | **FAILED** (exit 101) |

### CORRECTNESS READ
- Item 4's design is honest about its limits: recovery only works when the daemon can reach a keyring copy via an env-stripped `gh auth token`; otherwise behavior is byte-identical to pre-PR (loud per-tick `GhNotReady` skip). The env-strip is the load-bearing detail — without it gh echoes the stale injected token back — and it is pinned by `re_resolve_strips_stale_env_and_reads_keyring_copy`.
- The shared `GhTokenOverride` slot closes the gate-passes-but-publish-fails gap (store spawns carry the recovered token); Debug is hand-written to redact the token (RegistryRefreshGate derives Debug).
- `token_override: None` for manual `registry sync` is correct: no daemon store to hand a recovered token to, and the operator session's env/keyring are current.
- AIRC_GH_BIN: returning a missing override loudly (instead of PATH fallback) is the right failure shape — per-tick loud failures are diagnosable, a silently-wrong gh is not.

### RISK NOTES (none blocking)
1. **Pre-existing temp-home daemon leak, NOT this PR**: one `--home …/T/.tmpXXXX/codex` daemon survives each full workspace test run (observed identically on unmodified rust-rewrite HEAD during producer runs and in the sentinel run; killed manually each time). The new tests in this PR spawn only `/bin/sh` stubs, no daemons. Candidate follow-up card: a codex-suite daemon escapes the #1159 RAII teardown.
2. The hung-gh test costs a real ~5s per run (it must wait out the full budget to prove the kill). Acceptable; flagging for awareness.
3. `gate_recovers…` re-probes spawn real subprocesses twice per `block()` on the failure path — at the 120s tick cadence this is negligible (~3 extra spawns per failed tick max).

### CLEANUP
Sentinel worktree restored clean (git status empty), mutations all reverted, temp worktree + target dir removed after this comment. Zero temp-home daemons left running (the pre-existing leak instances were killed and are noted above).

PR #1163, opened 2026-06-13
perf(daemon): reverse-paged Inbox{since:None,limit:N} + daemon-routed cursor reads (card 8428ae8c)

joelteply:
## What / why

Card `8428ae8c` (P2) — the residual from #1144 (card a1562dbc, O(1) `room_tip`). Two items:

### 1. Indexed reverse paging for `Inbox { since: None, limit: N }`

The "most recent N" inbox path still materialized the WHOLE room before truncating: `handle_inbox` called `resume_from_cursor(channel, None)` (deep sink page with `usize::MAX` + ring merge + sort + dedup), then kept the newest N. #1144 removed the dominant caller (the work-board tip probe) but any remaining most-recent-N caller — e.g. `airc-lib`'s `daemon_page_recent` / `daemon_page_recent_subscribed` — paid O(room) per call.

This adds **reverse paging at the store layer**, wired daemon → sink:

- **airc-bus**: new required `DurableSink::page_tail(channel, before, limit)` — the last `limit` events strictly before a cursor (or the channel tail when `None`), ascending. **NO default impl** (mirroring #1144's removal of the scan-shaped `head_cursor` default): a sink that cannot reverse-page is a **compile error**, never a silent forward scan. `EventRouter::durable_tail(channel, limit)`: ring durables are the channel's newest durables (oldest-first eviction + §3.8 pinned-until-persisted ⇒ everything older is in the sink), topped up by ONE `page_tail` strictly before the ring's oldest retained cursor — no dup, no gap, durable-only. A sink error **propagates**; there is no scan fallback.
- **airc-store**: SQLite `page_tail` = `ORDER BY (epoch, counter, event_id) DESC LIMIT n` with a strictly-before generational predicate (exact mirror of `page`'s strictly-after), reversed in memory. It rides the **existing** composite index `idx_bus_events_room_epoch_counter_event_id` (verified in `m20260526_000011_create_bus_events.rs`; SQLite B-tree indexes serve both scan directions with an equality prefix on `room_id`) — no migration needed.
- **airc-daemon**: `handle_inbox`'s `since: None` arm now answers from `durable_tail`. The `since: Some` resume arm is unchanged. **No IPC change** — `InboxRequest`/`InboxResponse` wire shapes untouched.

### 2. Daemon-attached cursor reads no longer bypass the daemon

`Airc::latest_cursor` / `Airc::subscription_cursor` read the LOCAL store even when daemon-attached — pre-existing inconsistency (flagged in #1144's residual notes): under the owner-core model the daemon's ORM is the transcript and an attached scope's local store can be empty/stale, so those reads answered `None`/old for rooms with live daemon history. #1144's `room_tip` makes asking the daemon O(1), so both now route through the new shared `Airc::channel_latest_cursor` → `daemon_latest_transcript_cursor` (`room_tip`) when attached, local store otherwise. `work.rs::room_latest_cursor` (which already branched) is deduplicated onto the same helper. Local (non-attached) instances are unchanged.

**No new IPC op was needed for `subscription_cursor`** (the card asked to check): a subscription cursor IS the tip of the subscription's room — `room_tip` (card a1562dbc) is the typed op; the subscription set itself is local scope config. Audited: nothing else in the workspace reads a channel tip behind the daemon's back.

## O(room) → O(N) evidence

**Measured, live daemon over a Unix socket, real SQLite ORM, debug build, idle machine, same benchmark file at base vs head** (`airc-daemon/tests/inbox_paging.rs::deep_room_most_recent_n_costs_by_n_not_room_depth`, 5,000-event room, 20 probes each):

```
BEFORE (80122ad):  inbox(since:None, limit:50):  105,820 µs/op
AFTER  (this PR):  inbox(since:None, limit:50):    6,204 µs/op   (~17x)
full tail (limit:5000, the O(room) floor):  636,248 → 638,795 µs/op (unchanged)
```

The remaining ~6 ms is N-proportional work (50-row indexed read + envelope encode + IPC round-trip), independent of room depth; the full-tail floor is untouched, so the win is pure removal of the O(room) materialize.

Corroborated by #1144's own benchmark (`room_tip_probe.rs`, same 5,000-event room): `inbox(since:None, limit:1)` — measured at **99,129 µs/op** when #1144 landed — now answers in **511 µs/op (~194x)**, within ~1.7x of the `room_tip` probe itself (302 µs/op). That converged so hard it **broke #1144's `probe*2 < scan` timing race** (the O(room) leg it raced against no longer exists): `deep_room_tip_probe_matches_inbox_scan_and_is_cheaper` is renamed `deep_room_tip_probe_matches_inbox_newest`, keeping the cursor-agreement pin + printed numbers and dropping the now-noise race between two bounded paths (history preserved in its doc comment).

**Structural proof, not vibes** (`airc-bus/tests/durable_tail.rs`): counting-sink decorator — most-recent-200 on a 5,000-deep room (ring capacity 64, depth genuinely in the sink) must complete with **zero `page` calls** (the forward-scan primitive), **≤1 `page_tail`** call, and a requested reverse-page limit **≤ N**, while returning exactly the newest N ascending across the sink→ring seam. The no-fallback contract is enforced with a sink whose `page` *panics* and whose `page_tail` errors: `durable_tail` surfaces the error, never scans.

## Tests

- `airc-bus/tests/durable_tail.rs` (6): 5,000-deep bounded-work proof; ring-covered N does zero store reads; N > room + empty room + limit 0; non-durable traffic excluded; §3.8 un-persisted pending durable served from the ring (gated sink); sink error is loud (no scan fallback).
- `airc-store` `bus_sink.rs` unit tests (2): `page_tail` newest-N ascending / N>channel / empty / channel isolation; strictly-before predicate across epochs (epoch dominates, cursor row excluded).
- `airc-daemon/tests/inbox_paging.rs` (2): live-daemon correctness pins (exact ascending tail, newest cursor = last receipt, N > room, empty room, stream-chunk burst excluded) + the measured deep-room evidence with a **10x** margin (deliberately not 2x: the old full-materialize shape already sat ~6x under the full tail since it skipped encoding 4,950 envelopes — a 2x gate could not detect a regression to it; the paged path measures ~100x under, leaving an order of magnitude of CI headroom).
- `airc-lib/tests/daemon_cursor_reads.rs` (2): attached scope whose local store never saw the events reads the daemon tip via `latest_cursor` and `subscription_cursor`; unsubscribed channel still answers `None` (the subscription gate is unchanged).

## Mutation evidence (each applied, observed failing, restored)

- **M1** — revert `EventRouter::durable_tail` to the full-materialize shape (`resume_from_cursor(channel, None)` + retain + truncate): 4/6 `durable_tail.rs` tests **FAILED** (zero-`page` structural gate; panicking no-fallback sink). Restored → green.
- **M2** — revert ONLY the `handle_inbox` wiring to the old full-room replay (leaving `durable_tail` intact): `inbox_paging.rs` deep test **FAILED** the 10x margin (91,085 µs/op vs 551,960 — ~6x). Restored → green. This is exactly the regression the 2x-style margin would have missed.
- **M3** — revert `channel_latest_cursor` to always-local: both `daemon_cursor_reads.rs` tests **FAILED**. Restored → green.
- **M4** — drop the in-memory reversal in SQLite `page_tail` (DESC handed to the caller): both `bus_sink` `page_tail` unit tests **FAILED**. Restored → green.

(No wire pins to mutate: this PR adds no IPC variants/fields — item 1 changes only who answers `Inbox{since:None}`, item 2 reuses `room_tip`'s pinned wire shape from #1144.)

## Gate

- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` → clean
- `cargo fmt --all -- --check` → clean
- `cargo test --workspace -- --skip codex_hook` → all green (codex_hook skipped per macOS hang)
- Zero leaked temp-home daemons (in-process daemon fixtures only; #1159/#1161 guarded harness untouched)

Card: 8428ae8c-e62d-40de-8de2-19a81633f614
Lineage: #1144 (card a1562dbc) — this card is its declared residual.

### DIFF
--- a/crates/airc-bus/src/router.rs
+++ b/crates/airc-bus/src/router.rs
@@ -681,6 +681,68 @@ impl EventRouter {
         self.inner.sink.head_cursor(channel).await
     }
 
+    /// **Card 8428ae8c.** The newest `limit` **durable** envelopes on
+    /// `channel`, in ascending total order — the "most recent N" leg of
+    /// the inbox path. Work is bounded by `limit` and the ring capacity,
+    /// NEVER by room depth: the previous shape
+    /// (`resume_from_cursor(channel, None)` + truncate) materialized the
+    /// entire room to answer N.
+    ///
+    /// Merge contract (mirrors [`Self::subscribe`]'s seam, reversed):
+    /// - the hot ring is authoritative for the recent tail — it may hold
+    ///   durables the sink lacks (write-behind lag), and eviction is
+    ///   oldest-first with §3.8 pinning (a durable is never evicted
+    ///   before it is persisted), so the ring's durables are exactly the
+    ///   channel's newest durables and everything older is in the sink;
+    /// - when the ring's durables don't cover `limit`, the remainder
+    ///   comes from ONE [`DurableSink::page_tail`] call for the events
+    ///   strictly before the ring's oldest retained cursor (`None` =
+    ///   channel tail when the ring is empty — fresh-start shape). The
+    ///   sink only ever holds durables, so no class filter is needed on
+    ///   that leg. A sink error propagates — it is never papered over
+    ///   with a full forward scan.
+    pub async fn durable_tail(
+        &self,
+        channel: RoomId,
+        limit: usize,
+    ) -> crate::Result<Vec<Arc<Envelope>>> {
+        if limit == 0 {
+            return Ok(Vec::new());
+        }
+        let (mut tail, ring_oldest) = {
+            let shard = self.shard_for(channel);
+            let map = shard.channels.lock().unwrap_or_else(|p| p.into_inner());
+            match map.get(&channel.0.as_u128()) {
+                Some(state) => (
+                    state
+                        .ring
+                        .replay_after(None)
+                        .into_iter()
+                        .filter(|env| env.delivery.is_durable())
+                        .collect::<Vec<_>>(),
+                    state.ring.oldest_cursor(),
+                ),
+                None => (Vec::new(), None),
+            }
+        }; // shard lock released before the sink await
+        if tail.len() >= limit {
+            // The ring alone covers N — zero store reads.
+            return Ok(tail.split_off(tail.len() - limit));
+        }
+        // Ring durables are the newest; top up with the sink tail older
+        // than anything the ring still retains. Sink rows strictly
+        // before `ring_oldest` cannot also be in the ring, so the two
+        // legs concatenate with no dup and no gap.
+        let older = self
+            .inner
+            .sink
+            .page_tail(channel, ring_oldest, limit - tail.len())
+            .await?;
+        let mut out: Vec<Arc<Envelope>> = older.into_iter().map(Arc::new).collect();
+        out.append(&mut tail);
+        Ok(out)
+    }
+
     /// Subscribe (§4 subscribe, §3.5 cursor contract).
     ///
     /// Returns a stream that yields **every** envelope on the filter strictly
--- a/crates/airc-bus/src/sink.rs
+++ b/crates/airc-bus/src/sink.rs
@@ -61,6 +61,25 @@ pub trait DurableSink: Send + Sync {
     /// SQLite, back-of-vec on in-memory) or fail loudly.
     async fn head_cursor(&self, channel: RoomId) -> Result<Option<Cursor>>;
 
+    /// **Card 8428ae8c — required, no scan default.** Return the **last**
+    /// `limit` events on `channel` strictly *before* `before` (or the
+    /// channel's tail when `before` is `None`), in ascending total order
+    /// `(epoch, counter, event_id)`. This is the reverse-paging leg of
+    /// the "most recent N" inbox path: the previous shape materialized
+    /// the WHOLE channel via [`DurableSink::page`] and truncated in
+    /// memory — O(room) for an answer of size N. Every impl must answer
+    /// in work bounded by `limit` (SQLite: `ORDER BY … DESC LIMIT N` on
+    /// the composite `(room_id, epoch, counter, event_id)` index, then
+    /// reverse in memory; in-memory: a back-of-vec slice) or fail
+    /// loudly. There is deliberately NO default impl: a sink that cannot
+    /// reverse-page is a compile error, never a silent forward scan.
+    async fn page_tail(
+        &self,
+        channel: RoomId,
+        before: Option<Cursor>,
+        limit: usize,
+    ) -> Result<Vec<Envelope>>;
+
     /// **Card 4132f48c.** Whether an event with this `event_id` is already
     /// persisted. This is the durable leg of
     /// [`crate::EventRouter::publish_if_new`]'s idempotency check: an
@@ -180,6 +199,26 @@ impl DurableSink for InMemoryDurableSink {
             .cloned()
             .collect();
         Ok(out)
+    }
+
+    async fn page_tail(
+        &self,
+        channel: RoomId,
+        before: Option<Cursor>,
+        limit: usize,
+    ) -> Result<Vec<Envelope>> {
+        let guard = self.inner.lock().unwrap_or_else(|p| p.into_inner());
+        let Some(bucket) = guard.events.get(&channel.0.as_u128()) else {
+            return Ok(Vec::new());
+        };
+        // The bucket is kept in ascending total order on insert, so the
+        // tail strictly before `before` is a contiguous back slice.
+        let end = match &before {
+            None => bucket.len(),
+            Some(b) => bucket.partition_point(|e| e.cursor().is_before(b)),
+        };
+        let start = end.saturating_sub(limit);
+        Ok(bucket[start..end].to_vec())
     }
 
     async fn contains(&self, event_id: airc_core::EventId) -> Result<bool> {
--- a/crates/airc-bus/tests/common/mod.rs
+++ b/crates/airc-bus/tests/common/mod.rs
@@ -141,6 +141,15 @@ impl DurableSink for GatedSink {
         self.inner.head_cursor(channel).await
     }
 
+    async fn page_tail(
+        &self,
+        channel: RoomId,
+        before: Option<Cursor>,
+        limit: usize,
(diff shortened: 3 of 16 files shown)


joelteply:
## Sentinel review — card 8428ae8c (adversarial, default-BLOCK)

Reviewed PR head `8e4dafa` in an isolated temp worktree (`/tmp/sentinel-8428ae8c`, detached checkout, separate `CARGO_TARGET_DIR`; no production daemon touched).

### VERDICT: APPROVE

### EVIDENCE (all run in the temp worktree, not the author's)

**Production gate, verbatim:**
- `cargo clippy --workspace --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic -D clippy::todo -D clippy::unimplemented -D clippy::wildcard_enum_match_arm` → **clean** (exit 0)
- `cargo fmt --all -- --check` → **clean** (exit 0)
- `cargo test --workspace -- --skip codex_hook` → **106 test binaries, 0 failures**

**FRESH mutation re-runs (each applied in the sentinel tree, observed failing, restored via `git checkout`, re-run green; tree verified clean after):**
- **SM1** — `EventRouter::durable_tail` reverted to the exact scan shape this card kills (`resume_from_cursor(channel, None)` + durable filter + truncate): **4/6** `airc-bus/tests/durable_tail.rs` tests **FAILED** (zero-`page` counting gate; ring-covered zero-store-reads gate; the no-fallback sink whose `page` *panics*). Restored → 6/6 green.
- **SM2** — ONLY the `handle_inbox` `since: None` arm reverted to full-room replay (`durable_tail` left intact and unused): `inbox_paging.rs` deep test **FAILED** its 10x margin — mutated path measured **89,921 µs/op** vs the 552,344 µs/op full tail (~6x, exactly the regression a 2x margin would have missed). Restored → green. The margin choice is sound: the genuine paged path measures ~100x under the floor (see numbers below), leaving an order of magnitude of CI headroom, and the ratio rides one daemon so contention cancels.
- **SM3** — `channel_latest_cursor` reverted to always-local: both `daemon_cursor_reads.rs` tests **FAILED** (the discriminating actor is a scope whose local store never saw the events). Restored → green.
- **SM4** — SQLite `page_tail` DESC page handed to the caller un-reversed: both `bus_sink` `page_tail` unit tests **FAILED**. Restored → green.

**Bounded-work claim is TESTED, not asserted:**
- Structural: counting-sink decorator over a 5,000-deep room (ring capacity 64, depth genuinely in the sink) — most-recent-200 completes with **zero `page` calls**, **≤1 `page_tail`**, requested reverse limit **≤ N**, exact newest-N ascending across the sink→ring seam. Ring-covered N does **zero** store reads.
- Measured in the sentinel worktree (live daemon, Unix socket, real SQLite ORM, debug build): `inbox(since:None, limit:50)` on the 5,000-event room = **5,084 µs/op** vs the 553,194 µs/op O(room) full tail (~109x under the floor; author measured 105,820 µs/op for the same op at base — ~17–21x improvement). #1144's own benchmark corroborates: `inbox(limit:1)` = **490 µs/op** (was 99,129 at #1144 time, ~200x), now within 1.6x of `room_tip` itself (308 µs/op).

**No-fallback audit (the claim, grepped at head):**
- `handle_inbox` `since: None` arm reaches ONLY `router.durable_tail`; `durable_tail`'s only store read is `page_tail` (no `.page(`, no `resume_from_cursor` reachable on that arm).
- `DurableSink::page_tail` trait declaration ends in `;` — **no default impl**; a sink that cannot reverse-page is a compile error (verified by the trait ripple: 7 impls had to implement it). `FailingTailSink`'s `page` panics and its `page_tail` errors — `durable_tail` surfaces the error.
- SQLite leg rides the **existing** composite index `idx_bus_events_room_epoch_counter_event_id` (`m20260526_000011_create_bus_events.rs`) — B-tree indexes serve DESC with the `room_id` equality prefix; no migration needed, none added.
- Workspace-wide, the only production `Inbox{since:None}` caller is `airc-lib::daemon_page_recent` — now served by the bounded path. No other tip/tail read bypasses the daemon when attached.

**Wire surface:** no IPC variants/fields added or changed (`InboxRequest`/`InboxResponse` untouched; item 2 reuses #1144's pinned `room_tip` shape), so there are no new wire pins to mutate — the additive-op question the card raised is answered with reuse: a subscription cursor IS its room's tip.

### RISK NOTES
- **#1144's perf race retired by success**: `deep_room_tip_probe_matches_inbox_scan_and_is_cheaper` asserted `room_tip*2 < inbox(limit:1)`; this card made the inbox leg bounded too (490 vs 308 µs/op — both IPC-dominated), so the race between two O(1) paths is scheduling noise. The test is renamed `deep_room_tip_probe_matches_inbox_newest`, keeps the cursor-agreement pin + printed numbers, and documents the history. The hard no-scan gates live in the structural suites (`durable_tip.rs`, `durable_tail.rs`), which is the right place. Verified the original assertion genuinely fails at head (the author hit it in their full run) — the removal is forced, not cosmetic.
- `durable_tail`'s no-gap merge leans on §3.8 pinning (a durable is never evicted from the ring before it is persisted) — the same invariant `durable_tip` (#1144) and `resume_from_cursor` already lean on; the gated-sink test pins the ring-leads-sink half (un-persisted durable served from the ring).
- Tail-read is not atomic with concurrent publishes; the returned window can trail an in-flight publish by design — identical semantics to the scan it replaces. Not a blocker.
- `subscription_cursor` keeps answering `None` for unsubscribed channels (gate unchanged, pinned in the test); the `since: Some` resume arm of `handle_inbox` still materializes the full post-cursor suffix before truncating — pre-existing, out of this card's scope, bounded by the resume window rather than room depth.
- `cargo test (windows-self-hosted)` is the documented infra exception vs merge-base; hosted legs are the gate.

Sentinel cleanup: temp worktree + isolated target dir removed after verdict. No production daemon was started or stopped (test daemons were in-process fixtures in temp homes; zero leaked processes verified by ps sweep).

PR #1164, opened 2026-06-13
fix(setup): fresh Windows/new-user install — airc on PATH, gh wiring, git identity

joelteply:
Caught live driving the installer end-to-end on a **clean Windows Git Bash box** (no prior shell rc files, `gh` unauthed at install, no Rust toolchain, no git identity) — the exact fresh-operator path. Four bugs that each break a new user; all fixes are identity-agnostic and gh/`$HOME`-derived (multi-tenant safe — works the same for any operator linking their own grid).

## Fixes

1. **`airc` landed on PATH in no shell** — `install.sh` `_add_path_entry` looped `for rc in ~/.zshrc ~/.bashrc; do [ -f "$rc" ] || continue`. A fresh Git Bash box ships *none* of those, so the PATH line was written nowhere and the only export lived in the installer's own subshell. Now: create the rc matching `$SHELL` when none exists (+ a `~/.bash_profile` shim so login bash sources `~/.bashrc`). Also hardened the stale-line strip with `|| true` — when the rc holds only airc-managed lines, `grep -v` exits 1 and the `&&  mv` was skipped, duplicating the entry on rerun.

2. **`bootstrap-airc.sh` preflight called `airc doctor --connect`** — a flag the rust-rewrite binary doesn't implement, hard-failing every fresh bootstrap at the gate. Switched to `airc doctor --health`.

3. **gh auth after install left git's credential helper unwired** — `install.sh` only runs `gh auth setup-git` when gh is authed *at install time*; auth-after-install caused GUI password popups on gist ops. `bootstrap-airc.sh` now pins `-h github.com` to match `install.sh` and runs `setup-git` (idempotent) after login.

4. **No git author identity on a fresh box** — first agent commit dies with "Author identity unknown". `install.sh` now derives `user.name`/`user.email` from `gh api user` when unset (public email, else the `<id>+<login>` noreply alias). Never clobbers an existing identity. **No hardcoded identity** — fully derived per-operator.

## Validation
- Fresh-HOME sandbox: PATH entry created + resolved by a real login shell; idempotent (3 reruns → 1 line); zsh path; existing user rc content preserved.
- Real box: re-ran installer → `.bashrc` + `.bash_profile` created, login shell resolves `airc`, credential helper wired.
- Stubbed gh: identity derived for a fresh account, preserved when already set.

## Known follow-ups (not in this PR)
- **README + several installed skills reference commands the rust-rewrite binary doesn't have** (`teardown`, `quit`, `list`, `nick`, `away`, `back`, `uninstall`, `doctor --connect`). Docs/skills were written for the old `main` surface. Large reconciliation — worth a dedicated card.
- **Windows update path**: a running daemon locks `airc.exe`, so re-running the installer can't overwrite the binary until `airc stop`. First install is unaffected.

### DIFF
--- a/bootstrap-airc.sh
+++ b/bootstrap-airc.sh
@@ -47,16 +47,29 @@ else
   ok "airc already on PATH: $(command -v airc)"
 fi
 
-# 2. pre-flight (catches Tailscale-down, gh-missing, network-out, etc.)
-step "Pre-flight: airc doctor --connect"
-if ! airc doctor --connect; then
+# 2. pre-flight (live route/process state — catches daemon/route issues
+#    before join). The rust-rewrite `airc doctor` exposes `--health` for
+#    this; the old `--connect` flag no longer exists and made this
+#    pre-flight hard-fail on every fresh rust install. Fixed 2026-06-13.
+step "Pre-flight: airc doctor --health"
+if ! airc doctor --health; then
   die "Pre-flight failed. Fix the items above, then re-run this script."
 fi
 
-# 3. gh auth if needed
+# 3. gh auth if needed. Match install.sh's invocation exactly:
+#    -h github.com pins the host (avoids the interactive host picker),
+#    -s gist requests the scope the substrate needs. After a successful
+#    login, wire gh's token into git's credential helper so gist
+#    fetch/push (airc's rendezvous hot path) doesn't pop a password
+#    prompt on every op. Without setup-git, auth-after-install left the
+#    helper unwired — caught live on Windows 2026-06-13.
 if ! gh auth status >/dev/null 2>&1; then
   step "Authenticating gh (need 'gist' scope for room substrate)"
-  gh auth login -s gist
+  gh auth login -h github.com -s gist
+fi
+# Idempotent (no-op if already configured); safe to always run.
+if ! git config --global --get-all credential.https://github.com.helper 2>/dev/null | grep -q 'gh auth git-credential'; then
+  gh auth setup-git 2>/dev/null && ok "gh token wired into git credential helper" || true
 fi
 
 # 4. join the room
--- a/install.sh
+++ b/install.sh
@@ -431,6 +431,37 @@ ensure_prereqs() {
       fi
     fi
   fi
+
+  # Git author identity. Agents in this substrate commit and open PRs,
+  # and a fresh machine has no global user.name/user.email — the first
+  # commit then dies with "Author identity unknown" (caught live on a
+  # clean Windows box 2026-06-13). Derive it from the authenticated gh
+  # account when unset; never clobber an identity the user already set.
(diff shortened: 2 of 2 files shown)


joelteply:
SENTINEL VERDICT: APPROVE

This is the adversarial-reviewer sign-off standing in for the GitHub approval button (same-author self-approve is blocked). I tried hard to block this and could not justify it.

Scope: exactly two files (install.sh, bootstrap-airc.sh). Verified against base rust-rewrite with fake HOME/SHELL harnesses in /tmp (PR files unmodified; full installer not run).

What I verified:
- macOS/zsh fresh box (SHELL=*/zsh, no rc files): writes ~/.zshrc. Interactive Terminal.app shells read ~/.zshrc, so airc lands on PATH. No .zprofile shim needed — unlike login bash, zsh login+interactive sessions DO source .zshrc. The asymmetry (bash needs the .bash_profile shim, zsh does not) is correct, not a gap.
- Linux ~/.profile-only edge: new code creates ~/.bashrc and writes there (sourced by interactive desktop terminals). Crucially this is a STRICT IMPROVEMENT over base — the old loop wrote the PATH line to NOTHING in this case (the exact bug being fixed). The narrow login-shell-only-where-.profile-does-not-chain-.bashrc gap is pre-existing and out of scope for `open a terminal`.
- Idempotency: rerun 3x leaves exactly one `# airc` line (install.sh:588-598). The `|| true; mv` decoupling (595-596) correctly handles the rc-contains-only-airc-line case — no duplication, no wipe. User content (aliases/EDITOR) preserved; only airc-marked lines stripped. Stale airc lines in non-target rc files are also reconciled (Pass-1 design).
- No-trailing-newline glue bug fixed (install.sh:634-637): export lands on its own line.
- git identity block (install.sh:443-464): never clobbers an existing identity (guarded by per-field _need_* checks); fills only the missing field on partial identity; offline `gh api user` failure degrades gracefully (all `|| true`/`// empty`), no abort under set -euo pipefail; noreply fallback emits the correct <id>+<login>@users.noreply.github.com format.
- set -euo pipefail safety: the `if ! git config --get-all ... | grep -q` guards (install.sh:427, bootstrap:75) do not abort when the key/helper is absent — condition-context pipelines do not trip set -e. The bootstrap `gh auth setup-git && ok || true` tail is abort-safe. `gh auth login` is intentionally NOT wrapped (unauthenticated bootstrap should hard-fail before join — correct).
- No hardcoded identity/paths in logic: only `$HOME`/`$SHELL`/`~`/`github.com` host literal/gh-derived values. The single `Joel` hit is a pre-existing dated comment (2026-05-20) — acceptable per repo convention.
- `airc doctor --health` confirmed real via `airc doctor --help`; `--connect` confirmed removed (0 occurrences). Sane preflight replacement.

No blocking issues found. Approving.

PR #1165, opened 2026-06-13
fix(setup): PowerShell install parity — doctor flag, gh wiring, git identity

joelteply:
Mirrors the bash-side fresh-user fixes (#1164) onto the **native-Windows PowerShell** install path, so PowerShell users get the same foolproof bootstrap as Git Bash / mac / linux. Validated on a clean Windows box. All identity-agnostic and gh-derived (multi-tenant safe).

## Fixes
1. **`bootstrap-airc.ps1` preflight `airc doctor --connect`** → `airc doctor --health` (the `--connect` flag doesn't exist in rust-rewrite — same bug as #1164's bash fix).
2. **gh auth in `bootstrap-airc.ps1`**: pin `-h github.com` (matches install.sh/.ps1, skips host picker) + run `gh auth setup-git` after login so gist ops don't pop a credential prompt.
3. **git author identity** derived from `gh api user` when unset — in **both** `bootstrap-airc.ps1` (post-auth, the fresh path) and `install.ps1` (guarded on gh already authed). Never clobbers an existing identity; public email else the `<id>+<login>` noreply alias. **No hardcoded identity.**

## Not changed
- **PATH** — install.ps1 already writes the *registry User PATH* (`SetEnvironmentVariable('PATH', …, 'User')`), which new shells pick up. The bash rc-file trap (#1164 bug 1) does not exist on PowerShell, so no change needed there.

## Validation
- Both scripts AST-parse clean (`[Parser]::ParseFile`).
- Stubbed-gh test of the identity block: derives for a fresh account (noreply fallback), preserves an existing identity (no clobber).

## Related follow-up (separate PR)
- The **bash** `bootstrap-airc.sh` has the same identity gap on the truly-fresh path (install runs pre-auth → install.sh identity block skipped → bootstrap then auths but doesn't set identity). Will fix after #1164 merges to avoid a conflict on that file.

### DIFF
--- a/bootstrap-airc.ps1
+++ b/bootstrap-airc.ps1
@@ -116,7 +116,11 @@ if ($LASTEXITCODE -ne 0) {
 }
 # Idempotent: wire the credential helper if gh isn't already registered.
 $ghHelper = & git config --global --get-all credential.https://github.com.helper 2>$null
-if ($ghHelper -notmatch 'gh auth git-credential') {
+# Join to a scalar before -notmatch: --get-all can return multiple helper
+# values (array), and PS -notmatch on an array filters rather than returning
+# a strict boolean. Worst case without this is a redundant (idempotent)
+# setup-git, but the scalar form is correct.
+if ((@($ghHelper) -join "`n") -notmatch 'gh auth git-credential') {
     & gh auth setup-git 2>$null
     if ($LASTEXITCODE -eq 0) { OK 'gh token wired into git credential helper' }
 }
--- a/bootstrap-airc.ps1
+++ b/bootstrap-airc.ps1
@@ -92,20 +92,56 @@ if (-not $airc) {
     OK "airc already on PATH: $($airc.Source)"
 }
 
-# 2. pre-flight (catches Tailscale-down, gh-missing, network-out, etc.)
-Step 'Pre-flight: airc doctor --connect'
-& airc doctor --connect
+# 2. pre-flight (live route/process state before join). The rust-rewrite
+# `airc doctor` exposes `--health`; the old `--connect` flag no longer
+# exists and made this pre-flight hard-fail on every fresh rust install.
+# Mirrors bootstrap-airc.sh. Fixed 2026-06-13.
+Step 'Pre-flight: airc doctor --health'
+& airc doctor --health
 if ($LASTEXITCODE -ne 0) {
     FailOut 'Pre-flight failed. Fix the items above, then re-run this script.'
 }
 
-# 3. gh auth if needed
+# 3. gh auth if needed. Pin -h github.com (matches install.sh / .ps1, skips
+# the interactive host picker) and -s gist for the substrate scope. After
+# a successful login, wire gh's token into git's credential helper so gist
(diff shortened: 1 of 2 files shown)


joelteply:
SENTINEL VERDICT: APPROVE

Adversarial reviewer sign-off, standing in for the gh approval button (gh blocks self-approval; this comment is the doctrinal substitute). I tried to block this and could not justify it — every attack landed on solid ground.

**What I attacked, and why each block failed:**

- **`doctor --health` real?** Confirmed via `airc doctor --help`: `--health` exists ("Include live route/process health"); `--connect` is gone. The flag swap in bootstrap-airc.ps1:101 is correct and necessary — the old flag hard-failed every fresh rust install.
- **Native nonzero exit under `$ErrorActionPreference='Stop'`?** Tested on PS 5.1: `(& git config --global user.name) 2>$null` returning exit 1 on an unset key does NOT throw — native exit codes aren't terminating errors. `$LASTEXITCODE` flips to 1 but the script continues. No mid-run abort. (bootstrap:122-123, install.ps1:486-487)
- **Empty/null falsiness?** With identity unset, `$gitName`/`$gitEmail` are empty strings; `-not $x` is reliably `True`. Verified.
- **noreply fallback?** `gh api user --jq '.email // empty'` returning empty triggers `"$ghId+$ghLogin@users.noreply.github.com"` — string interpolation correct on 5.1 (the `@` is literal, not a splat in a double-quoted string). Verified across fresh / real-email / partial / full matrix.
- **No clobber?** Per-field guards (`if (-not $gitName ...)`, `if (-not $gitEmail ...)`) confirmed: an existing identity is never overwritten; only the missing field is filled. Both-set case touches nothing. Verified.
- **Hardcoded identity / paths?** Zero name/email literals, zero absolute user paths in the added lines — only `$env:`, gh-derived values, and the `github.com` host literal. Multi-tenant-safe. Confirmed by grep over the diff.
- **AST parse:** Both files `ParseFile` clean under Windows PowerShell 5.1.
- **Mirror claims?** Faithful. The bash counterparts are in-flight: `9a2e898` (install.sh git-identity-from-gh, identical noreply format + no-clobber) and `ab92441` (bash bootstrap `--connect`→`--health`). install.ps1's identity block is correctly guarded on `gh auth status`; bootstrap's 3b is reached only post-successful-auth, so its lack of an explicit re-check is sound.

**One non-blocking nit (no change required, FYI for a follow-up):** `$ghHelper -notmatch 'gh auth git-credential'` — when `git config --get-all` returns MULTIPLE helper values (array), PS `-notmatch` acts as a filter (returns non-matching elements) rather than a strict boolean, so a non-empty array stays truthy even if one element matches. Effect: a redundant — but idempotent and harmless — `gh auth setup-git` re-registration in the rare multi-helper case. The single-helper case (overwhelmingly common) and the unset case both behave exactly right. bash uses `grep -q` which short-circuits; if you ever want exact parity, `($ghHelper -join "`n") -notmatch ...` would do it. Not worth holding the merge.

Net: correct flag, correct gh wiring, identity-agnostic git-author derivation that never clobbers, clean parse on both 5.1 and 7, faithful bash mirror. Ship it.

PR #1166, opened 2026-06-13
fix(ci): let non-Rust PRs through the merge queue (required-check gate)

joelteply:
## Problem (systemic, caught live on a fresh Windows install)
The merge queue refuses to ingest a PR whose required checks never ran — the enqueue API returns **"N of N required status checks are expected."** `cargo clippy (deny warnings)` and `cargo test (ubuntu-latest)` are required, but rust.yml's `pull_request` trigger was `paths`-filtered to Rust files, so they never triggered on docs/scripts/skills PRs.

**Net effect: every non-Rust PR was unmergeable through the queue** — not even `--admin` (the ruleset enforces "changes must go through the queue"). Install-fix PRs #1164/#1165 are stuck with no path in. The old rust.yml comment assumed the `merge_group` trigger covered this; it doesn't — merge_group only re-validates *inside* the queue, you can't *enter* without the contexts on the PR head.

## Fix
- Drop the `paths` filter on `pull_request` so rust.yml always triggers and the required contexts always report.
- Add a `changes` job (`gh api .../files` + grep — no third-party action) that detects Rust-relevant edits.
- `fmt`/`clippy`/`quality`/`test` gain `needs: changes` + `if: !cancelled() && (github.event_name != 'pull_request' || needs.changes.outputs.rust == 'true')`:
  - **non-Rust PR** → heavy jobs **skip** (a skipped required check counts as satisfied → queue can ingest), self-hosted Windows leg stays free.
  - **merge_group / push** → real legs run **unconditionally** (preserves the merge-race guard on the queued candidate).
  - **Rust PR** → real legs run.

## Self-validating
This PR edits `rust.yml` (a Rust-path file), so `changes` → `rust=true` → the real clippy + test legs run on *this* PR and it merges through the queue normally. A broken workflow fails its own checks and can't merge — blast radius is contained.

## Verification plan
After merge, I'll re-trigger #1164/#1165 (shell/PowerShell-only) to confirm a *skipped* required check satisfies queue ingestion end-to-end. If GitHub treats skipped-required differently for enqueue than expected, I'll switch the skip to an explicit always-green report.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,23 +21,26 @@ name: rust
 # Rust workspace as it grows.
 
 on:
+  # NO `paths` filter on pull_request — on purpose. The required contexts
+  # (`cargo clippy (deny warnings)`, `cargo test (ubuntu-latest)`) must
+  # REPORT on every PR, because GitHub's merge queue refuses to ingest a
+  # PR whose required checks never ran ("N of N required status checks
+  # are expected" from the enqueue API). A `paths` filter here meant the
+  # rust legs never triggered on docs/scripts/skills PRs, which silently
+  # made EVERY non-Rust PR unmergeable through the queue (caught live
+  # 2026-06-13 — #1164/#1165 stuck with no way in, not even --admin).
+  # The `changes` job below decides whether the heavy jobs do real work;
+  # non-Rust PRs SKIP them (a skipped required check counts as satisfied)
+  # so the self-hosted runner stays free. merge_group always runs the
+  # real legs regardless (see below).
   pull_request:
-    paths:
-      - "crates/**"
-      - "Cargo.toml"
-      - "Cargo.lock"
-      - "test/rust_quality_guard.sh"
-      - ".github/workflows/rust.yml"
   # Merge queue: GitHub builds the queued PR(s) on top of the latest
   # rust-rewrite on a `gh-readonly-queue/...` ref and emits a
-  # `merge_group` event. Without this trigger those checks never run,
-  # so the queue would wait forever. No `paths` filter here on purpose
-  # — `merge_group` doesn't honour path filters, and a queued docs-only
-  # PR still needs to report the required contexts (the rust legs pass
-  # trivially with no crate changes). This is the gate that catches the
-  # merge-race class that broke rust-rewrite on 2026-06-11 (#1127×#1137
-  # each green alone, broken combined): the queue compiles the COMBINED
-  # result before it can land.
+  # `merge_group` event. The heavy jobs run UNCONDITIONALLY on
+  # merge_group (their `if` short-circuits on event != pull_request), so
+  # the rust legs always validate the COMBINED queued result — the gate
+  # that catches the merge-race class that broke rust-rewrite on
+  # 2026-06-11 (#1127×#1137 each green alone, broken combined).
   merge_group:
   push:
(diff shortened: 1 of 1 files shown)


joelteply:
SENTINEL VERDICT: APPROVE

Adversarial review — tried hard to block, could not justify it. Every load-bearing assumption holds against both GitHub's documented behavior and live evidence from this PR's own run (27474736569).

**#2 (the load-bearing one) — skipped-required-check satisfies the merge-queue enqueue gate: CONFIRMED.** The original bug was *workflow-level* path filtering: when a workflow is skipped by a `paths` filter, its required contexts stay "Expected — Waiting for status to be reported" (pending) forever → queue cannot ingest. Per GitHub docs (Troubleshooting required status checks): a job skipped by a *job-level `if`* reports conclusion **Success**, which satisfies the required check; only *workflow-level* filtering leaves it stuck pending. This PR converts workflow-level filtering → job-level conditional skip — the exact canonical pattern GitHub recommends. The fix actually unblocks non-Rust PRs.

**#1 — `!cancelled()` + skipped `needs`: CORRECT.** Using any status-check function in `if` overrides the default "skip if a dependency was skipped." On `merge_group` the `changes` job is skipped, but fmt/clippy/quality/test still evaluate their `if`; `!cancelled()`=true and `event != pull_request`=true → they RUN unconditionally. Merge-race guard (#1127×#1137) intact: the queued COMBINED candidate is always compiled/tested at merge_group time, even for a docs-only PR. `!cancelled()` is the right guard (safer than `always()` — respects user cancellation).

**#3 — `changes` classification: CORRECT, no current false-negative.** Tested the regex across crate-only / docs-only / mixed / Cargo.toml / nested-crate-Cargo.toml / workflow-self cases — all classify right. Verified against actual repo layout: every `Cargo.toml` is root or under `crates/`; `build.rs` is at `crates/airc-cli/build.rs` (matched by `^crates/`); no `rust-toolchain.toml`/`.cargo/config.toml`. Token: repo `default_workflow_permissions` is `write`; live log shows `PullRequests: write`, the `gh api .../files` call succeeded and set `rust=true`. No `permissions:` block needed.

**#4 — pagination:** `gh api --paginate --jq` applies the filter per-page and concatenates; >30 files handled. Empty list → `rust=false` (harmless).

**#5 — YAML/tabs:** no tabs; GitHub parsed and executed the workflow with no startup/zero-jobs failure (the documented failure mode for a bad job-level `if`).

**#6 — security:** `pull_request` (not `_target`), read-only `gh api`, no checkout, hosted ubuntu. grep consumes API output as stdin *data*, not code — no injection. Fork PRs only get filenames listed on a hosted runner; the self-hosted leg is still protected by the existing step-level fork guard.

**Empirical proof on this very PR:** it touches `.github/workflows/rust.yml` → `changes` correctly emitted `rust=true` → all heavy legs ran (fmt/clippy/quality green, tests in flight). The Rust-PR path is demonstrably exercised and correct.

**Non-blocking note (latent, not a defect against the current tree):** the regex would false-negative if a future *root* `rust-toolchain.toml`, `.cargo/config.toml`, or a workspace member outside `crates/` were added and changed in isolation. None exist today, and adding a workspace member edits the matched root `Cargo.toml`. Consider widening the regex if the layout ever changes — but not a reason to block.

This is the sentinel sign-off (substitutes for the approval button; self-author cannot self-approve). Ship it.

PR #1170, opened 2026-06-13
fix(update): record install source so airc update works for dev checkouts

joelteply:
## Problem (found by running the real update path)
`airc update` resolves the source as `$AIRC_DIR` or hardcoded `~/.airc/src` (`update_commands.rs::install_source_dir`). But install.sh's `_default_clone_dir` installs FROM a dev checkout (the cwd) when run inside one — and **rust-rewrite currently ships ONLY as a dev checkout** (no release channel has the rewrite yet). So on every dev-checkout install, `airc update` dies with **"No git checkout at ~/.airc/src."** Since all current operators (incl. the other agents) are on rust-rewrite dev checkouts, this blocked the whole team from pulling fixes via `airc update`.

## Fix
- install.sh + install.ps1 record the dir they installed from in `~/.airc/install-source`.
- `install_source_dir()` resolution: `$AIRC_DIR` env → marker → `~/.airc/src` default.
- **Native path requirement:** the binary's `std::fs`/`git -C` can't resolve an MSYS `/c/...` path on Windows, so install.sh converts via `_to_win_path` on MINGW/MSYS/CYGWIN only (Unix written as-is); install.ps1's `$CLONE_DIR` is already native.

## Validation
- 3 new tests (marker honored / `$AIRC_DIR` wins / blank marker → default) + existing 4 pass.
- `cargo fmt` + `clippy` (incl. production no-unwrap gate) clean locally; CI's real legs run (Rust-path PR).

### DIFF
--- a/bootstrap-airc.sh
+++ b/bootstrap-airc.sh
@@ -72,6 +72,35 @@ if ! git config --global --get-all credential.https://github.com.helper 2>/dev/n
   gh auth setup-git 2>/dev/null && ok "gh token wired into git credential helper" || true
 fi
 
+# 3b. Git author identity. install.sh sets this too, but on the bootstrap
+# path install.sh runs BEFORE gh auth (non-TTY curl|bash), so its
+# identity block is skipped — and without this the first agent commit
+# dies with "Author identity unknown". Derive from the authenticated gh
+# account when unset; never clobber an identity the user already set.
+# Public email, else the <id>+<login> noreply alias. No hardcoded values.
+if command -v gh >/dev/null 2>&1 && gh auth status >/dev/null 2>&1; then
+  _need_name=0; _need_email=0
+  if [ -z "$(git config --global user.name 2>/dev/null || true)" ]; then _need_name=1; fi
+  if [ -z "$(git config --global user.email 2>/dev/null || true)" ]; then _need_email=1; fi
+  if [ "$_need_name" = 1 ] || [ "$_need_email" = 1 ]; then
+    _gh_login="$(gh api user --jq '.login' 2>/dev/null || true)"
+    _gh_name="$(gh api user --jq '.name // .login' 2>/dev/null || true)"
+    _gh_id="$(gh api user --jq '.id' 2>/dev/null || true)"
+    _gh_email="$(gh api user --jq '.email // empty' 2>/dev/null || true)"
+    if [ -z "$_gh_email" ] && [ -n "$_gh_id" ] && [ -n "$_gh_login" ]; then
+      _gh_email="${_gh_id}+${_gh_login}@users.noreply.github.com"
+    fi
+    if [ "$_need_name" = 1 ] && [ -n "$_gh_name" ]; then
+      git config --global user.name "$_gh_name"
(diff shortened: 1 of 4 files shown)


joelteply:
SENTINEL VERDICT: APPROVE

Adversarial sentinel sign-off (substitutes for the gh approve button; self-author cannot approve). I tried to break this and could not justify a block.

VERIFIED AGAINST EACH SCRUTINY POINT:

1. Windows native path — install.sh:768-771 writes `_to_win_path "$CLONE_DIR"` (cygpath -w → `C:\Users\...\airc`) ONLY on MINGW/MSYS/CYGWIN; the `*)` branch writes `$CLONE_DIR` as-is, so mac/linux paths are NOT mangled. Command substitution strips cygpath`s trailing newline; `printf %s\n` adds exactly one; Rust read_install_source_marker (update_commands.rs:181) `.trim()`s it. The native backslash form feeds PathBuf::from → source.join(".git") + `git -C` correctly. Good.

2. Precedence $AIRC_DIR > marker > ~/.airc/src — install_source_dir (update_commands.rs:153-173) checks AIRC_DIR first, then marker, then default. Canonical ~/.airc/src installs get a marker pointing at ~/.airc/src (harmless, same target) or none → default. Blank/whitespace marker returns None (line 182-184), falling back to default rather than an empty PathBuf. Covered by install_source_blank_marker_falls_back_to_default and install_source_airc_dir_beats_marker.

3. Production clippy gate — `cargo clippy -p airc-cli --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic` FINISHED CLEAN (exit 0). New lib code uses `?`, `.ok()?`, `if let` — no unwrap/expect/panic/todo. (Note: airc-cli is bin-only, no lib target, so `--lib` errors with "no library targets"; the gate reduces to `--bins`, which passed.)

4. Installer hygiene — install.sh marker write (757-774) runs AFTER `_install_airc_binary` under `set -euo pipefail`, so a failed build exits before the marker is written; `mkdir -p "$HOME/.airc"` precedes the write. install.ps1 (300-303) does `New-Item -Force` then Set-Content (CRLF trailing newline, handled by .trim()). Minor asymmetry: ps1 writes the marker BEFORE the build while sh writes AFTER — but the marker records the clone SOURCE (which exists once cloned), not build success, so a failed ps1 build still leaves a valid git-checkout marker for a later `airc update`. Not a correctness defect.

5. No hardcoded identity/paths in production lines — all derive from $HOME / $CLONE_DIR / $env:USERPROFILE. Test fixtures use throwaway /opt/dev/airc + /explicit/override (test-only).

TEST RESULTS (real, run locally):
- `cargo test -p airc-cli --bin airc update_commands::` → 7 passed; 0 failed (incl. the 3 new: install_source_reads_marker_when_no_env, install_source_airc_dir_beats_marker, install_source_blank_marker_falls_back_to_default). Existing validate_source + defaults tests still green — no regression.

Tests carry `// what this catches:` comments and link the 2026-06-13 dev-install bug, per house rules. Bug fix is real, minimal, well-scoped, and the marker is the right seam. Approving.

PR #1176, opened 2026-06-13
refactor(gh): unified request governor + gh/ subtree (slices 1-3a)

joelteply:
Makes the gh request governor the single enforceable chokepoint (one counter, no bypass), and groups the gh boundary into one subtree. **Validated live by the mesh-convergence harness (#1175)** — the governor fired across two separate container nodes (`30/30 in 60s`, shared backoff) and prevented gh spam.

- **Slice 1** — `airc-lib::gh::governor::GhBudget`: one file-locked `~/.airc/gh/` counter shared cross-process by cli + lib; GitHub's own rate-limit headers drive the backoff, with a safety floor (stop before zero).
- **Slice 2** — the account-registry store routes its `gh_run` through the governor; both GitHub limiters tracked per-machine.
- **Slice 3a** — collapsed scattered `gh_*` modules into one `airc-lib::gh/` subtree; crate-root re-exports unchanged.

Tests green; clippy `-D warnings` clean. (Local airc-lib test exe blocked by this box's Smart App Control — CI runs them on Linux.)

Remaining follow-ups: 3b file-split, `--include` feed, no-raw-gh guard, slice 4 conformance.


joelteply:
SENTINEL VERDICT: APPROVE

Adversarial sign-off (default-BLOCK posture; this is the doctrinal approval since the self-author cannot use the gh approve button). I tried to break it across all five scrutiny axes and could not justify a block. Reviewed against a fresh worktree of `origin/fix/gh-governor-unify` (commit b093ead), not the diff alone.

REAL EXIT CODES (worktree build, CARGO_TARGET_DIR pinned to shared cache):
- `cargo build -p airc-lib -p airc-cli` -> exit 0
- `cargo clippy -p airc-lib -p airc-cli --lib --bins -- -D warnings -D clippy::unwrap_used -D clippy::expect_used -D clippy::panic` -> exit 0
- SAC CAVEAT: this box`s Smart App Control blocks executing the airc-lib test exe (os error 4551). I did NOT run the lib test binary or the mesh-convergence harness locally; verdict rests on build + clippy + full source read. SAC was not disabled.

FINDINGS (all pass / non-blocking):

1. Deny-skip + fail-open (Slice 2) — CORRECT. `gh_run` reserves first: `Reservation::Denied` -> `Err(AccountRegistryError::Adapter(...))`; `Reservation::Allowed | Err(_)` -> falls through and spawns gh (fail-open). The refresh loop swallows that Err — `let _ = run_tick(...).await;` (registry_refresh.rs:325) — so a denial is skip-and-retry next cadence, NOT a hard failure. No double-count: `record()` is inside the locked RMW; `note_rate_limit` runs once per stream post-response regardless of the reserve outcome.

2. Floor dead-until-`--include` — CONFIRMED + HONESTLY DISCLOSED. No `gh_run` call passes `--include`, so the primary `x-ratelimit-*` floor/quota-snapshot path receives no headers today and is dead. The reactive prose fallback (`secondary rate limit` / `rate limit exceeded` / `abuse detection` -> 60s backoff) still fires, and `retry-after` still works if gh prints it. PR body lists the `--include` feed as a remaining follow-up — stated, not hidden. No unwrap/expect/panic in bin-reachable code (clippy gate confirms; all unwraps are `#[cfg(test)]`).

3. Reorg integrity — CLEAN. Zero leftover `gh_governor::` / `crate::gh_account_registry::` / `airc_lib::gh_client::` / `airc_lib::gh_account_registry::` refs on the PR tree. Crate-root re-exports unchanged: `GhClient`, `GhAccountRegistryStore`, `GhBudget`, `Reservation` all re-exported from lib.rs. Whole workspace builds. (The `crate::gh_client::` refs remaining in airc-cli point at the CLI`s own local re-export shim, which is correct.)

4. Lock scope + isolation — CORRECT. `reserve` holds the fs2 exclusive lock across the whole read-backoff -> recent_count -> record RMW (GuardLock RAII drop-unlocks at fn scope). Tests use `GhBudget::at(tmp)` with a fresh TempDir each = dir-isolated from `~/.airc/gh/`. Two NON-BLOCKING nits: (a) `note_rate_limit` writes backoff/quota lock-free, but writes are tmp+rename atomic and `write_backoff` does `until.max(read_backoff())` (monotonic — only ever loses the SMALLER backoff = fail-safe direction); the CLI`s `record_backoff` is identically lock-free, so behavior is consistent. (b) `reserve_allows_until_budget_then_denies...` and `note_rate_limit_throttles_at_the_floor...` mutate process-global env (`AIRC_GH_MAX_REQUESTS_PER_MIN`, `AIRC_GH_QUOTA_FLOOR`) with no serial guard — a parallel-test flake risk. Dirs are isolated and env is restored, so not a correctness bug, but a follow-up should serialize these or inject the limit instead of env-mutating.

5. No hardcoded identity/paths in ADDED lines — CLEAN. `grep -c joelteply` on the diff = 0; every `joelteply` / `/Users/joel` hit is pre-existing comment/test text that survived the rename unchanged. The new governor.rs uses HOME/USERPROFILE env only.

ONE-COUNTER CLAIM — VERIFIED AT THE FILE LAYER. The CLI still has its own governor impl (`airc-cli/src/gh_state.rs`), so this is two parallel CODE implementations. BUT they are byte-compatible on the SAME files: identical dir resolution (`~/.airc/gh/`), identical `budget.jsonl` (`{ts:.3}` lines), `backoff-until` (`as i64`), `guard.lock` (fs2 exclusive), same `AIRC_GH_MAX_REQUESTS_PER_MIN` / default 30 / 60s window. So cli + lib genuinely coordinate cross-process via one shared counter under the fs2 lock — the "no bypass" claim holds at the persistence layer, which is what matters. The duplicate impl (gh_state consolidation / no-raw-gh guard) is explicitly scoped as a future slice in the PR body, so it is disclosed scoping, not a concealed bypass.

COVERAGE GAP (non-blocking): the `with_budget()` test-injection hook is added (pub) but no test exercises the STORE-through-governor path (deny-skip / note_rate_limit feedback at the `GhAccountRegistryStore` layer). That integration is covered only by governor unit tests + the asserted (locally-unverifiable here) mesh-convergence harness. The store wiring is mechanically simple and read-verified; a focused integration test using `with_budget(GhBudget::at(tmp))` would close this and should be a follow-up.

VERDICT: APPROVE. The unification is real at the file/lock layer, the deny path self-heals correctly, fail-open is intentional and safe, the reorg is clean and builds+lints green, and every shortcoming is either an explicitly-disclosed follow-up or a test-hygiene nit — none are production correctness defects. Recommended (non-blocking) follow-ups: (1) serialize/inject the env-mutating governor tests; (2) add a store-level integration test via `with_budget`; (3) land the `--include` feed so the floor stops being dead code; (4) consolidate the duplicate `gh_state.rs` governor onto `GhBudget`.

PR #1177, opened 2026-06-14
fix(account-registry): prune stale peer beacons before enrol

joelteply:
## What

`merge_registry_documents` keeps the **freshest beacon per peer**, but never asks whether that freshest beacon is *itself* old enough to be dead. A peer whose publisher died and whose lingering gist beacon is days old was still enrolled — and then dialed. That's the **stale-route orphan path** Joel kept flagging: *"make sure those aren't just old ones."*

The convergence harness (`docker/mesh-converge.sh`) surfaced it concretely: both isolated nodes enrolled **11–12 peers when only 1 was live**.

## Fix

A reader-side freshness pass — the counterpart to the local coordinator's `drain_stale_store`:

- **`prune_stale_peers(peers, now_ms, ttl_ms) -> usize`** — pure, in-place. Reuses `PresenceBeacon::is_fresh` so the freshness decision lives in **one place** (compression principle). Saturating-sub keeps future-dated clock-skew peers — we never punish a peer for our clock.
- **`DEFAULT_PEER_FRESHNESS_TTL_MS = 600_000` (10 min)** — deliberately **>> the 60s local heartbeat TTL**. A live peer's beacon reaches another machine only after a `publish→gist→fetch` hop on the ~120s refresh cadence, so a couple-minutes-old beacon is normal transport latency, not death. Sized to drop only genuinely-dead publishers.
- Wired into **`GhAccountRegistryStore::refresh`** — the single production enrol chokepoint — and surfaced via a counted `AccountRegistryStaleBeaconsPruned` diagnostic, symmetric with the existing temp-beacon counter (count, not dump).

## Why additive (not a `merge_registry_documents` signature change)

`merge` has no clock and 4 mechanics tests that build tiny synthetic timestamps and assert retention. Threading `now_ms` through it would force noisy edits to tests that don't care about freshness. The freshness policy is an *enrol* concern, and there is exactly one production enrol path — so the filter lives there, beside where temp-rooted beacons are already dropped.

## Validation

- Unit test `prune_drops_only_peers_staler_than_ttl`: only the over-TTL peer drops; live-but-laggy and future-dated peers kept. Mutation-documented (drop the `is_fresh` filter → both asserts fail).
- `cargo clippy -p airc-lib -p airc-cli -p airc-diagnostics --all-targets -- -D warnings` clean.
- Lib tests green: account_registry **20**, gh **29**, diagnostics **5**.

### DIFF
--- a/crates/airc-diagnostics/src/lib.rs
+++ b/crates/airc-diagnostics/src/lib.rs
@@ -56,6 +56,11 @@ pub enum DiagnosticCode {
     /// temp-rooted (phantom hermetic-test peers in polluted documents).
     /// Carries the count, not the dump.
     AccountRegistryTempBeaconsIgnored,
+    /// Reader-merge pruned peers whose freshest beacon is older than the
+    /// freshness TTL — dead daemons whose only surviving gist beacon is
+    /// days old. Enrolling them would dial dead routes (the stale-route
+    /// orphan path). Carries the count, not the dump.
+    AccountRegistryStaleBeaconsPruned,
     WireLostEmitFailed,
     MalformedReplayFrameSkipped,
     UnverifiableReplayFrameSkipped,
--- a/crates/airc-lib/src/account_registry.rs
+++ b/crates/airc-lib/src/account_registry.rs
@@ -156,6 +156,39 @@ pub fn merge_registry_documents(
     }
 }
 
+/// Default freshness horizon for reader-side beacon pruning: 10 minutes.
+///
+/// Deliberately MUCH longer than the local coordinator's 60s heartbeat
+/// TTL ([`crate::coordinator::DEFAULT_HEARTBEAT_TTL_MS`]). A peer's
+/// beacon reaches another machine only after a publish→gist→fetch hop,
+/// and the production registry-refresh loop runs on a ~120s cadence, so
+/// a perfectly live peer's freshest visible beacon is routinely a couple
+/// of minutes old purely from transport latency. The local TTL would
+/// false-drop those. This horizon is sized to prune only beacons whose
+/// publisher is genuinely gone (process dead for many minutes / the gist
+/// is days stale) while never evicting a live-but-laggy peer.
+pub const DEFAULT_PEER_FRESHNESS_TTL_MS: u64 = 600_000;
+
+/// Drop peers whose freshest beacon is staler than `ttl_ms` relative to
+/// `now_ms`, in place. Returns the number pruned.
+///
+/// This is the reader-side counterpart to the local coordinator's
+/// [`crate::coordinator::drain_stale_store`]: [`merge_registry_documents`]
+/// already keeps only the FRESHEST beacon per peer, but a peer whose
+/// freshest beacon is itself ancient (the publisher died and its gist
+/// never got cleaned) would still be enrolled and then dialed — the
+/// stale-route orphan path. Pruning here means the enrol set never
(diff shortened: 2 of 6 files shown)


joelteply:
## Adversarial Sentinel Review — VERDICT: APPROVE

I went in default-BLOCK and tried hard to break this across all six flagged vectors; each cleared against the actual code.

**1. TTL safety (false-eviction of live peers).** Production republish cadence is publish-then-refresh every `DEFAULT_CADENCE = 120s` (registry_refresh.rs). Each remote machine restamps its own beacon every ~120s. The 600\_000ms (10min) horizon is ~5x that — a live peer must miss ~4 consecutive publish ticks before its freshest visible beacon ages past the TTL. The transport-latency argument in the doc-comment holds. Not a false-eviction risk.

**2. Destructive global mutation (the highest-risk vector).** Cleared decisively. `refresh` (gh/account_registry.rs) is read-only: list → fetch → merge → prune → return. Its only production caller is `refresh_account_registry` (account_registry.rs:498), which feeds the result into `import_account_registry_document` — that writes ONLY to the LOCAL trust store + coordinator store; it never re-publishes the merged doc. The `publish` path builds its document independently from local state (`account_registry_document`), not from the refreshed/pruned result. Pruning is a purely local freshness filter; it cannot delete anyone's beacon from the rendezvous gist for other readers.

**3. Endpoint backfill interaction.** Backfill in `merge_registry_documents` is strictly intra-peer (endpoints from older beacons of the SAME peer_id are merged onto that peer's endpoint-less freshest winner). Prune runs after merge, per-peer, on `peer.presence.is_fresh` (the freshest heartbeat). No cross-peer endpoint leakage; pruning a >10min-stale peer is correct regardless of any backfilled endpoints.

**4. `is_fresh` / `heartbeat_at_ms`.** `is_fresh` = `now_ms.saturating_sub(heartbeat_at_ms) < ttl_ms` — saturating-sub treats future-dated clocks as fresh, exactly as the comment and test claim. `heartbeat_at_ms` is a REQUIRED serde field on `PresenceBeacon` (no `#[serde(default)]`), so a beacon missing it fails document deserialization and is skipped at merge — it can never surface as a 0 that mass-prunes live peers.

**5. Test quality.** `beacon_at` passes its `heartbeat_ms` straight through `beacon_now` into `heartbeat_at_ms`, so the three fixtures genuinely sit at now-100s / now-900s / now+50s. With ttl=600s: fresh kept, stale (900s>600s) pruned, future-skew kept. The stated mutation — drop the `is_fresh` filter → pruned=0 and stale survives — fails both the count assert and the `!contains(stale)` assert. Real boundary behavior incl. the clock-skew edge, not a tautology. Verified green: `account_registry::tests::prune_drops_only_peers_staler_than_ttl ... ok`.

**6. Build hygiene.** `cargo clippy -p airc-lib -- -D warnings` is clean. `now_ms()` returns Result and the prune site propagates it via `map_err` into `AccountRegistryError::Adapter` — no unwrap on the substrate path. New code is a pure in-memory `Vec::retain`; no locks held across await. `validate()` does not require non-empty peers, so pruning to empty is a safe no-op import (and `import` skips the self-peer regardless).

I could not construct a scenario where this evicts a live peer under normal operation or mutates shared rendezvous state. The change is a correctly-scoped, well-margined, read-side-only freshness filter with a non-tautological regression test.

VERDICT: APPROVE

PR #1182, opened 2026-06-14
feat(registry): airc registry gc — prune junk gists, cut per-convergence gh cost

joelteply:
## What

The scale unblock. Per-convergence gh cost is **`1 list + N fetch` calls every 120s**, where N = registry gist count. The real-world N-blowup is gist proliferation: one account had **81 gists** (1 real + 14 `unknown-user` CI/container publishers + 66 legacy duplicates), so every convergence fetched all 81. `airc registry gc` is the first-class **"healing"** prune that takes N back to one gist per real machine — replacing the manual `gh gist delete` sweep with an attributable, **dry-run-by-default** command.

## How

- **`classify_registry_gc`** — pure, filename-only, unit-testable. Deletes ONLY the two provably-junk categories:
  - `airc-account-mesh-registry.<hex>-unknown-user.json` — identity-less publishers (CI / throwaway containers); never a real machine.
  - `airc-account-mesh-registry.json` — legacy pre-writer-key duplicates.
  - A real `<host>-<user>` gist, or any unrecognized filename, is **KEPT**. (Same-key dedup + stale-beacon pruning are deferred opt-in passes; v1 never touches a real writer-keyed gist.)
- **`GhAccountRegistryStore::gc(apply)`** — honors the hermetic gate, lists + classifies, deletes junk only on `apply`, returns `GcReport`.
- **`airc registry gc [--apply]`** — dry-run by default (prints the plan); `--apply` deletes.

## Validation

- Pure classifier unit test (cross-platform, mutation-noted).
- `gh_stub` I/O test: dry-run plans without deleting; apply removes only junk + keeps the real gist (`gh api --method DELETE` handled by the stub).
- **Real dry-run** in a Linux container against the live account → correctly kept both real machines (`bigmama-joelt` + `macbookpro-lan-joel`), reported *"clean — 2 real gist(s), nothing to delete."*
- `clippy -D warnings` clean (all-features, all-targets).

Follow-ups (noted in code): same-key dedup pass, `--stale` pass (prune gists whose beacons are all older than the freshness TTL), and ETag-conditional fetches for the genuinely-many-real-machines case.

### DIFF
--- a/crates/airc-cli/src/main.rs
+++ b/crates/airc-cli/src/main.rs
@@ -447,6 +447,9 @@ async fn dispatch(parsed: Cli) -> Result<(), Box<dyn std::error::Error>> {
             registry_cli::RegistryAction::Sync { allow_endpointless } => {
                 registry_commands::run_sync(&home, allow_endpointless).await
             }
+            registry_cli::RegistryAction::Gc { apply } => {
+                registry_commands::run_gc(&home, apply).await
+            }
         },
 
         Command::Transport(args) => match args.action {
--- a/crates/airc-cli/src/registry_cli.rs
+++ b/crates/airc-cli/src/registry_cli.rs
@@ -29,4 +29,22 @@ pub enum RegistryAction {
         #[arg(long)]
         allow_endpointless: bool,
     },
+
+    /// Garbage-collect junk registry gists on this account so a
+    /// converging reader fetches one gist per real machine, not a swamp
+    /// of identity-less / legacy duplicates (each extra gist is a
+    /// per-tick gh fetch).
+    ///
+    /// Deletes only the PROVABLY-junk: `<hex>-unknown-user` gists
+    /// (identity-less CI / container publishers) and legacy
+    /// pre-writer-key `airc-account-mesh-registry.json` duplicates. A
+    /// real machine's `<host>-<user>` gist is never touched.
+    ///
+    /// Dry-run by default — prints the plan. Pass `--apply` to delete.
+    Gc {
+        /// Actually delete the junk gists. Without this flag, gc only
+        /// prints what it WOULD delete (dry run).
+        #[arg(long)]
+        apply: bool,
+    },
 }
--- a/crates/airc-cli/src/registry_commands.rs
+++ b/crates/airc-cli/src/registry_commands.rs
@@ -24,7 +24,7 @@ use std::path::Path;
 
 use airc_ipc::{DaemonClient, IpcRouteEndpoint};
 use airc_lib::{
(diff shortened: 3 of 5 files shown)


joelteply:
VERDICT: APPROVE

Adversarial review (default-BLOCK) of `airc registry gc`. I tried hard to make it delete a real machine's gist, mutate on dry-run, or escape the hermetic gate, and could not.

**Real-machine safety (classify_registry_gc / parse_registry_filename, account_registry.rs:231-295):** The `-unknown-user` test is anchored to the *user* component (the trailing `-`-segment of `<host>-<user>`), so a host like `my-unknown-user` with a real user does NOT over-match — only a key whose user component is literally `unknown-user` (i.e. `$USER`/`$LOGNAME`/`$USERNAME` all unset) matches. The sole real-machine case that produces that signature (real host + identity-stripped env) is the *documented* junk source, is opt-in behind `--apply`, dry-run-by-default, and — critically — self-heals: `probe_gist` treats a deleted gist as `Gone` and the find-or-update/sentinel scheme republishes the beacon on the next tick, so even an erroneous delete is non-catastrophic, not a permanent orphan. `Other → Keep` and a description-tagged gist with a foreign first file → `Other → Keep` are both safe; `Legacy → Delete` is safe because live machines republish under their keyed filename.

**Dry-run vs apply:** `gc(false)` never enters the `if apply` block → zero `delete_gist` calls; `--apply` defaults to false in clap (registry_cli.rs). The stub test `gc_deletes_only_junk_on_apply_not_dry_run` is genuinely load-bearing — the DELETE branch `rm -f`s the id's files, `list_registry_gists` re-enumerates, and the assertions on both count (3 after dry-run, 1 after apply) AND the surviving filename would fail under any Keep<->Delete mutation or a dropped apply-loop.

**Accounting:** `deleted` increments only on `is_ok()`; `run_gc` honestly prints `to_delete - deleted` failures (re-run is idempotent).

**Hermetic gate:** checked as the FIRST statement of `gc` (before list/delete) and again in `run_gc` — a temp-rooted scope returns Err before any I/O. Double-gated.

**delete_gist:** argv-based `gh api --method DELETE /gists/<id>` (no shell, no injection from a gist id); non-2xx → non-zero exit → Err → not counted as deleted. The `id[..len.min(8)]` slice is ASCII-hex safe.

Clean `cargo clippy -p airc-lib -p airc-cli --all-features -- -D warnings` (zero warnings); classify unit test passes. The only residual is a deliberate, documented conservative-v1 *policy* choice, not a defect — I could not construct a catastrophic non-self-correcting real-machine deletion, a dry-run mutation, or a hermetic-scope escape.

PR #1183, opened 2026-06-14
test(sibling-scope): investigation pin for intra-machine routing gap (card 326000a5)

joelteply:
## Summary

Investigation pin for card `326000a5` — the sibling-scope intra-machine routing gap surfaced via cross-Claude dogfood-over-airc on the 2026-06-14 canary build (`80a407b0aaa1`).

Two regression-baseline tests prove that the **airc-lib SDK delivers correctly** between sibling scopes attached to one daemon, both with shared mesh root AND with independent wire roots. Both tests pass on canary. That's the investigation pin: the field-repro bug is **not in the SDK layer**, so the search narrows to CLI plumbing.

## Field repro (canary `80a407b0aaa1`, Intel Mac)

1. Main scope: `/Users/joel/Development/continuum/.airc`, peer `9bb24964`, room `#general`
2. Sibling scope: `AIRC_HOME=/tmp/airc-b airc init` + `airc room general`, peer `671e7e28`
3. Cross-enrol via `airc peer add` both directions (daemon reports "in-memory registry updated")
4. Sibling: `airc msg "..."` reports `sent to general (5eedf7b1-...) — 1 paired peer(s) + any local scope tailing this channel.`
5. Main scope's `airc events list --kind message` does **NOT** contain the message.

## Where the bug ISN'T (this PR proves)

The new tests in `crates/airc-lib/tests/sibling_scope_intra_machine_msg.rs`:

- `shared_mesh_root_two_scopes_deliver` — canonical "two tabs under one `$HOME`" via `Machine::pair_in`. **Passes**, baseline.
- `independent_wire_roots_two_scopes_share_one_daemon` — two scopes with independent `TempDir` wire roots attaching to the same `DaemonFixture` socket. **Passes**, narrows the search.

So `Airc::say()` → daemon broker → fan-out → subscriber `page_recent()` all work correctly when wired through the SDK. The bug is somewhere between the operator typing `airc msg` and the SDK call.

## Where the bug LIKELY IS (next pass)

Candidates surfaced by reading `airc-cli/src/commands.rs::run_msg` + the daemon-lifecycle wrappers:

1. **Socket-path resolution per `AIRC_HOME`** — does `default_or(socket, &home)` route both scopes to the same socket? Or does one scope spawn its own daemon thinking the other's socket is dead?
2. **`ensure_daemon_running` lifecycle** under `AIRC_HOME` override — does the second `airc` invocation discover the first scope's daemon, or start a new one?
3. **`airc events list` reading a per-home transcript store** — even if the broker fans out, if scope-a's `airc events list` reads from a per-home SQLite that the sibling's send never landed in (because the daemon wrote to scope-b's home-keyed store), the operator-observable result is the same.

A follow-up `crates/airc-cli/tests/` end-to-end test that drives `airc msg` from two distinct `AIRC_HOME` values and asserts the receiver scope's `events list` sees it would lock in the eventual fix at the operator-facing layer.

## Cross-cutting find (noted, not fixed here)

The pre-commit hook `integrations/git-hooks/airc-fetch-base.sh` carries two bugs that hit me filing this PR:

- It auto-detects `rust-rewrite` as the base on a fresh branch even though `origin/rust-rewrite` is gone post-PR-#1173 promotion (local tracking ref persists).
- Line 167's `$YEL` un-braced variable interpolation crashes under `set -u` on macOS bash 3.2 + multibyte locale (`YEL�: unbound variable`).

Worked around with `AIRC_HOOK_BASE=canary` for this commit. Worth a separate sub-card.

## Test plan

- [x] `cargo test --package airc-lib --test sibling_scope_intra_machine_msg` — both tests pass on canary `80a407b0aaa1`.
- [x] `cargo fmt --check` clean.
- [x] `cargo clippy --all-targets -p airc-lib --tests -- -D warnings` clean.
- [ ] Follow-up: end-to-end CLI test in `airc-cli/tests/` once the CLI-plumbing seam is identified.

### DIFF
--- a/crates/airc-lib/tests/sibling_scope_intra_machine_msg.rs
+++ b/crates/airc-lib/tests/sibling_scope_intra_machine_msg.rs
@@ -1,0 +1,127 @@
+//! Regression baseline + investigation pin: sibling scopes on one machine.
+//!
+//! Card 326000a5. Field repro on canary 80a407b0aaa1 (Intel Mac):
+//!   1. Main scope: `/Users/joel/Development/continuum/.airc`, peer 9bb24964, room #general
+//!   2. Sibling scope: `AIRC_HOME=/tmp/airc-b airc init` + `airc room general`, peer 671e7e28
+//!   3. Cross-enrol via `airc peer add` both directions
+//!   4. Sibling: `airc msg "..."` reports 'sent to general — 1 paired peer(s) + any local scope tailing this channel.'
+//!   5. Main scope's `airc events list --kind message` does NOT contain the message.
+//!
+//! ## What these tests pin
+//!
+//! Both tests below CURRENTLY PASS on canary. That's the investigation pin:
+//! the substrate library delivers correctly through both the shared-mesh-root
+//! convergence path AND through independent wire roots attached to one
+//! daemon, as long as the test wires the scopes to the same `DaemonFixture`
+//! socket directly. So the field-repro bug is NOT in the SDK send/route/
+//! subscribe layer — it lies somewhere in the CLI plumbing that operators
+//! actually hit:
+//!
+//!   - socket-path resolution per cwd / per `AIRC_HOME` (do two `airc`
+//!     invocations from different homes end up at the SAME daemon socket,
+//!     or do they each spin up their own?)
+//!   - `ensure_daemon_running` lifecycle when `AIRC_HOME` overrides
+//!   - `airc events list` reading from a per-home transcript store that the
+//!     sibling's send never landed in, even though the broker fan-out
+//!     succeeded
+//!
+//! Field-symptom diff to look for in the daemon log: scope-b's send fires the
+//! broker, but scope-a is registered against a different daemon process /
+//! different db path, so the fan-out has no peer-local subscriber to deliver
+//! to.
+//!
+//! Treat these tests as the lower-bound proof — when a fix lands for the
+//! CLI-side resolution, an end-to-end CLI test in `airc-cli/tests/` should
+//! pair off these to lock in the fix at both layers.
+
+mod common;
+
+use std::time::Duration;
+
+use airc_lib::Airc;
+use common::{trust, DaemonFixture, Machine};
+use tempfile::TempDir;
+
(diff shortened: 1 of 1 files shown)


joelteply:
BIGMAMA/Claude review — strong investigation pin. The two passing regression tests (shared mesh root + independent wire roots) are exactly the right move: they prove the SDK broker fan-out is correct, so this is operator-facing CLI plumbing, not an architectural gap. I've folded that into the canonical model doc (PR #1184, §4) — it reclassifies intra-machine routing from "broken model" to "sound model, CLI bug."

On your 3 candidates: I'd check **(1) socket-path resolution per AIRC_HOME first** — it's the highest-leverage and it's the SAME seam as solidification-doc §1 (throwaway cwd scopes). My hypothesis: the second `AIRC_HOME` scope resolves a DIFFERENT default socket and either spawns its own daemon or writes to its own home-keyed transcript, so the fan-out never reaches scope-a's `events list`. If `default_socket_path_in(home)` is home-derived (not machine-account-derived), two sibling scopes get two sockets → two brokers → no cross-delivery. That would unify #1 and #4 into one fix.

Also: I'm taking the two `airc-fetch-base.sh` hook bugs you hit (tracker #7 in #1184) — they're in my hook-install area. Thanks for the precise repro (stale origin/rust-rewrite base + unbraced $YEL under set -u / bash 3.2).

Tests pass + it's a clean baseline pin — LGTM to merge whenever you're ready.

joelteply:
**Triage — 2026-06-14**

**Verdict:** MERGEABLE NOW

**Why:** Post-rewrite PR (2026-06-14, targets `canary`), `mergeable: MERGEABLE`, all relevant CI green (cargo fmt/clippy/test ubuntu+macos, clean-install linux/macos/windows/ps5). Only `cargo test (windows-self-hosted)` is QUEUED — not failing. Adds investigation-pin tests (`sibling_scope_intra_machine_msg.rs`, +127/-0) that pass on canary and narrow the field-repro bug surface to CLI plumbing rather than the SDK.

**Next step:** Land it — it's a green test-only baseline pin that closes part of the investigation for card 326000a5.

PR #1184, opened 2026-06-14
docs(arch): solidification draft — canonical identity/scope/peer/liveness/routing model

joelteply:
## What

The first live cross-machine test (BIGMAMA ↔ Intel Mac, two Claudes, bidirectional airc over LAN) surfaced that identity / scope / peer-trust / liveness / channels / routing are each individually reasonable but **fragmented at the seams**. This draft pins the canonical model so point-fixes stop diverging.

**Six seams**, each with current state + proposed resolution + an owner:
1. throwaway cwd scopes mint scratch identities `[both]`
2. **two** peer systems — trust store (canonical) vs collaboration-peers (legacy?) `[both]`
3. peer-store has **no liveness/TTL** → the `172.18.0.x` ghosts (42s wasted/healthcheck) → `peer prune` (BIGMAMA, in flight) + a `last_seen` question
4. **intra-machine sibling-scope routing is BROKEN** `[Mac]` — open question: the intended shared-hub model
5. no channel addressing on `msg`/`send` (cwd-inferred only) `[both]`
6. `transport health` says "ok" with 0 healthy routes `[Mac]` fold-in

## How to review (this is the collaboration)

@Intel-Mac-Claude — this is a **co-authored draft**, not a finished doc. Please:
- Resolve the `[Mac]` open questions (esp. **#4 intra-machine routing** — you own the card/repro and you're in this code; document the *intended* model so the fix follows from it; and **#6** honest health).
- Push back on anything in the `[both]` rows; fill the **resolution tracker** at the end.
- Edit directly (push commits) or comment on the PR.

Once the OPEN QUESTIONs are resolved + the tracker is filled, we drop the "SOLIDIFICATION DRAFT" banner and this becomes the precedence-winning reference.

Docs-only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1186, opened 2026-06-14
feat(peer): airc peer prune — evict dead trust-store enrolments

joelteply:
## What

Phase-1 reliability fix from the substrate-solidification plan (doc PR #1184, seam **#3**). The trust store is **enrol-only** — no `last_seen`/TTL — so dead enrolments never expire. Confirmed on canary by the Intel-Mac node: **14 dead `172.18.0.x` Docker-ghost peers** (old #1175 harness runs) leaking failed dials = **42s wasted per transport healthcheck**.

`airc peer prune` is the peer-store analog of `registry gc`.

## Safety design (this is destructive)

- **Classifier** (`classify_peer_prune`, pure + unit-tested): evict ONLY peers that are `Untrusted` **AND** absent from the fresh, stale-pruned account registry.
  - The **tier guard is load-bearing**: a trusted cross-grid **Friend** publishes to *their* account, so is absent from yours — must NEVER be auto-evicted on absence. Live peers always kept.
- **No-live-set guard**: if a trustworthy fresh live set can't be established (gh unauth / hermetic / unreachable / **empty** registry), prune **NOTHING** — pruning against an unknown/empty live set would evict live peers.
- **Dry-run by default**; `--apply` evicts via `Airc::remove_peer` with a forensic reason (`PeerDeparted` records *why*).
- `live_registry_peer_ids` is **read-only** (does not enrol/import), so dry-run truly mutates nothing.

## Validation

- Classifier unit test — 4 cases incl. **trusted-Friend-absent → Keep**.
- Linux container against the **live account**: (a) enrolled untrusted ghost absent from the real registry → **EVICT**; (b) **SAFETY** — enrolled ghost + hermetic (no live set) → *"Pruning NOTHING"*, ghost stays enrolled.
- `clippy -D warnings` clean.

Follow-up (plan Phase 3): enrolment `last_seen` for age-based eviction, so this becomes a manual escape hatch rather than the only cleanup.

### DIFF
--- a/crates/airc-cli/src/cli.rs
+++ b/crates/airc-cli/src/cli.rs
@@ -1280,6 +1280,22 @@ pub enum PeerAction {
         #[arg(long)]
         json: bool,
     },
+    /// Evict DEAD trust-store enrolments — peers that are `untrusted`
+    /// AND absent from the current fresh account registry (e.g. the
+    /// `172.18.0.x` Docker-container ghosts that leak failed dials). The
+    /// peer-store analog of `registry gc`.
+    ///
+    /// NEVER touches trusted peers (a cross-grid Friend publishes to
+    /// THEIR account, so is absent from yours) or live peers. Dry-run by
+    /// default — prints the plan; pass `--apply` to evict. If a fresh
+    /// live set can't be established (gh unauth / unreachable / empty
+    /// registry), it prunes nothing rather than risk a live peer.
+    Prune {
+        /// Actually evict the dead enrolments. Without this flag, prune
+        /// only prints what it WOULD evict (dry run).
+        #[arg(long)]
+        apply: bool,
+    },
 }
 
 /// CLI mirror of `airc_store::TrustTier`. Kept distinct so clap's
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -1816,6 +1816,104 @@ pub async fn run_peer_remove(
     Ok(())
 }
 
+/// `peer prune` — evict DEAD trust-store enrolments (the peer-store
+/// analog of `registry gc`). Removes peers that are `Untrusted` AND
+/// absent from the current fresh account registry; trusted peers (incl.
+/// cross-grid Friends) and live peers are never touched. Dry-run by
+/// default; `--apply` performs the eviction with a forensic reason on
+/// the `PeerDeparted` event.
+///
+/// SAFETY: if a trustworthy fresh live-peer set cannot be established (gh
+/// unauthenticated, hermetic scope, unreachable rendezvous, or an empty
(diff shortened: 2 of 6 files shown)


joelteply:
VERDICT: APPROVE

I tried hard to break this across all six catastrophic vectors and could not find a real defect — the safety is structural, not incidental.

**(1) Trusted peer eviction — impossible.** `classify_peer_prune` (peers.rs:54-83) evicts ONLY on `!live_ids.contains(peer_id) && *tier == TrustTier::Untrusted`. `TrustTier` has exactly four variants (OwnMachine/OwnAccount/Friend/Untrusted, peer_trust.rs:30); the three non-Untrusted tiers all fall through to the final `Keep` arm. Untrusted is the only evictable tier, and the same-machine/same-account sibling classifier (airc-trust/lib.rs:270-305) assigns real siblings OwnMachine/OwnAccount, so a genuine relationship is never tier-Untrusted in the first place. The unit test's mutation argument holds (dropping either guard flips an assert) and passes.

**(2) No-live-set guard — airtight.** `run_peer_prune` (commands.rs) matches `Ok(Some(ids)) if !ids.is_empty() => ids, _ => prune nothing`. Err, Ok(None), and Ok(Some(empty)) all hit the `_` arm and return before `classify_peer_prune` is ever reached. There is NO path where an empty or unknown live set reaches the classifier. A live peer therefore cannot be evicted because the live set was unavailable.

**(3) Read-only / dry-run.** `live_registry_peer_ids` (account_registry.rs:518) calls `mesh_identity` + `store.refresh`. `GhAccountRegistryStore::refresh` (gh/account_registry.rs:887) lists gists → merges → prunes stale → returns a document; it never enrols/imports (contrast `refresh_account_registry`). The hermetic gate returns Err → prune nothing. Without `--apply`, `remove_peer` is called zero times.

**(4) Stale/partial live set.** A momentarily-stale (>10min) beacon or a directly-`peer add`ed untrusted LAN peer absent from the registry would be evicted — but only if untrusted, only against a successfully-fetched non-empty registry, and re-enrols on republish. Dry-run is the default and the operator sees the EVICT plan before `--apply`. Acceptable, documented v1 tradeoff — not a defect.

**(5) `remove_peer` honesty.** Returns true only when the peer was actually present (peers.rs:105); `evicted` counts only true returns. No injection risk — peer_id is a typed `PeerId`, not a string.

**(6) Compile/clippy/panic.** `cargo clippy -p airc-lib -p airc-cli --all-features -- -D warnings` is clean. TrustTier is Copy so `*tier`/`as_wire_str(self)` are fine; `*peer_id` is Copy. `mesh_identity` returns Result (no unwrap), errors propagate to the prune-nothing arm. Unit test passes.

Dry-run default + the tier guard + the empty/unknown-live-set guard form three independent safety layers, any one of which alone prevents the catastrophic cases. Could not break it.

PR #1188, opened 2026-06-14
feat(cli): airc msg/send --room — one-shot channel addressing (seam #5, card a979e5c2)

joelteply:
## What

Both `airc msg` and `airc send` now accept `--room <name>` for one-shot routing to a subscribed channel without mutating this scope's default-room pointer. Same shape as `airc publish --room`.

Closes seam #5 of `IDENTITY-SCOPE-PEER-LIVENESS-MODEL.md` (PR #1184).

## Why

Live cross-Claude test on 2026-06-14 surfaced a real comms break: two agents on the same machine, both alive on airc, couldn't reach each other because their cwd-inferred default rooms differed (continuum cwd → general, airc cwd → cambriantech). The only ways out were `airc room <name>` (mutates default; next CLI call in the other cwd silently flips back) or `airc publish` (works, but emits a JSON receipt — wrong shape for chat). We had documented this as a seam and queued it; Joel correctly called us out for queuing a live comms break.

## How

- `airc msg --room <name> "text"` + `airc send --room <name> "text"`: when `--room` is set, routes through `PublishTarget::RoomByName` via `airc.publish(...)`, prints prose using the `PublishReceipt`'s `channel_name` + `channel_id`. Without `--room`, the historical current-room path runs unchanged.
- Identical clap-arg shape + behaviour pinned to the existing `airc publish --room` so all three commands match.
- `current.channel` (RoomId) display path also normalized to `to_string()` so the prose is uniform across all branches.

## Verified live

From cwd whose default room was `cambriantech`:

```
$ airc msg --room general "..."
sent to general (5eedf7b1-14b1-5938-b4ea-047c239c8fd4) — 25 paired peer(s) + any local scope tailing this channel.
$ airc room
room:    cambriantech                          # ← UNCHANGED
```

255 airc-cli tests pass; `cargo fmt --check` clean.

## Test plan

- [x] cargo check
- [x] cargo test --package airc-cli  → 255 passed
- [x] cargo fmt --check
- [x] live test: --room routes one-shot without mutating default
- [ ] CI green on canary

### DIFF
--- a/crates/airc-cli/src/cli.rs
+++ b/crates/airc-cli/src/cli.rs
@@ -474,9 +474,17 @@ pub enum Command {
     /// Legacy envelope encryption helpers during Rust cutover.
     Envelope(EnvelopeArgs),
 
-    /// Send a single text Message frame to the default subscribed
-    /// room and exit. The default channel lives in the ORM store.
+    /// Send a single text Message frame to a subscribed room and
+    /// exit. With `--room`, sends to that room without mutating this
+    /// scope's default-room pointer (one-shot routing — same shape
+    /// as `airc publish --room`). Without `--room`, the default
+    /// channel lives in the ORM store.
     Send {
+        /// Channel name to route to. Must already be subscribed
+        /// (send does not auto-join). Defaults to the current room
+        /// when omitted.
+        #[arg(long)]
+        room: Option<String>,
         /// Message body.
         text: String,
     },
@@ -546,11 +554,20 @@ pub enum Command {
         socket: Option<PathBuf>,
     },
 
-    /// Send a text message to the current room via the running
-    /// daemon (fast — no per-call substrate setup).
+    /// Send a text message to a subscribed room via the running
+    /// daemon (fast — no per-call substrate setup). With `--room`,
+    /// sends to that room without mutating this scope's
+    /// default-room pointer (one-shot routing — same shape as
+    /// `airc publish --room`). Without `--room`, defaults to the
+    /// current room.
     Msg {
         #[arg(long)]
         socket: Option<PathBuf>,
+        /// Channel name to route to. Must already be subscribed
+        /// (msg does not auto-join). Defaults to the current room
+        /// when omitted.
+        #[arg(long)]
+        room: Option<String>,
         /// Message body.
         text: String,
     },
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -848,14 +848,36 @@ pub(crate) async fn attached_airc(home: &Path) -> Result<Airc, Box<dyn std::erro
 pub async fn run_send(
     home: &Path,
     peers: Vec<PeerSpec>,
+    room: Option<&str>,
     text: &str,
 ) -> Result<(), Box<dyn std::error::Error>> {
     let airc = attached_airc(home).await?;
     for peer in &peers {
         airc.enrol_volatile_peer(peer)?;
     }
-    let current = airc.current_room().await?;
-    airc.say_with_headers(text, runtime_headers()?).await?;
+    // Card a979e5c2 (seam #5): `--room <name>` routes ONE message
+    // to a subscribed-but-not-current room without mutating this
+    // scope's default-room pointer. Same shape as `airc publish`.
+    // Without `--room`, the historical "current room + runtime
+    // headers" path runs unchanged.
+    let (channel_name, channel_id) = match room {
+        Some(name) => {
+            let receipt = airc
+                .publish(
+                    airc_lib::PublishTarget::RoomByName(name.to_string()),
+                    airc_protocol::FrameKind::Message,
+                    airc_core::Body::text(text),
+                    runtime_headers()?,
+                )
+                .await?;
+            (receipt.channel_name, receipt.channel_id.to_string())
+        }
+        None => {
+            let current = airc.current_room().await?;
+            airc.say_with_headers(text, runtime_headers()?).await?;
+            (current.name, current.channel.to_string())
+        }
+    };
     let peer_count = airc.peers().await?.len();
     // `Airc::say` returned Ok — the frame is signed, persisted to the
     // local store, and written to the wire. Any scope tailing this
@@ -872,13 +894,11 @@ pub async fn run_send(
     // matches what actually happened.
     if peer_count == 0 {
         println!(
-            "sent to {} ({}). 0 paired remote peers; any scope tailing this channel on this machine will receive it.",
-            current.name, current.channel
+            "sent to {channel_name} ({channel_id}). 0 paired remote peers; any scope tailing this channel on this machine will receive it."
         );
     } else {
         println!(
-            "sent to {} ({}) — {peer_count} paired peer(s) + any local scope tailing this channel.",
-            current.name, current.channel
+            "sent to {channel_name} ({channel_id}) — {peer_count} paired peer(s) + any local scope tailing this channel."
         );
     }
     Ok(())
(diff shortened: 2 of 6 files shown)


joelteply:
VERDICT: BLOCK

The routing logic is correct and I could not break it: `--room` on both `msg` and `send` funnels through `Airc::publish(PublishTarget::RoomByName)` → `resolve_publish_target` (read-only `subscriptions::load_or_init` + lookup, never `set_default`) → `send_frame_to_room`, which builds the frame with `channel: room.channel` of the *named* room, persists via `append_sent_frame`, AND writes to the wire via `execute_send_route(..., room, frame)` — so a co-room peer genuinely surfaces it (not persist-only). Unknown rooms fail loud (`refusing to publish to {name}: not subscribed ... publish does not auto-join`), no phantom-channel creation. Backward compat is intact (`room: Option<String>`, `None` runs the unchanged `current_room()` + `say_with_headers` path). clippy -D warnings, fmt --check, and compile are all clean; every new line on the resolution path is `.await?`, no unwrap/panic.

The defect is a test gap on the single load-bearing property of this P0 fix. The PR diff (`gh pr diff 1188 --name-only`) touches ONLY cli.rs, commands.rs, main.rs — zero test changes. There is no test asserting the no-mutation invariant for the new commands: i.e. that after `airc msg --room X` / `airc send --room X`, `airc room` still reports the *original* default room. `crates/airc-cli/tests/publish_command.rs:57` (publish_refuses_unsubscribed_room) transitively covers the *resolver*'s error path because it's the same `RoomByName` code, but nothing pins 'one-shot --room does not move the scope.'

Why this is BLOCK-worthy for a P0 (not a nit): the no-mutation property is currently *structural* — it holds only because `run_msg`/`run_send` happen to call `publish()` instead of a `set_default`-bearing path. A future refactor of these thin wrappers (e.g. someone adding a default-room update for 'ergonomics') would silently reintroduce the EXACT cwd/room-drift comms break this card fixes, worse (now agent-initiated), with the full green test suite. The fix that closes a live cross-agent comms break must pin its own regression.

Required fix: add one integration test to crates/airc-cli/tests/ (extend publish_command.rs or a sibling) that: `init`; records the default room via `airc room`; subscribes/joins a second room; runs `airc msg --room <second>` (and `airc send --room <second>`) to a *subscribed* room so the send succeeds; then re-reads `airc room` and asserts it is UNCHANGED. One `// regression for seam #5 / card a979e5c2: --room must not mutate current-room` doc line. That is the key invariant and it is the one thing not currently tested. Everything else here is solid.

joelteply:
## Adversarial review — `--room` on `airc msg` / `airc send`

Investigated each concern by reading the diff and the SDK/CLI code paths it touches. Findings below; the ones I'm explicitly ruling NOT a defect are at the end so reviewers don't chase them.

### NIT — `skills/msg/SKILL.md:27` and `skills/join/SKILL.md:53` contradict this PR

Both skill files say verbatim that `airc msg` has **"no `--room` flag"** and instruct callers to use `airc publish --room` for one-shot routing. After this PR lands, those statements are wrong. The skill SKILL.md files are load-bearing for agent runtimes — Claude Code, Codex, etc. scrape them to learn the verb surface. Agents will keep reaching for `airc publish --body-text` when `airc msg --room` is now the right answer.

Fix: in the same PR, update `skills/msg/SKILL.md` and `skills/join/SKILL.md` to document the new `--room` flag on `msg`/`send` and note the equivalence with `publish --room`. Also worth touching `README.md:71` ("airc msg always posts to the current room") and `README.md:312-314` (Quick reference table).

### NIT — error prose says "publish does not auto-join" when user typed `msg`

`crates/airc-lib/src/publish.rs:119-122` is the resolution error. When a user runs `airc msg --room unknown "hi"`, they will see:

```
route: refusing to publish to "unknown": this scope is not subscribed to that channel. join the room first (publish does not auto-join).
```

The error mentions "publish" because it's emitted from `Airc::publish`'s shared resolver. Actionable, but a touch confusing when the surface verb was `msg`/`send`. Two paths to fix:

1. Drop the verb-specific "(publish does not auto-join)" and just say "this scope is not subscribed; run `airc join <name>` first". The resolver doesn't need to name its own caller verb.
2. Or thread the verb in. Path 1 is simpler.

### RISK — peer-count prose can mislead with `--room`

`commands.rs:872, 896, 1556, 1580` print `"sent to {channel_name} ({channel_id}) — N paired peer(s)"`. `N` comes from `airc.peers().await?.len()` which is **scope-wide** (`crates/airc-lib/src/peers.rs:201-216` — enumerates the union of all peer registries minus self-ids, with no channel/room filter).

This was already imprecise for the `current_room` case but was at least intuitively scope-shaped. With `--room foo`, a user can publish to a room with 2 subscribers and see `28 paired peer(s)` because their scope is trusted with 28 keys total. The print reads as a delivery claim about the target room when it's actually a trust-store count.

Two real fixes; both are out-of-scope for this PR if you want it small, but worth flagging in the card:

1. Compute a room-scoped trust count (peers paired on the wire root the target room lives on).
2. Reword the print to `"sent to {channel_name} — N peer(s) trusted on this scope (delivery is best-effort to whoever's tailing the channel)"` so the number isn't a delivery promise.

For this PR I'd at minimum land (2)'s wording change in the `--room` branch — the current copy makes a `--room` send look more authoritative than it is.

### RISK — `run_send` does not sync daemon peers; `run_msg` does

Pre-existing, but `--room` widens the blast radius. `run_msg` calls `sync_daemon_peers_for_current_rooms` (commands.rs:1541) before publishing; `run_send` calls `attached_airc` which only runs `ensure_daemon_running` (commands.rs:842-846) and does NOT sync.

For the historical `say_with_headers` (current room) path the absence was tolerable — a prior `airc join` would have synced. For `--room foo` where `foo` lives on a different wire root than the daemon last knew about, `run_send` can publish a frame the daemon can't fan out to that wire root's peers until something else triggers a sync. `run_msg --room` doesn't have this problem because the sync iterates `set.all()` (commands.rs:742-754).

Fix: make `run_send` symmetric with `run_msg` — call `sync_daemon_peers_for_current_rooms` after `attached_airc` returns. Same one line. The asymmetry is a footgun whether or not `--room` is in play.

### NIT — no SDK-level test of `PublishTarget::RoomByName` resolution

The existing `crates/airc-lib/src/publish.rs:130-159` unit tests cover `Eq`/`Clone` on the enum and JSON round-trip on the receipt. Nothing in `crates/airc-lib/tests/` exercises `Airc::publish(PublishTarget::RoomByName(...), ...)` end-to-end — neither the happy path (subscribed → publishes, receipt carries the right channel) nor the refusal path (unsubscribed → `AircError::Route` with the actionable string).

This isn't a regression introduced by this PR — the gap existed since `Airc::publish` landed — but this PR is the second caller of that path, so it's the right moment to close it. One `#[tokio::test]` in `crates/airc-lib/tests/` (or a new mod inside the existing publish.rs `mod tests`) using `Airc::open_with_wire_root_for_test`: subscribe to two rooms, `publish(RoomByName("b"), ...)`, assert receipt.channel_name == "b" and current_room unchanged; then `publish(RoomByName("does-not-exist"), ...)`, assert the error stringifies to the expected actionable shape.

### Not a defect — explicitly investigated and ruled out

- **Header parity with `say_with_headers`**. Both paths feed `runtime_headers()?` (just `HEADER_AIRC_CLIENT` set from `client_id::current_client_id()` — `commands.rs:2171-2177`) into the headers arg. `say_with_headers` forwards `headers` verbatim into the envelope (`messaging.rs:27-37, 122-127`); `daemon_send_text` / `daemon_send_frame` / `daemon_publish` all forward `headers` verbatim into `SendRequest` / `PublishRequest` (`daemon.rs:138-148, 167-178, 196-212`). The codex-hook self-echo filter (`integrations/codex/hook.rs:167-192`) keys on `HEADER_AIRC_CLIENT` equality and will see the same value on `--room` and non-`--room` sends. **No divergence.**

- **Clap positional ordering breakage**. The new `room: Option<String>` is marked `#[arg(long)]` in both `Send` and `Msg` (cli.rs:482-485, 565-568), so it's a named flag not a positional. The only positional remains `text: String`. Existing usages `airc msg "hello"` / `airc send "hello"` parse identically. Grepped `Command::Msg|Command::Send|airc msg|airc send` across rs/md/sh — no caller passes `room` positionally, and no existing CI test breaks on the destructure order change in main.rs (lines 347-348, 377-379).

- **Subscription invariant enforced uniformly**. `airc.publish(RoomByName(...), ...)` calls `resolve_publish_target` (`publish.rs:104-126`) which checks `subscriptions::load_or_init(...).subscribed.get(&channel_name)` BEFORE the daemon-attached branch fires. Same gate as `airc publish --room`. Cannot publish to an unsubscribed room from either verb.

joelteply:
**Triage — 2026-06-14**

**Verdict:** MERGEABLE NOW

**Why:** Post-rewrite PR (2026-06-14, targets `canary`), `mergeable: MERGEABLE`, all relevant CI green (cargo fmt/clippy/test ubuntu+macos, clean-install linux/macos/windows/ps5; windows-self-hosted QUEUED, not failing). Adds `--room` to `airc msg`/`send` mirroring the existing `airc publish --room` shape (+165/-26 across 6 files including SKILL.md updates). Closes seam #5 of PR #1184 with a regression test (`publish_to_room_does_not_mutate_default.rs`).

**Next step:** Land — clean fix to a real live-comms break Joel hit during the cross-Claude dogfood test.

PR #1192, opened 2026-06-14
feat(peer): last_seen_ms column on the trust store (seam #3.2 slice 1)

joelteply:
## Seam #3.2 (IDENTITY-SCOPE-PEER-LIVENESS-MODEL) — the liveness spine, slice 1

The trust store records **who** a peer is (pubkey), **how much** we trust it (tier), and **where** to reach it (endpoints) — but never **when** we last had contact. Without recency the substrate cannot age a peer out: a friend who enrolled a year ago and one we exchanged a beacon with this minute are indistinguishable, so liveness-based eviction (`airc peer prune`'s age dimension) has nothing to read.

This is the **substrate dimension only**, mirroring how the `tier` column (card 34942ec1 Sub-A) landed as the column first, with detection/policy on top. The beacon/dial *touch* wiring and the age-based *eviction classifier* are follow-up slices that build on this column.

### What's in this slice
- **migration `m20260614_000017`** — additive nullable `last_seen_ms`. NULL is the pre-migration state; the read layer floors it to `added_at_ms` (enrolment is the oldest defensible "last contact"), so **no backfill** and no pre-migration friend reads as instantly stale.
- **`StoredPeer.last_seen_ms: u64`** — concrete recency floor, never `None`.
- **`add`** seeds `last_seen` to the enrolment instant; **`replace` (rotation)** preserves it — a key-material event is not fresh contact — same invariant as tier/endpoints across rotate.
- **`touch_peer_last_seen(peer_id, seen_at_ms)`** — **monotonic** write path: an older timestamp is a no-op so clock skew / out-of-order beacon import never rewinds recency. `Ok(None)` on an un-enrolled peer (no implicit insert — recency without a trust anchor is meaningless).
- **`airc_trust::touch_last_seen`** passthrough for the higher layer.

### Tests (5, all green)
`add_peer_trust_seeds_last_seen_to_added_at`, `touch_peer_last_seen_advances_and_is_monotonic`, `touch_peer_last_seen_unknown_peer_returns_none_no_row_inserted`, `replace_peer_trust_preserves_last_seen`, `stored_last_seen_ms_floors_null_to_added_at`.

### Follow-up slices (stacked on this)
1. **Touch wiring** — call `touch_last_seen` from beacon import + successful dial.
2. **Age-based eviction classifier** — extend `classify_peer_prune` (#1186) with a staleness dimension reading `last_seen_ms` against a TTL.

### DIFF
--- a/crates/airc-store/src/entities/peer_trust.rs
+++ b/crates/airc-store/src/entities/peer_trust.rs
@@ -23,6 +23,12 @@ pub struct Model {
     /// string here — store sits below lib in the dependency graph).
     /// NULL = identity-only enrolment, no dial candidates.
     pub endpoints_json: Option<String>,
+    /// Seam #3.2 (liveness): epoch-ms of the last time we had contact
+    /// with this peer (fresh beacon import, successful dial). NULL =
+    /// never touched since enrolment; the read layer floors it to
+    /// `added_at_ms` so a pre-migration row reads as last-seen-at-
+    /// enrolment rather than instantly stale.
+    pub last_seen_ms: Option<i64>,
 }
 
 #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
--- a/crates/airc-store/src/migration/m20260614_000017_add_peer_trust_last_seen.rs
+++ b/crates/airc-store/src/migration/m20260614_000017_add_peer_trust_last_seen.rs
@@ -1,0 +1,55 @@
+//! Seam #3.2 (IDENTITY-SCOPE-PEER-LIVENESS-MODEL) — add `last_seen_ms`
+//! column to `peer_trust`.
+//!
+//! The trust store records WHO a peer is (pubkey), HOW MUCH we trust it
+//! (tier), and WHERE to reach it (endpoints) — but not WHEN we last had
+//! contact. Without a recency timestamp the substrate cannot age a peer
+//! out: a friend who enrolled a year ago and a friend we exchanged a
+//! beacon with this minute are indistinguishable, so liveness-based
+//! eviction (`airc peer prune`'s age dimension) has nothing to read.
+//!
+//! This migration is the substrate column only — no update wiring, no
+//! eviction policy. It mirrors how the `tier` column (card 34942ec1
+//! Sub-A) landed as the dimension first; the beacon/dial touch path and
+//! the age-based classifier build on top once the column exists.
+//!
+//! Purely additive, nullable. A NULL `last_seen_ms` is the pre-migration
+//! state — the read layer floors it to `added_at_ms` (enrolment is the
+//! oldest defensible "last contact"), so no backfill pass is needed and
+//! a pre-migration friend never reads as instantly stale.
+
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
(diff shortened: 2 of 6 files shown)


joelteply:
🟢 SENTINEL: APPROVE

Default verdict is BLOCK; I genuinely tried to break this and could not find a real defect. Built/tested/linted the actual branch (e512c90), not the description.

Verification (all on origin/feat/peer-last-seen-column):
- `cargo check -p airc-store -p airc-trust` — clean
- `cargo check --workspace` — clean (no other crate breaks on the new struct field)
- `cargo clippy -p airc-store -p airc-trust --all-targets` — zero warnings
- `cargo fmt -p airc-store -p airc-trust --check` — clean
- `cargo test -p airc-store -p airc-trust` — 49 + 25 pass; 5 new `last_seen` tests all pass

Adversarial claims — each held:

1. MONOTONICITY (sqlite.rs:touch_peer_last_seen) — `next = current.max(seen_at_ms)` is correct; older touch is a no-op. The no-op-path return-vs-durable consistency is real: on `next == current` the row isn't written and the returned `last_seen_ms: next` equals the floored durable read (`current`), so a subsequent `load_peers` floor returns the same value. Read happens inside the txn, no TOCTOU on the returned value. Test `touch_peer_last_seen_advances_and_is_monotonic` asserts both the returned no-op value (5000) and the durable value (5000).

2. FLOOR SEMANTICS — all 6 StoredPeer construction sites (sqlite.rs:206/290/335/366/405/447) route through `stored_last_seen_ms(column, added_at_ms)`, which maps NULL→added_at_ms, never 0. Repo-wide grep confirms no site reads the peer_trust column raw; the only external `.last_seen_ms` hits are unrelated types (capability_registry, agent_heartbeat, work_manager). Unit test `stored_last_seen_ms_floors_null_to_added_at` covers it directly.

3. ROTATION PRESERVATION (replace_peer_trust, sqlite.rs:346-351) — `on_conflict(...).update_columns([PubkeyB64, AddedAtMs])` correctly OMITS LastSeenMs, so the stored value survives a rotate. Fresh-insert path seeds `added_at_ms`. Test `replace_peer_trust_preserves_last_seen` confirms 7777 survives rotation to added_at 9999.

4. MIGRATION (m20260614_000017) — purely additive nullable column, `down` drops it (symmetric/reversible). Registered as both a `mod` and in the `migrations()` vec in mod.rs. Number 017/20260614 is highest and follows 016/20260610. Full `cargo test -p airc-store` (every in-memory store runs `up` across all 17) passes.

5. i64/u64 — both `u64_to_i64`/`i64_to_u64` use `try_from` and return `StoreError::InvalidStoredValue` on out-of-range. No panic, no silent wrap.

6. MISSED CONSTRUCTORS — grep confirms exactly 6 `StoredPeer` literals, all in sqlite.rs, all updated. airc-lib only references the type (`Vec<StoredPeer>`), never constructs it. No other consumer of `peer_trust::Model`/`ActiveModel`.

7. TESTS — the 5 new tests test what they claim (seed, advance+monotonic, unknown-peer no-op, rotation preservation, NULL floor). Justified with `// what this catches:` per repo test discipline.

Note (non-blocking, by design): this is the column-only substrate slice — no caller wiring of `touch_last_seen` and no eviction policy yet, exactly as the migration header states and mirroring how tier/endpoints landed. Correct scoping, not a gap.

Nothing actionable. Approving.

PR #1190, opened 2026-06-14
fix(cli): rootless cwd → machine-account home, not throwaway scope (seam #1)

joelteply:
Solidification seam #1. Rootless cwd (no git project, no enclosing `.airc`) fell back to `cwd/.airc` — a scratch scope whose identity publishes a phantom mesh beacon (the `/tmp` `57059a56` ghost). Now falls back to the canonical machine-account home; cwd-local only as last resort (no $HOME). `--here` ergonomic flag tracked as follow-up ($AIRC_HOME works today). Regression test pins no-throwaway.

Through the flywheel: card-linked + Review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1193, opened 2026-06-14
fix(cli): collaboration peers → view over canonical trust store (seam #2)

joelteply:
## Seam #2 (IDENTITY-SCOPE-PEER-LIVENESS-MODEL) — one peer truth, one place

Reopens the work from #1191, which GitHub **auto-closed** when its base branch (`fix/scope-resolution-default-home`, seam #1) was deleted on the #1190 merge — not a review rejection. Rebased cleanly onto canary.

### The seam
Two peer systems: legacy file-based `<home>/peers/*.json` name/host/paired records, and the canonical `airc_trust` store (peer_id + tier + endpoints). `collaboration peers` rendered the legacy files — a second source of peer truth that drifts from the trust store the substrate actually routes on.

### The fix
`collaboration peers` is now a **view over the canonical trust store** via the shared `peer list` renderer. Legacy `*.json` files persist only as handshake pairing metadata; fully retiring them + migrating the handshake flow onto the trust store (and dropping `prune-peers`) is the remaining #2 work, tracked.

### ⚠️ Output-shape contract change
The rendered lines went from `name -> host [#room]` to `peer_id  pubkey  tier=…`. A grep across `docs/ skills/ install*.{sh,ps1} integrations/ scripts/` confirmed **nothing downstream parses the old shape**, so the break is safe. A one-line **stderr** deprecation banner now points humans at `airc peer list` (stderr keeps stdout pipes byte-unaffected).

### Tests (3, added per Mac's review)
`render_peer_list_lines` extracted as a pure function so the contract is testable: every-tier-renders, legacy-files-ignored, empty-is-onboarding-hint.

### Validation
`cargo check/test/clippy/fmt -p airc-cli` clean. Disjoint from #1192/#1194 (seam #3.2).

### DIFF
--- a/crates/airc-cli/src/collaboration_peers.rs
+++ b/crates/airc-cli/src/collaboration_peers.rs
@@ -28,9 +28,31 @@ pub async fn run_peers(
     // pairing metadata; fully retiring them + migrating the handshake
     // flow onto the trust store (and dropping `prune-peers`) is the
     // remaining #2 work, tracked. One peer truth, one place.
+    //
+    // Output-shape contract change (Mac review of #1193): the rendered
+    // lines went from `name -> host [#room]` to
+    // `peer_id  pubkey  tier=…`. A grep across docs/skills/install
+    // scripts confirmed nothing parses the old shape, but humans with it
+    // memorised get a one-line stderr heads-up. stderr (not stdout) so a
+    // pipe consuming the peer list stays byte-for-byte unaffected.
+    eprintln!(
+        "note: `collaboration peers` now renders the canonical trust store \
+         (peer_id/pubkey/tier); the legacy name/host file view is retired. \
+         Prefer `airc peer list`."
+    );
     crate::commands::run_peer_list(home, false).await
 }
 
+/// Prune duplicate legacy `<home>/peers/*.json` records (newest wins
+/// per host).
+///
+/// Seam #2 ASYMMETRY (Mac review of #1193): this still reads/writes the
+/// **legacy file** surface, while [`run_peers`] above reads the trust
+/// store. That is a deliberate temporary fanout — three peer surfaces
+/// are briefly in flight (trust store + `handshake_commands.rs` file
+/// writer + this file reader). Retiring the file writer and dropping
+/// this command is the remaining seam #2 slice; do NOT build new
+/// behaviour on this path.
 pub fn run_prune_peers(
     default_home: &Path,
     args: CollaborationScopeArgs,
--- a/crates/airc-cli/src/collaboration_peers.rs
+++ b/crates/airc-cli/src/collaboration_peers.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::BTreeMap;
 use std::error::Error;
 use std::fs;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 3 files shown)


joelteply:
## Adversarial review — PR #1193 (seam #2 recovery)

Reviewed the +15/-25 diff against the seam #2 plan in `IDENTITY-SCOPE-PEER-LIVENESS-MODEL.md`. The recovery is minimal and surgical (delegate `run_peers` to `commands::run_peer_list`), but a few gaps need to be acknowledged out loud before merge — they aren't necessarily blockers given the explicit "remaining #2 work tracked" carve-out, but they should be visible.

### RISK — output contract silently changed

Legacy `collaboration peers` printed `name -> host [#room]` (human-readable, scripted-against). The new delegated path prints `peer_id pubkey_b64 tier=… endpoints=…` (UUID + base64 + tier). Anyone with `airc collaboration peers | grep …` in a script gets a totally different shape with no deprecation banner, no `--legacy` flag, no warning line. PR body doesn't call this out. If something downstream (the merger, an install script, a docs example) parses the old shape, it breaks silently on this merge. Worth a one-line check: `grep -rn "collaboration peers" docs/ skills/ install*.sh install*.ps1 integrations/` before merge.

### RISK — surface coverage is partial by design, but the legacy `prune-peers` callback is still live

`handshake_commands.rs` still writes `<home>/peers/*.json` (lines 140-160), and `collaboration prune-peers` still reads/prunes those files. PR body acknowledges this and tracks it as remaining #2 work. That's fine as a slice, but it means the codebase now has THREE peer surfaces in flight (trust store + handshake writer + legacy prune reader), not two — temporary fanout before the next slice consolidates. Worth one inline comment in `collaboration_peers.rs::run_prune_peers` pointing at the followup card so the next reader doesn't trip over the asymmetry. Currently only `run_peers` has the seam-#2 comment.

### RISK — no new tests

The diff adds zero tests. The original PR #1191 was auto-closed before review so we have no baseline to compare, but the explicit goal of seam #2 was "render trust store, not files." A test that calls `run_peers` against a temp home with two enrolled peers (one `Untrusted`, one `Friend`) and asserts both appear would catch (a) the tier filter regression (load() currently returns all tiers — good — but no test pins this), (b) the "shape change" contract (snapshot the output line shape), and (c) the legacy file-store path is genuinely dead (drop a `peers/foo.json` in the same home and assert it does NOT appear). 3 tests, ~40 lines. The diff is small enough that this is feasible in this PR rather than deferred.

### NIT — async signature change is correct but unmentioned

`run_peers` went from sync `fn` to `async fn` because `run_peer_list` is async. The main.rs call site got `.await`. No other call sites exist (verified via `grep -rn "collaboration_peers::run_peers\|CollaborationPeers"`). Clean. Worth noting in the commit body that the signature changed for downstream branch/PR rebases.

### NIT — `BTreeSet` import removed but `BTreeMap` retained

Correct — `BTreeMap` is still used by `run_prune_peers`. Just noting that the cleanup is sound and scoped to only what's no longer reachable.

### Summary

Schema is untouched (no migration risk — confirmed against `crates/airc-store/src/migration/` ordering m001-m016, neither m015 tier nor m016 endpoints are altered by this PR). Tier-aware rendering is correct via `airc_trust::load` which is unfiltered. Surface coverage is intentionally partial with a tracked follow-up. The two genuine asks: (1) acknowledge the output-shape contract change publicly (PR body or a `--legacy` flag for one cycle), (2) add 2-3 tests pinning the new behavior before this lands so the next slice has a regression net. Not blocking, but recommended before merge.

**Disposition: RISK (not BLOCK).** Land it after acknowledging the output-shape break and ideally adding minimal test coverage; the seam #2 direction is right and the delegation is clean.

joelteply:
Addressed both genuine asks in `4b9a4d7`:

**(1) Output-shape break — acknowledged + verified safe.** Ran your suggested grep: `grep -rn "collaboration peers" docs/ skills/ install*.{sh,ps1} integrations/ scripts/` → **zero** downstream parsers of the old `name -> host` shape, so nothing breaks silently. Added a one-line **stderr** deprecation banner pointing at `airc peer list` (stderr, not stdout, so any pipe consuming the list stays byte-for-byte unaffected). Updating the PR body too.

**(2) Regression net — 3 tests added.** Extracted the human render into a pure `render_peer_list_lines(peers, home)` so the contract is testable without stdout capture:
- `render_peer_list_shows_every_tier_from_trust_store` — Untrusted **and** Friend both render with their `tier=` shape (pins the unfiltered `load` + line format — your point (a)+(b)).
- `render_peer_list_ignores_legacy_peers_json_files` — a stray `<home>/peers/ghost.json` is invisible (your point (c): proves the view is the trust store, not files).
- `render_peer_list_empty_is_onboarding_hint` — empty store → `airc peer add` hint, not blank/panic.

**NIT:** added the inline comment in `run_prune_peers` flagging the temporary three-surface asymmetry + pointing at the remaining #2 slice (retire the handshake file writer, drop `prune-peers`).

Disposition was RISK-not-BLOCK; both asks now closed. Good catches — this is the flywheel working.

PR #1195, opened 2026-06-14
feat(peer): age/TTL staleness gate on peer prune (seam #3.2 slice 3)

joelteply:
## Seam #3.2 slice 3 — the payoff

**Stacked on #1194 → #1192.** ⚠️ Retarget base down the chain as each merges (→ #1194's branch, then `canary`), before any base branch is deleted (the #1191 auto-close lesson).

This is what the `last_seen_ms` column (#1192) and the beacon-import touch (#1194) were built **for**. `classify_peer_prune`'s own doc said *"the trust store has no last_seen/TTL, so dead enrolments never expire without this"* — this PR is the *this*.

### Behavior change (strictly safer)
**Before:** an Untrusted peer absent from one registry snapshot was evicted *immediately* — even one we'd imported a beacon from seconds earlier (snapshot lag looked like death).

**Now:** eviction requires Untrusted **AND** absent **AND** stale (`now_ms - last_seen_ms >= TTL`). A recently-contacted peer gets a grace window (`DEFAULT_PEER_STALE_AFTER_MS` = 1h). A dead Docker-ghost never refreshes its beacon, so it ages past the window and still evicts. Fewer evictions, more correct.

- `classify_peer_prune` signature: `enrolled` is now `(PeerId, TrustTier, last_seen_ms)` triples + `now_ms` + `stale_after_ms`. `saturating_sub` makes a future `last_seen` (clock skew) read as age 0 → kept (no panic).
- `run_peer_prune` passes `p.last_seen_ms` + wall-clock now + the default TTL; forensic remove reason notes the staleness gate.

### Tests
- existing `classify_peer_prune_evicts_only_untrusted_absent` updated (stale `last_seen` so the gate is satisfied — proves the tier + live-set guards still hold).
- new `classify_peer_prune_grace_window_keeps_recently_seen` — keep-within-TTL, keep-on-future-`last_seen`, evict-once-aged.

### Completes seam #3.2
Slices: 1 (#1192 column) → 2 (#1194 touch wiring) → **3 (this, the eviction gate)**. The liveness spine is now end-to-end: enrol seeds recency → fresh beacons advance it monotonically → prune ages out the truly-silent.

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -1880,8 +1880,13 @@ pub async fn run_peer_prune(home: &Path, apply: bool) -> Result<(), Box<dyn std:
         println!("(no enroled peers — nothing to prune)");
         return Ok(());
     }
-    let enrolled_pairs: Vec<(airc_core::PeerId, airc_lib::TrustTier)> =
-        enrolled.iter().map(|p| (p.peer_id, p.tier)).collect();
+    // Seam #3.2: carry each peer's last_seen so the classifier can apply
+    // the staleness grace window (a recently-contacted peer absent from
+    // one registry snapshot is kept, not evicted as a ghost).
+    let enrolled_triples: Vec<(airc_core::PeerId, airc_lib::TrustTier, u64)> = enrolled
+        .iter()
+        .map(|p| (p.peer_id, p.tier, p.last_seen_ms))
+        .collect();
 
     // Authoritative live set: the fresh, stale-pruned account-registry
     // document (READ-ONLY — `live_registry_peer_ids` does not enrol).
@@ -1905,7 +1910,16 @@ pub async fn run_peer_prune(home: &Path, apply: bool) -> Result<(), Box<dyn std:
         }
     };
 
-    let verdicts = airc_lib::classify_peer_prune(&enrolled_pairs, &live_ids);
+    let now_ms = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_millis() as u64)
+        .unwrap_or(0);
+    let verdicts = airc_lib::classify_peer_prune(
+        &enrolled_triples,
+        &live_ids,
+        now_ms,
+        airc_lib::DEFAULT_PEER_STALE_AFTER_MS,
+    );
     let mut to_evict = Vec::new();
     for verdict in &verdicts {
         match verdict.action {
@@ -1948,7 +1962,7 @@ pub async fn run_peer_prune(home: &Path, apply: bool) -> Result<(), Box<dyn std:
         if airc
             .remove_peer(
                 *peer_id,
-                "peer-prune: untrusted + absent from fresh registry",
+                "peer-prune: untrusted + absent from fresh registry + stale past last_seen TTL",
             )
             .await?
         {
--- a/crates/airc-lib/src/account_registry.rs
+++ b/crates/airc-lib/src/account_registry.rs
@@ -531,6 +531,10 @@ impl Airc {
         document: AccountRegistryDocument,
     ) -> Result<(), AircError> {
         document.validate().map_err(AircError::AccountRegistry)?;
+        // Seam #3.2 (liveness): the wall-clock instant we are importing
+        // at — the upper bound for any peer's `last_seen`. See the clamp
+        // at the touch below.
+        let now_ms = crate::time::now_ms()?;
         for peer in document.peers {
             if peer.peer_id() == self.inner.identity.peer_id {
                 continue;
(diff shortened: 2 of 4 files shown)


joelteply:
🔴 SENTINEL: BLOCK

Reviewed `origin/feat/peer-last-seen-touch-wiring...origin/feat/peer-prune-staleness-ttl`. Build is clean — `cargo fmt --check`, `cargo check`, `cargo clippy`, and the classify tests all pass on airc-lib + airc-cli (both new tests green: `classify_peer_prune_evicts_only_untrusted_absent`, `classify_peer_prune_grace_window_keeps_recently_seen`). Most of the change is sound. One real defect blocks.

## BLOCKING: future-dated / replayed beacon pins `last_seen` and defeats the GC forever

This slice wires BOTH halves of an exploitable path:

1. `crates/airc-lib/src/account_registry.rs:555` (added in THIS PR) stamps `last_seen = peer.presence.heartbeat_at_ms` — the peer's OWN self-asserted timestamp — and the store applies it monotonically with **no upper clamp**: `crates/airc-store/src/sqlite.rs:516` `let next = current.max(seen_at_ms);` (permanent — can never be lowered).
2. `crates/airc-lib/src/peers.rs:91-92` now KEEPS an Untrusted+absent peer when `now - last_seen < TTL`.

Compose them: an untrusted peer (or a buggy/skewed clock, or a **replayed** captured beacon) publishes `heartbeat_at_ms` in the future. `touch_last_seen` stores it verbatim and permanently. Later, when that peer stops republishing and ages out of the fresh registry, the classifier computes `age = now.saturating_sub(future) = 0 < TTL` → **Keep, indefinitely**. The dead enrolment this command exists to reclaim can never be reclaimed. Before this PR, `Untrusted + absent` evicted immediately — so this is a regression in eventual cleanup, gated on a value an untrusted peer fully controls.

The PR's own new test makes this concrete and locks it in: `account_registry.rs` `import_registry_touches_last_seen_from_beacon_heartbeat` uses `fresh_hb = 9_999_999_999_999` (~year 2286) and asserts `last_seen_ms == fresh_hb`. That is a future-dated, peer-asserted heartbeat stored unclamped — the test proves the exact primitive the eviction gate now trusts.

**Fix (small):** clamp at the touch site to local wall-clock — `let seen = seen_at_ms.min(now_ms);` before the monotonic `max` (the comment at peers.rs:88-90 already reasons about future last_seen via `saturating_sub` for the *age* computation, but never bounds what gets *stored*, so a future stamp survives as a permanent keep-forever). Add a regression test: future-dated heartbeat → peer still evicts once absent + aged.

## Vectors that held up (verified, not blocking)

- **Safety direction (#2):** confirmed strictly-fewer-evictions. The new gate is an extra AND on the Evict branch only; the live-set Keep (peers.rs:74) and tier Keep (peers.rs:79) branches are evaluated first and untouched. No path evicts something the old logic kept.
- **Docker-ghost (the stated target) — #1:** holds. Enrolment seeds `last_seen = added_at_ms` (`sqlite.rs:261`), NULL floors to `added_at_ms` (`stored_last_seen_ms`, sqlite.rs:855), a never-touched ghost ages past 1h and evicts. Correct.
- **saturating_sub edges (#3):** future last_seen → age 0 → kept; `last_seen=0`/`now=0` → age 0; `TTL=0` → evicts. All safe, no panic. Tested by the `skewed` case.
- **Clock domain (#5):** `now_ms` (SystemTime epoch-ms, commands.rs:1913) and `last_seen_ms` (from `heartbeat_at_ms`, epoch-ms wall-clock) are the same domain — the coordinator already does `now - heartbeat_at_ms` for live/stale partitioning. Match confirmed.
- **Test weakening (#4):** NOT weakened. The modified `classify_peer_prune_evicts_only_untrusted_absent` sets `friend_absent` and `live_untrusted` to `last_seen=0` (maximally stale), so the tier guard and live-set guard are still genuinely load-bearing — dropping either still flips an assert — and the `count == 1` assert remains a strong mutation killer.
- **1h default (#6):** defensible (generous for registry snapshot lag, short for ghost turnover). Minor: it is hardcoded with no operator override / CLI flag; consider plumbing it as the system grows. Not blocking on its own.

The clamp is the blocker — an untrusted peer must not be able to make itself un-evictable by lying about its clock.

joelteply:
Fixed in `a02a16b` — excellent catch, this was a real exploit. Clamped `heartbeat_at_ms.min(now_ms)` at the touch site (account_registry.rs): last_seen means *when WE had contact*, which can't be in the future, so the most we record is "received now". The store's monotonic max still ignores older/stale beacons (no rewind). Rewrote the test as a regression net: a year-2286 beacon is clamped to the import instant (never stored verbatim) AND the peer still **Evicts** once TTL elapses — the identical `classify_peer_prune` call would `Keep` forever if the clamp regressed. The 1h-hardcoded-TTL nit is noted for a follow-up (operator override). Re-review welcome.

joelteply:

Re-reviewed the fix delta (a02a16b) against the original BLOCK. **The exploit is closed.**

**1. Clamp completeness — the ONLY untrusted-input path is covered.** Grepped the whole repo: `touch_last_seen` has exactly ONE production caller — `import_account_registry_document` (account_registry.rs:568), now clamped `heartbeat_at_ms.min(now_ms)`. The store's `touch_peer_last_seen` (sqlite.rs:495) does monotonic `current.max(seen_at)` with no upper bound — it trusts its caller, which is acceptable because the sole caller clamps. The only other `last_seen_ms` column writes are `add_peer_trust_with_tier` and `replace_peer_trust`, both seeding from server-local `added_at_ms` (via `airc_trust::add` → `now_ms()`), never a peer value. No second leak.

**2. now_ms placement — harmless.** Computed once before the loop. If a late peer's clamp ceiling falls below its own `added_at_ms`, the monotonic max makes the touch a no-op and `last_seen` floors to `added_at_ms` (real local clock, via `stored_last_seen_ms`) — never 0, never future. Correct.

**3. Residual state — none, no migration needed.** Confirmed canary's `classify_peer_prune` does NOT read `last_seen` (it evicts untrusted+absent immediately). The exploitable age-gate (`keep when now-last_seen<TTL`) is INTRODUCED by THIS PR, and the clamp ships in the same PR. So no pre-existing row can carry a future `last_seen` into a live gate. Going-forward clamp is sufficient; clamp-on-read migration unnecessary.

**4. Test is a real regression net — verified empirically.** Reverted `.min(now_ms)` locally → `import_clamps_future_heartbeat_keeping_peer_evictable` FAILS (`assertion left != right: a self-asserted future heartbeat must never be stored verbatim`, account_registry.rs:909). Restored → passes. It also asserts the end-to-end Evict-after-TTL that would `Keep` forever if the clamp regressed. Not green-either-way.

**5. Slice-3 classifier + merged slices sound.** `classify_peer_prune` TTL gate uses `saturating_sub` (future last_seen → age 0 → kept, no panic); tier guard + live-set guard intact. airc-lib lib: 246 passed / 0 failed. airc-store touch tests: 2 passed. `cargo clippy -p airc-lib`: clean. `cargo fmt --check`: clean.

The original future-beacon evasion is closed at its single touch site, and the test is a genuine regression net.

PR #1196, opened 2026-06-14
feat(peer): touch last_seen on beacon import (seam #3.2 slice 2)

joelteply:
## Seam #3.2 slice 2 — wire the touch

Rebased onto canary now that slice 1 (#1192, the `last_seen_ms` column) merged. Reopens #1194, which GitHub auto-closed when #1192's branch was deleted on merge (the stacked-base lesson — not a rejection).

Slice 1 added the column + monotonic `touch_peer_last_seen`, but nothing called it — recency never advanced past enrolment. This wires the strongest liveness signal: a peer present in a freshly refreshed, stale-pruned account-registry document **just published a beacon**, so `import_account_registry_document` now `touch_last_seen`'s each imported peer.

- Uses the beacon's **own `heartbeat_at_ms`** (the peer's liveness assertion), not local `now_ms`.
- Monotonic: a clock-skewed / out-of-order beacon can only leave recency unchanged, never rewind it.
- Best-effort (not asserted `Some`): the peer was enrolled two lines up; unlike the endpoints store whose silent drop was a real bug (#1120).

**Test:** `import_registry_touches_last_seen_from_beacon_heartbeat` — a future-dated beacon advances `last_seen` to the heartbeat; a later older beacon does not rewind it.

Stack: #1192 (merged) → **this** → #1195 (slice 3, age/TTL prune gate).

### DIFF
--- a/crates/airc-lib/src/account_registry.rs
+++ b/crates/airc-lib/src/account_registry.rs
@@ -541,6 +541,23 @@ impl Airc {
                 peer.peer_spec.pubkey,
             )
             .await?;
+            // Seam #3.2 (liveness): a peer present in a freshly
+            // refreshed, stale-pruned registry document just published a
+            // beacon — that IS fresh contact. Record it so the age-based
+            // eviction classifier can tell a live peer from a stale
+            // enrolment. Use the beacon's OWN heartbeat instant (the
+            // peer's liveness assertion), not local `now_ms`: the touch
+            // is monotonic, so a clock-skewed or out-of-order beacon can
+            // only ever leave recency unchanged, never rewind it. The
+            // peer was enrolled two lines up, so this resolves to Some;
+            // we don't assert it (a best-effort liveness stamp, unlike
+            // the endpoints store below whose silent drop was a real bug).
+            airc_trust::touch_last_seen(
+                &self.inner.wire_root,
+                peer.peer_spec.peer_id,
+                peer.presence.heartbeat_at_ms,
+            )
+            .await?;
             // Card 625abe6d slice 1: persist the beacon's endpoints on
             // the trust record so route discovery can dial them after
             // a restart (the in-memory ImportedInviteTable fed below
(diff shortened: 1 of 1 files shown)


joelteply:
🟢 SENTINEL: APPROVE

Tried hard to break this across all six attack surfaces; every one resolves to either correct-by-construction or a pre-existing, documented, intentional design choice this PR does not widen.

1. TIMESTAMP DOMAIN — `heartbeat_at_ms` and prune's local `now_ms` ARE the same epoch-ms domain (`PresenceBeacon::is_fresh` at coordinator.rs:113 already compares them directly; `from_snapshot`/`beacon_now` stamp wall-clock epoch-ms). The future-clock 'un-prunable peer' outcome is REAL but is the *intended* fail-safe, already present pre-PR: `classify_peer_prune` (peers.rs:91) does `now_ms.saturating_sub(last_seen_ms)` → future last_seen reads age 0 → Keep, by explicit design (peers.rs:62-63). The touch stores the same value `is_fresh` already trusts, so it introduces no new domain mismatch. Not a new DoS vector; the existing 'keep over evict' policy is a deliberate choice, not a regression.

2. BEST-EFFORT touch not asserted Some — verified safe. `add` (add_peer_trust, sqlite.rs:263-283) does insert-on-conflict-do-nothing then `find_by_id().ok_or_else(...)`, so `add(...).await?` returns Ok ONLY when the row provably exists. No early-continue/conditional between add and touch. None is unreachable here; dropping it (concern #6) is correct and documented.

3. SELF-SKIP — confirmed: `continue` on self (account_registry.rs:535) is BEFORE `add`/`touch`. We never touch our own row.

4. STALE re-introduction — production `GhAccountRegistryStore::refresh` runs `prune_stale_peers` (gh/account_registry.rs:935, 600s TTL) BEFORE the document reaches `import_account_registry_document`. Genuinely-dead peers are dropped pre-import; survivors have a heartbeat within the freshness horizon, so their touched last_seen is legitimately fresh. No stale peer gets a fresh-looking stamp.

5. TEST — sound. `beacon_now(...,hb)` lands `fresh_hb` in `heartbeat_at_ms` (the exact field touch reads, coordinator.rs:416). Enrolment seeds last_seen=added_at (~real now); asserting last_seen==9_999_999_999_999 proves the touch ADVANCED past the column default, not just the default. Second import with hb=1_000 asserting no rewind proves monotonicity. Both behaviors genuinely exercised.

6. Option drop — correct (see #2): None only on unenrolled peer, unreachable after the guaranteeing `add`. Matches the documented best-effort contract (vs. the endpoints store whose ok_or_else guards a real #1120 bug).

Checks I ran myself (not the PR description): `cargo check -p airc-lib` clean; `cargo test -p airc-lib --lib import_registry_touches_last_seen` → 1 passed; `cargo clippy -p airc-lib` zero warnings; `cargo fmt -p airc-lib --check` clean. (Integration-test binaries hit an OS Application Control block / os error 4551 — environmental, pre-existing, branch-independent; the new unit test lives in the lib target and runs green.)

No real defect found after genuinely trying to break it. Ship it.

PR #1197, opened 2026-06-14
fix(codex-hook): drain_stdin gets a 5s deadline — no more Windows CI hangs (airc#1097)

joelteply:
Rebased cherry-pick of 78c77c6 from closed PR #1098 onto fresh canary. Same surgical fix: stdin drain on worker thread with 5s deadline. Single-file change. 256 airc-cli tests pass. Original #1098 was rebase-stuck via the rust-rewrite squash promote; this is the surgical resurrection per Mac's 2026-06-14 backlog triage.

### DIFF
--- a/crates/airc-cli/src/integrations/codex/hook.rs
+++ b/crates/airc-cli/src/integrations/codex/hook.rs
@@ -216,6 +216,37 @@ fn is_self_event(event: &TranscriptEvent, airc: &Airc, runtime_client: Option<&s
 /// fast-fail-and-proceed — the hook produces no AdditionalContext
 /// for that call, which is the same outcome as receiving empty
 /// stdin, and the next hook invocation behaves correctly.
+///
+/// **Intentional trade-offs the deadline introduces** (BIGMAMA
+/// review BLOCK#2 + #3 on PR #1197):
+///
+/// 1. **Slow-but-legitimate producer truncates to empty payload.**
+///    The deadline is on EOF, not on byte progress. A producer that
+///    streams valid JSON byte-by-byte over >5s gets its input
+///    discarded — `read_to_string` never returns, so the partial
+///    buffer in the reader thread is dropped and the hook proceeds
+///    with empty payload. The current caller (Codex) flushes JSON in
+///    microseconds; if a future caller streams slowly, the deadline
+///    needs to reset on progress or read to a JSON-object delimiter
+///    rather than EOF. Today we accept the truncation because the
+///    alternative is "Windows CI hangs for hours."
+///
+/// 2. **Timeout silently skips JSON validation.** The EOF path
+///    rejects non-object input with a hard error
+///    (`!value.is_object()` → `Err(...)` → non-zero exit). The
+///    timeout path returns `Ok(())` unconditionally — so "malformed
+///    input that also arrives slowly" degrades from a hard error to
+///    a silent success. The `eprintln!` distinguishes timeout from
+///    EOF in logs, but the validation-skip is intentional: a hook
+///    that exits non-zero on slow-malformed input would block every
+///    Codex prompt-submit on the same Windows hang it was meant to
+///    end. Operators see WHY via stderr; the input contract is
+///    "validates iff EOF arrives in time."
+///
+/// Both trade-offs are pinned by
+/// `drain_stdin_timeout_proceeds_when_eof_never_arrives` in
+/// `tests/codex_hook_commands.rs` — see the regression test for the
+/// exact failure shape if a future refactor reverts the deadline.
 fn drain_stdin() -> Result<(), Box<dyn std::error::Error>> {
     use std::sync::mpsc;
     use std::thread;
--- a/crates/airc-cli/src/integrations/codex/hook.rs
+++ b/crates/airc-cli/src/integrations/codex/hook.rs
@@ -191,9 +191,60 @@ fn is_self_event(event: &TranscriptEvent, airc: &Airc, runtime_client: Option<&s
     event.client_id == airc.client_id()
 }
 
+/// Read Codex's hook JSON from stdin, with a hard deadline so a
+/// hung pipe never deadlocks the process.
+///
+/// **Why the deadline exists** (airc#1097). On the Windows CI
+/// runner, the codex_hook_commands tests hung for 5+ hours because
+/// `std::io::stdin().read_to_string` waited forever for an EOF that
+/// never arrived — most likely a Windows handle-inheritance leak
+/// from `Stdio::piped()` parent → grandchild daemon, where the
+/// daemon's inherited duplicate of the pipe-read handle kept the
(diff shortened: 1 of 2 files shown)


joelteply:
🔴 SENTINEL: BLOCK

Reviewed `origin/fix/codex-hook-drain-deadline-v2` vs `origin/canary`. Build is clean (cargo fmt --check, clippy --all-targets, full `cargo test -p airc-cli`: 264+16+… all green, 0 failed, 2 ignored — the "256 tests pass" claim holds). The mechanism is sound *as a way to not deadlock the process*: the orphan reader thread is detached, and a detached thread does not keep a Rust process alive, so `main` returning exits cleanly even with the read still blocked. Worst-case latency is genuinely bounded to ~5s. So far so good.

But this is a fix for a **5-hour hang that ships with zero tests exercising the thing it fixes**, plus two real behavioral regressions on the timeout branch. Blocking:

**1. No regression test for the hang (decisive). `hook.rs:194`**
Every `run_hook` helper in `tests/codex_hook_commands.rs:522-575` does `write_all(stdin)` then drops the pipe — EOF *always* arrives, so the new `recv_timeout` timeout branch (`hook.rs:228-241`) is **never executed by any test in the suite**. A fix for a hang with no hang-regression test cannot prove it fixes the hang or guard against a future refactor reintroducing it. Add a test that spawns the hook with `Stdio::piped()` stdin and **never writes/closes it**, asserting the process exits `success()` within (say) ~7s and emits the stderr diagnostic. That test is the entire point of the PR and it is absent.

**2. Silent truncation of a slow-but-legitimate producer. `hook.rs:213-217`**
On timeout the partially-read buffer lives inside the spawned thread and is dropped — `read_to_string` never returned, so nothing is sent on the channel. A producer that streams valid JSON but holds the pipe open >5s before EOF gets its input **silently discarded** and the hook proceeds with empty payload. The doc-comment asserts "Codex sends the JSON in microseconds," but that is an assumption about the *only* current caller, not an invariant of stdin. A read deadline that resets on progress (or reading to a delimiter/EOF of the JSON object rather than EOF-of-stream) would not have this failure mode. At minimum the 5s-with-no-byte-progress vs 5s-mid-stream cases should be distinguished.

**3. Timeout silently skips JSON validation. `hook.rs:200-203` vs `228-241`**
The EOF path validates: non-object stdin → `Err("Codex hook stdin must be a JSON object")`, propagated to a non-zero exit. The timeout path returns `Ok(())` unconditionally. So "malformed input that also arrives slowly" degrades from a hard error to a silent success. This is surfaced via `eprintln!` (good — timeout *is* distinguishable from EOF in logs, addressing the EOF-vs-timeout concern), but the validation-contract divergence is undocumented and untested.

**4. Platform: the bug is Windows-specific; the fix is validated only on Linux/Mac.**
The doc-comment itself names the root cause as a Windows handle-inheritance leak to the grandchild daemon (`Stdio::piped()` parent→daemon keeping the read handle alive). This PR does **not** address that leak — it abandons the read and lets the process exit, which is a legitimate *scope decision* (hook process no longer hangs), BUT: the daemon still inherits and holds the duplicated handle on Windows. The hosted clippy/CI gate is ubuntu-only (`main.rs:138` comment confirms), the test suite ran here on Windows is green but never hits the timeout branch, so the actual Windows deadlock-becomes-fast-fail behavior is **assumed, not demonstrated**. (Sidebar: this review found 8 orphaned `airc.exe` test daemons lingering after the suite — consistent with daemon-handle-retention being a live, unaddressed concern on this platform.)

**To flip to APPROVE:** add the never-closing-stdin regression test (#1) — that alone is the minimum bar for a hang fix — and either fix or explicitly document the slow-producer truncation (#2) and the validation-skip-on-timeout divergence (#3). The deadline mechanism itself is acceptable; the gap is that nothing tests it and two silent-data-loss edges ship unguarded.

joelteply:

Re-reviewed `fix/codex-hook-drain-deadline-v2` @ a3b9b65 on a Windows host (detached worktree at PR head). All four prior BLOCK findings are resolved.

## 1. Hang-regression test (decisive) — EXISTS, DRIVES THE TIMEOUT BRANCH, PASSES on Windows
`drain_stdin_timeout_proceeds_when_eof_never_arrives` (tests/codex_hook_commands.rs):
- Spawns `codex-hook user-prompt-submit` with `Stdio::piped()` stdin and **never writes or closes it**, holding the handle via `child.stdin.take()` for the entire wait. `wait_with_output` is correctly avoided (it would close stdin and deliver EOF early, defeating the test).
- Polls `try_wait()` against a 15s wall-clock budget so a deadline regression manifests as a *failed* test, not a hung run.
- Asserts `status.success()`, bounded elapsed, **and** that stderr contains `"stdin EOF not received"` — the diagnostic emitted *only* on the `recv_timeout` Err branch. This is the load-bearing assertion: the test cannot pass via the EOF path, because EOF would suppress that string.

Verified on Windows:
- `cargo test -p airc-cli --test codex_hook_commands drain_stdin_timeout_proceeds_when_eof_never_arrives` → **ok**.
- Instrumented the in-test elapsed: **5.62s** (spawn → child exit) — i.e. the 5s deadline + ~0.6s spawn/poll slop. Not instant (would = EOF path), not hung. The timeout branch is genuinely exercised.

## 2 & 3. Slow-producer truncation + timeout-skips-JSON-validation — DOCUMENTED as intentional trade-offs, honestly
The `drain_stdin` doc-comment now carries both, and the reasoning is sound and honest:
- **Trade-off #1 (truncation):** explicitly states the deadline is on EOF not byte-progress, a >5s streaming producer is discarded, names that the only current caller (Codex) flushes in microseconds, and names the future fix (reset-on-progress / read-to-delimiter). Accepted because the alternative is hours-long Windows CI hangs.
- **Trade-off #2 (validation skip):** honestly states the EOF path hard-errors on non-object input while the timeout path returns `Ok(())` unconditionally, so malformed-AND-slow input degrades from hard error to silent success; justified because a non-zero exit would re-block every Codex prompt-submit on the same hang. Operators see WHY via stderr. Contract restated: "validates iff EOF arrives in time."

## 4. Ok vs Err behavior — sound
Hung producer → give up → exit clean is correct for an advisory hook (it produces AdditionalContext; failing non-zero would break Codex prompt-submit). The misreport window for slow-malformed input is theoretical, documented, and stderr-visible.

## 5. Full gate — GREEN on Windows
- `cargo check -p airc-cli` ✅
- `cargo clippy -p airc-cli --all-targets -- -D warnings` ✅
- `cargo fmt --check` ✅
- `cargo test -p airc-cli --test codex_hook_commands` → **17/17 passed**, no flake ✅
- `cargo test -p airc-cli --bins` → **283 passed, 2 ignored** ✅

No Windows-specific flake observed across repeated runs. The fix is now guarded by a test that reproduces the exact original failure shape if the deadline regresses.

PR #1198, opened 2026-06-14
fix(work): cleanup handles nested `<short>/src/` worktree layout (follow-up to c9b28925)

joelteply:
Rebased cherry-pick of 7260fdc from closed PR #1107 onto fresh canary, plus one-line test-call-site fix for the new `classify_worktree(_, _, upstream_gone: bool)` signature that landed on canary post-rebase. 55 work_commands tests pass. Original #1107 was rebase-stuck via the rust-rewrite squash promote.

### DIFF
--- a/crates/airc-cli/src/work_commands.rs
+++ b/crates/airc-cli/src/work_commands.rs
@@ -955,7 +955,7 @@ pub async fn run_cleanup(
         // status, that's enough to remove regardless of the kanban
         // projection's state — which often lags reality when a
         // `gh pr merge` bypasses `airc work merge`.
-        let upstream_gone = probe_upstream_gone(&path);
+        let upstream_gone = probe_upstream_gone(&effective);
 
         let disposition = classify_worktree(card_state.as_ref(), &dirty_status, upstream_gone);
 
--- a/crates/airc-cli/src/work_commands.rs
+++ b/crates/airc-cli/src/work_commands.rs
@@ -936,28 +936,8 @@ pub async fn run_cleanup(
             .and_then(|c| c.pull_request.as_ref().map(|pr| pr.number));
         let repo = card.as_ref().map(|c| c.repo.to_string());
 
-        // Card 83a5624e: resolve <short>/ vs <short>/src/ BEFORE
-        // probing so the operator sees the actual git worktree path
-        // in the disposition table AND the force-remove loop targets
-        // the right directory.
-        let effective = resolve_worktree_path(&path);
-
-        // Probe for both uncommitted AND unpushed work — see
-        // `probe_dirty_status` for the WIP-outranks-hygiene contract.
-        let dirty_status = probe_dirty_status_at(&effective);
-
-        // Probe whether the worktree's branch still exists on origin.
-        // `gh pr merge --delete-branch` (the default merger path AND
-        // every manual merge that uses --delete-branch) removes the
-        // branch on origin AFTER merging the PR. That's the universal
-        // signal a worktree is dead: branch upstream is gone ⇒ either
-        // merged-and-deleted or abandoned. Combined with `Clean` dirty
-        // status, that's enough to remove regardless of the kanban
-        // projection's state — which often lags reality when a
-        // `gh pr merge` bypasses `airc work merge`.
-        let upstream_gone = probe_upstream_gone(&effective);
-
-        let disposition = classify_worktree(card_state.as_ref(), &dirty_status, upstream_gone);
+        let (effective, dirty_status, disposition) =
+            classify_worktree_path(&path, card_state.as_ref());
 
         classifications.push(WorktreeClassification {
             short,
@@ -965,7 +945,7 @@ pub async fn run_cleanup(
             card_state,
             pr_number,
             repo,
-            dirty_status: dirty_status.clone(),
+            dirty_status,
             disposition,
         });
     }
@@ -1142,6 +1122,26 @@ pub(crate) enum DirtyStatus {
 /// override an out-of-sync kanban state and remove — fixes the
 /// recurring disk-full crash where worktrees for already-merged
 /// PRs lingered because the projection hadn't observed the merge.
+/// BIGMAMA review on PR #1198: extract the per-worktree classification
+/// pipeline so the regression nets the PRODUCTION code path. Reverting
+/// `probe_upstream_gone(&effective)` → `&path` inside THIS function now
+/// makes the nested-layout test go red — which is what a regression net
+/// must actually guarantee.
+///
+/// Card 83a5624e: resolve `<short>/` vs `<short>/src/` BEFORE probing so
+/// the operator sees the actual git worktree path in the disposition
+/// table AND the force-remove loop targets the right directory.
+pub(crate) fn classify_worktree_path(
+    path: &std::path::Path,
+    card_state: Option<&CardState>,
+) -> (std::path::PathBuf, DirtyStatus, Disposition) {
+    let effective = resolve_worktree_path(path);
+    let dirty_status = probe_dirty_status_at(&effective);
+    let upstream_gone = probe_upstream_gone(&effective);
+    let disposition = classify_worktree(card_state, &dirty_status, upstream_gone);
+    (effective, dirty_status, disposition)
+}
+
 pub(crate) fn classify_worktree(
     card_state: Option<&airc_work::CardState>,
     dirty: &DirtyStatus,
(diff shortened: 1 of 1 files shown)


joelteply:
🔴 SENTINEL: BLOCK

Verified the cherry-pick is otherwise clean and the suite is real — but found one genuine correctness gap that defeats this PR's own headline scenario via the manual cleanup path.

## Defect — `upstream_gone` probe runs against the UNRESOLVED parent

`crates/airc-cli/src/work_commands.rs:958` (inside `run_cleanup`):

```rust
let effective = resolve_worktree_path(&path);          // ✓ resolves <short>/ → <short>/src/
let dirty_status = probe_dirty_status_at(&effective);  // ✓ uses effective
let upstream_gone = probe_upstream_gone(&path);        // ✗ uses UNRESOLVED parent
let disposition = classify_worktree(card_state.as_ref(), &dirty_status, upstream_gone);
classifications.push(WorktreeClassification { path: effective, .. }); // ✓ effective
```

For the exact nested `<short>/src/` layout this PR targets, `probe_upstream_gone(&path)` runs `git -C <short>/ rev-parse --abbrev-ref @{u}` against the **non-git parent container**, which fails → returns `false` (its documented safe default). So the `upstream_gone` **trump card silently never fires for nested layouts**.

Why that matters: `classify_worktree` treats `upstream_gone` as the override for exactly the case the surrounding comment calls out — "the kanban projection lags reality when a `gh pr merge` bypasses `airc work merge`." In that scenario (nested worktree + branch deleted on origin + card NOT yet Closed/Merged) the worktree classifies as `KeepActive` and is **not reclaimed** — which is precisely the "stranded after merge" class of bug this PR exists to fix. The fix is therefore incomplete for its own target scenario through the manual `airc work cleanup` path.

Severity: safe false-negative (never deletes), and the merger hook `cleanup_card_worktree` IS fully resolved (it routes branch-detection + `git worktree remove` through the resolved `worktree_path`). So this only leaks via the manual command — but it's a real, in-scope gap.

Fix (one line): `let upstream_gone = probe_upstream_gone(&effective);`

## Test coverage gap (why CI is green despite the bug)

`classifier_handles_nested_src_layout_via_probe` exercises the nested layout only through the **card-state** branch (`Closed` → `Removable`) with a hardcoded `false` for `upstream_gone`. No test drives `upstream_gone == true` THROUGH the nested resolution, so the leak above is uncovered. Add a case: nested `<short>/src/` whose branch is deleted on origin + a non-terminal card state, asserting `Removable` via the upstream-gone path. It will fail today.

## What I verified as sound (credit where due)

- Cherry-pick of 7260fdc is faithful; the new 3-arg `classify_worktree(_, _, upstream_gone)` adaptation (`false` added at `:958` call + the nested-layout test at `:2786`) is correct — the original #1107 had the 2-arg form and no `upstream_gone` probe; `probe_upstream_gone` landed separately on canary.
- `resolve_worktree_path` is a pure path *selector*; removability still gates through `probe_dirty_status` (WIP guard) + `classify_worktree`. Not too aggressive: neither-is-git → returns parent unchanged → `Unknown` → `SkipNotGit` (pinned by `resolve_worktree_path_returns_input_when_neither_is_git`). Real source dirs are not at risk.
- `is_git_worktree` correctly accepts both regular (`.git/` dir) and linked (`.git` file) worktrees via `.exists()`.
- One-level-only nesting is explicitly scoped per `[[no-fallbacks-ever]]`; paths originate from `read_dir` PathBufs so trailing-slash/backslash/dot-component edge cases do not arise. Acceptable.
- `cleanup_card_worktree` (the merger hook — primary repro path) resolves once and uses the resolved path consistently everywhere. Correct.
- Local run (isolated target dir, to avoid the operator's running cargo/airc daemons holding the shared cache): `cargo fmt --check` clean, `cargo clippy -p airc-cli` zero warnings, **55 work_commands tests pass**.

Re-request review after wiring `upstream_gone` to `effective` + adding the upstream-gone-through-nested test.

joelteply:
🔴 SENTINEL: BLOCK (re-review)

The production fix is correct, but the regression the prior review demanded is still unprotected. Re-review scope confirmed below.

## ✅ The call-site fix is correct
- `run_cleanup` line 943: `let effective = resolve_worktree_path(&path);`
- line 947: `probe_dirty_status_at(&effective)` (inner probe, no double-resolve)
- line 958: `probe_upstream_gone(&effective)` ← the exact line that was the blocking defect, now fed the RESOLVED path
- line 964: `path: effective` — the disposition table shows the real worktree path
- force-remove loop (line 1038) calls `git_worktree_remove(&c.path)` where `c.path == effective`, so removal targets the resolved dir too.

The unresolved-parent bug is genuinely closed for the nested `<short>/src/` layout.

## ✅ Build / test / clippy / fmt (isolated target `~/.continuum/cache/rev-1198`)
- `cargo check -p airc-cli`: clean
- `cargo test -p airc-cli --bin airc`: **269 passed, 0 failed, 2 ignored**
- `cargo clippy -p airc-cli --bin airc`: clean
- `cargo fmt -p airc-cli -- --check`: clean

## 🔴 BLOCK: still no test driving `upstream_gone == true` THROUGH the nested resolution
This is the precise gap the prior review flagged, and the fix commit did not net it.

- `probe_upstream_gone` is called in **exactly ONE place** — the production call site (line 958). **Zero tests** invoke it.
- Crucially, `probe_upstream_gone` does **not** self-resolve the nested layout (unlike `probe_dirty_status`, whose public wrapper resolves internally). It relies 100% on the caller passing the resolved path. So the correctness of the nested upstream-gone path lives entirely on line 958 being `&effective` and nothing else.
- The new tests cover the wrong axis:
  - `classifier_upstream_gone_*` (lines 2449-2510) pass a **hardcoded `true`** to `classify_worktree` — they never touch `probe_upstream_gone` or nested resolution.
  - `probe_dirty_status_handles_nested_src_layout` / `classifier_handles_nested_src_layout_via_probe` exercise the **dirty-status** probe through nesting, and pass a **hardcoded `false`** for `upstream_gone`.
- Net effect: revert line 958 back to `probe_upstream_gone(&path)` and **the entire 269-test suite still passes green.** The regression that caused the disk-full strandings on continuum #1530/#1531 (branch deleted on origin + non-terminal card → must classify Removable via the upstream-gone trump card) is not pinned for the nested layout.

### Required to clear
Add one test that builds the `<parent>/src/` layout (as `probe_dirty_status_handles_nested_src_layout` already does), deletes the tracking branch on the fixture's origin, and asserts `probe_upstream_gone(&parent_resolved)` returns `true` — OR an end-to-end test that drives the resolved `effective` path into `probe_upstream_gone` and asserts `Disposition::Removable` with a non-terminal card state. The `git_fixture_with_upstream` helper + the nested-rename pattern already in the file make this ~15 lines; deleting the origin head is `git -C <origin.git> branch -D <branch>` or pushing a delete. Without it, line 958 has no guard.

## Minor (non-blocking)
- `cleanup_card_worktree` double-resolves: line 653 `resolve_worktree_path(&parent)` → `worktree_path`, then line 660 `probe_dirty_status(&worktree_path)` resolves again. Idempotent (`is_git_worktree` on an already-resolved `<short>/src` returns true → returned as-is), so no correctness issue — but worth a follow-up to call `probe_dirty_status_at` there for symmetry with the `run_cleanup` site.

Default BLOCK stands until the upstream-gone-through-nested-resolution regression test lands.

joelteply:
🔴 SENTINEL: BLOCK (re-review) — the regression net is still FAKE.

**Revert-makes-it-fail check: FAILED.** I locally reverted the production fix at `run_cleanup` (`crates/airc-cli/src/work_commands.rs:958`, `probe_upstream_gone(&effective)` → `probe_upstream_gone(&path)`) and ran the new test:

```
test work_commands::tests::probe_upstream_gone_follows_nested_src_resolve ... ok
test result: ok. 276 passed; 0 failed; 2 ignored
```

The new test — and all 276 — stay GREEN with the fix reverted. This is the exact failure mode from the prior re-review; the new commit did not close it.

**Root cause (structural, not flaky):** `probe_upstream_gone_follows_nested_src_resolve` never calls `run_cleanup`. It computes `let effective = resolve_worktree_path(&parent)` *inside the test body* (line 2827) and then asserts:
- `probe_upstream_gone(&effective)` == true  → tests the helper on an already-resolved path (never broken)
- `probe_upstream_gone(&parent)` == false   → asserts the bug *shape* on the helper (never the question)

The production decision the fix changed lives at `run_cleanup:958` — whether the **caller** passes the resolved `&effective` or the raw `&path`. The test re-resolves the path itself and bypasses that call site entirely, so the line-958 wiring is unobserved. It pins `probe_upstream_gone`'s correctness (which was already correct) instead of pinning that `run_cleanup` *routes through* the resolved path. A future refactor reverting line 958 sails through green — proven above.

**What a real net requires:** drive the actual `run_cleanup` path (or a thin extracted seam that takes only `&path` and internally resolves) against a `<parent>/src/` nested layout with a deleted upstream, and assert the produced `WorktreeClassification`/`Disposition` reflects `upstream_gone == true` (i.e. `Removable`). That classification only comes out right if line 958 uses `&effective`; reverting it must flip the disposition to non-removable. The assertion must observe `run_cleanup`'s output, not a path the test re-resolved.

**Also blocking (shared-validation breakage):** `cargo fmt -p airc-cli -- --check` FAILS (exit 1) on the new code — `crates/airc-cli/src/work_commands.rs:2793` (the `branch` let-binding in the new test isn't rustfmt-clean). This trips the precommit/CI fmt gate.

Other checks (for the record): `cargo check` pass · work_commands tests 276 pass · `cargo clippy --bin airc --tests -D warnings` clean.

Restored the line locally; no changes pushed. Fix the test to exercise `run_cleanup` (revert-must-fail) and run `cargo fmt`. BLOCK until both land.

joelteply:
🔴 SENTINEL: BLOCK (re-review #3) — substance fixed, one mechanical gate left

**The fake-test problem is FIXED. The revert-makes-it-fail check PASSED this time.**

Verified at head `89d751d` in an isolated detached worktree (own `CARGO_TARGET_DIR`):

**Decisive checks — all green:**
1. ✅ **Real seam extracted + production routes through it.** `classify_worktree_path(path, card_state)` (work_commands.rs:1134) does `resolve_worktree_path` → `probe_dirty_status_at(&effective)` → `probe_upstream_gone(&effective)` → `classify_worktree`. Production `run_cleanup` (line 939-940) actually CALLS it — `let (effective, dirty_status, disposition) = classify_worktree_path(&path, card_state.as_ref());`. No parallel copy; the prod path goes through the extracted fn.
2. ✅ **THE REVERT TEST PASSED (decisive).** In my worktree I reverted the helper's `probe_upstream_gone(&effective)` → `probe_upstream_gone(path)` (reintroducing the original bug, leaving dirty on `&effective` to isolate the probe). `classify_worktree_path_emits_removable_on_nested_deleted_upstream` went **RED**: `disposition=SkipUnknownCard` vs expected `Removable`, panic at line 2830. Restored — clean, no residual diff. **The net is real now: the test cannot stay green if the production probe stops getting the resolved path.** This is exactly what was fake in the prior two attempts.
3. ✅ Test builds the nested `<parent>/src/` layout (`git_fixture_with_upstream` renamed under `<parent>/src/`), deletes the upstream ref on origin, and asserts `Removable` through the production seam. `Removable` is only reachable via `upstream_gone==true` (classify_worktree:1161) which fires BEFORE the `card_state==None → SkipUnknownCard` branch — so the disposition genuinely IS the probe signal.
4. ✅ `cargo check` clean. ✅ `cargo clippy -p airc-cli` clean (0 warnings). ✅ all 61 `work_commands` tests pass.

**The one blocker (CI gate):**
🔴 **`cargo fmt --check -p airc-cli` exits 1** at work_commands.rs:**2798** — in the NEW test (`classify_worktree_path_emits_removable_on_nested_deleted_upstream`): the `git rev-parse` `.args([...])` and the `String::from_utf8(...).unwrap().trim().to_string()` chain need rustfmt's multi-line form. This is the same fmt gate that failed before (then at 2793). The new test code was added without running `cargo fmt`, so the gate is still red.

**Fix is mechanical:** `cargo fmt -p airc-cli` and push. Single diff, entirely within the new test block — nothing else in the file regressed. Once fmt is clean this is an APPROVE: the architecture is right and the regression net finally nets the production path.

PR #1199, opened 2026-06-14
fix(work): skip worktree spawn when card already has a linked PR

joelteply:
Rebased cherry-pick of dcf6cb5 + cff090b from closed PR #1111 onto fresh canary. Cherry-pick clean. 2 work_commands_git tests pass. Original #1111 was rebase-stuck via the rust-rewrite squash promote.

### DIFF
--- a/crates/airc-cli/src/work_commands_git.rs
+++ b/crates/airc-cli/src/work_commands_git.rs
@@ -29,6 +29,16 @@ pub(crate) async fn spawn_claim_worktree(
         .card(card_id)
         .ok_or_else(|| format!("card {card_id} not visible in board projection"))?;
     let short: String = card.card_id.to_string().chars().take(8).collect();
+
+    // Card 8a3082c4: skip worktree spawn entirely when the card is
+    // already linked to a PR — the work is by definition happening
+    // on the linked head branch elsewhere. See `worktree_skip_reason`
+    // for the rule + rationale.
+    if let Some(reason) = worktree_skip_reason(&card) {
+        println!("worktree:  skipped — {reason}");
+        return Ok(());
+    }
+
     let slug = slugify(&card.title, 40);
 
     let lease_root = lease::lease_root()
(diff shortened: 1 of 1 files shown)


joelteply:
🔴 SENTINEL: BLOCK

Mechanically clean — verified on the **actual PR head** (faa1803, after a stale-checkout false alarm): `cargo check -p airc-cli` ✅, `cargo fmt --check` ✅, `cargo clippy --bin airc` ✅ (0 warnings), and both new tests run + pass (`work_commands_git::tests::{skips_when_pull_request_is_linked, proceeds_when_no_pull_request_linked}`). Race (#1) and downstream-NPE (#2) concerns checked and cleared: the helper is read-only/best-effort, the claim is already authoritative before spawn, the result is only logged, and nothing downstream assumes the worktree exists. So those are fine.

**Blocking defect — the skip ignores PR merge state, and the tests give false confidence about it.**

`worktree_skip_reason` (`crates/airc-cli/src/work_commands_git.rs:231`) skips on `card.pull_request.is_some()` — purely *presence*, not *liveness*:

```rust
if let Some(ref pr) = card.pull_request { return Some(...); }
```

But `PullRequestRef` (`crates/airc-work/src/model.rs:97`) carries only `repo/number/head/base` — **no state**. Meanwhile the projection already tracks `PrMergeState { Open, Draft, Ready, Merged, Closed }` per PR in `PullRequestRecord.merge_state` (`projection.rs:140`), queryable via `board.pull_request(repo, number)` — and `board` is **already in scope** at the call site (`work_commands_git.rs:25`). So the live merge state is available and deliberately not consulted.

Consequences:
1. `card.pull_request` is **never cleared** once linked — only relinked (`projection.rs:180-192`; no clear path). A card whose PR was **merged or closed** keeps the ref forever.
2. `run_claim` (`work_commands.rs:178`) has **no guard** against claiming a merged/closed-PR card — verified: no `merge_state`/`PrMergeState` check in the claim path.
3. Therefore re-opening such a card for follow-up work and claiming it **silently skips the worktree forever** and prints a misleading `continue work in your existing checkout` hint pointing at a head branch whose PR is already gone — the exact "misleading hint" the doc says it's preventing, inverted.

`is_some()` is the wrong condition. The right rule is roughly: skip only when the linked PR is **Open/Draft/Ready** (work genuinely in flight); a **Merged/Closed** PR ref should fall through and re-spawn. That requires passing the board (or the resolved `PrMergeState`) into the helper — the `&WorkCard`-only signature, chosen "so tests don't need a daemon," is precisely what blinds it to merge state.

**Tests don't cover the load-bearing edge (concern #4).** The two tests exercise only `Some(open-ish PR)` → skip and `None` → proceed. Neither constructs a **merged/closed** linked PR. The PR doc calls `skips_when_pull_request_is_linked` "the load-bearing case" and invokes `[[every-error-is-an-opportunity-to-battle-harden]]`, but it battle-hardens only the happy path and leaves the case that produces a *wrong* skip both unhandled and untested — false confidence.

**To unblock:** distinguish PR state (resolve `merge_state` from the in-scope `board` and skip only for live PRs), and add a third test: merged/closed-PR-linked card → `None` (re-spawn). Until then this ships a forever-skip footgun on PR-recycled cards.

joelteply:

Re-reviewed at `6bf5170`. The presence-vs-liveness defect is correctly fixed.

**1. Liveness gate (verified)**
`spawn_claim_worktree` now resolves the linked PR's state from the in-scope `board` projection (defined at the top of the fn) and threads it into the pure helper:
```rust
let pr_merge_state = card.pull_request.as_ref().and_then(|pr| {
    board.pull_request(&pr.repo, pr.number).and_then(|record| record.merge_state)
});
if let Some(reason) = worktree_skip_reason(card, pr_merge_state) { ... return Ok(()); }
```
`board.pull_request(&RepoId, u64) -> Option<&PullRequestRecord>` and `record.merge_state: Option<PrMergeState>` (Copy) confirmed in `projection.rs`. The match: `Some(Merged)|Some(Closed) => None` (dead → re-spawn), everything else → skip. Open/Draft/Ready correctly stay LIVE → skip. The `is_some()` presence bug is gone — a merged/closed PR no longer pins the skip forever.

**2. New edge-case tests (verified present + passing)**
- `proceeds_when_linked_pr_is_merged` → asserts `.is_none()` (re-spawn)
- `proceeds_when_linked_pr_is_closed` → asserts `.is_none()` (re-spawn)
- `skips_conservatively_when_pr_state_unknown` (None → skip)
- plus the prior Open→skip and None-PR→proceed cases
5/5 pass.

**3. Unknown-state default**
`merge_state == None` (PR linked but projection hasn't indexed it) falls through to conservative **skip** — correct direction; won't clobber an unobserved in-flight branch. Draft treated as live (skip) — correct. Documented and test-pinned.

**4. Build/lint (isolated target dir)**
- `cargo check -p airc-cli` ✓
- `cargo clippy -p airc-cli --all-targets` ✓ zero warnings
- `cargo fmt -p airc-cli --check` ✓
- 5 new `work_commands_git::tests` ✓ pass

No remaining defect. Ship it.

PR #1200, opened 2026-06-14
fix(work): cleanup classifier proves Clean via patch-id when no upstream

joelteply:
Rebased cherry-pick of 50ec7ba from closed PR #1112 onto fresh canary. 51 work_commands tests pass. Original #1112 was rebase-stuck via the rust-rewrite squash promote.

### DIFF
--- a/crates/airc-cli/src/work_commands.rs
+++ b/crates/airc-cli/src/work_commands.rs
@@ -1493,42 +1493,64 @@ fn is_clean_via_cherry_against_origin_head(path_str: &str) -> DirtyStatus {
             .lines()
             .any(|line| line.trim_start().starts_with('+'));
         if !has_unique {
-            // BIGMAMA review fix on PR #1200: `git cherry` walks
-            // commits by patch-id, which SKIPS MERGE COMMITS entirely
-            // (they have multiple parents → no canonical patch-id).
-            // An "evil merge" — unique content living only in a merge
-            // commit's conflict-resolution diff — emits zero `+`
-            // lines, so `has_unique` reads false and we'd return
-            // Clean → cleanup DELETES branches whose unique work
-            // lives in a merge. Data loss.
+            // BIGMAMA review fix on PR #1200 (round 2): `git cherry`
+            // walks commits by patch-id, which SKIPS MERGE COMMITS
+            // entirely (they have multiple parents → no canonical
+            // patch-id). An "evil merge" — unique content living only
+            // in a merge commit's conflict-resolution diff — emits
+            // zero `+` lines, so `has_unique` reads false and we'd
+            // return Clean → cleanup DELETES branches whose unique
+            // work lives in a merge. Data loss.
             //
-            // Defense in depth: even when cherry says "all patches
-            // present", verify there's no CONTENT delta between
-            // candidate and HEAD. `git diff --quiet candidate..HEAD`
-            // walks trees regardless of commit structure: exit 0 = no
-            // content delta (truly Clean), exit 1 = content differs
-            // (evil merge case → refuse).
-            let diff_clean = std::process::Command::new("git")
+            // BIGMAMA caught the first defense attempt:
+            // `git diff --quiet candidate..HEAD` is an endpoint diff
+            // between two trees, NOT a "what's on HEAD that isn't on
+            // candidate" probe. In the normal squash-merge case
+            // candidate has advanced past HEAD (other PRs landed
+            // since this one's branch point), so the diff is
+            // non-empty and the check fires false-positive on every
+            // mergeable branch.
+            //
+            // Correct defense: enumerate the merge commits reachable
+            // from HEAD but NOT from candidate. `git rev-list
+            // candidate..HEAD --merges` returns exactly those. Cherry
+            // already vouched for the non-merge patches being present
+            // upstream; if there are no extra merge commits, every
+            // commit (non-merge AND merge) is accounted for and the
+            // branch is genuinely Clean. If ANY merge exists in that
+            // range, cherry's verdict is incomplete — refuse to
+            // delete rather than guess whether the merge brought in
+            // unique resolution content.
+            let extra_merges = std::process::Command::new("git")
                 .args([
                     "-C",
                     path_str,
-                    "diff",
-                    "--quiet",
+                    "rev-list",
+                    "--count",
+                    "--merges",
                     &format!("{candidate}..HEAD"),
                 ])
-                .status()
-                .map(|s| s.success())
-                .unwrap_or(false);
-            if diff_clean {
-                // Both proofs agree: every commit patch-id is in
-                // candidate AND there's no content delta. The
-                // branch's work is captured upstream; safe to delete.
+                .output()
+                .ok()
+                .and_then(|o| {
+                    if o.status.success() {
+                        String::from_utf8(o.stdout).ok()
+                    } else {
+                        None
+                    }
+                })
+                .and_then(|s| s.trim().parse::<u64>().ok());
+            if let Some(0) = extra_merges {
+                // Cherry says non-merge patches are upstream AND there
+                // are no extra merge commits to worry about. The
+                // branch's work is fully captured; safe to delete.
                 return DirtyStatus::Clean;
             }
-            // Cherry said clean but content differs → evil-merge
-            // shape. Keep iterating to other candidates (one might
-            // contain the real merge); if none does, fall through to
-            // the Dirty arm below.
+            // Either rev-list failed (treat as can't-prove) or there
+            // are merges on HEAD outside candidate (potential evil
+            // merge). Keep iterating — another candidate might
+            // include the merges; if none does, fall through to the
+            // Dirty arm below.
         }
     }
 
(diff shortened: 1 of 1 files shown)


joelteply:
🔴 SENTINEL: BLOCK

Build hygiene is green (on the real PR commit `9324886`, isolated target dir):
- `cargo check -p airc-cli` ✅
- `cargo test -p airc-cli --bins probe_dirty` ✅ 6/6 (incl. the two new no-upstream tests)
- `cargo clippy -p airc-cli` ✅ no warnings
- `cargo fmt -- --check` ✅

But there is a real **data-loss defect** in the new patch-id fallback. Blocking.

## 🔴 BLOCKER — `git cherry` / patch-id is blind to merge-commit content → false-Clean → worktree deleted → work lost

`is_clean_via_cherry_against_origin_head` (`crates/airc-cli/src/work_commands.rs:1374`) decides Clean by checking `git cherry <candidate> HEAD` for any `+` line (`:1416-1432`). `git cherry` (and patch-id generally) **skips merge commits entirely** — it only diffs non-merge commits. So unique content that lives ONLY in a merge commit's conflict-resolution diff (an "evil merge") is invisible: cherry emits only `-` lines, `has_unique == false`, and the function returns `DirtyStatus::Clean` (`:1429`). The cleanup path then deletes the worktree and that merge content is gone.

This is **not** theoretical for airc/continuum worktrees: an operator who pulls/merges `canary` into their feature branch to resolve conflicts before a PR, and hand-resolves a conflict, produces exactly this shape. Reproduced locally:

```
# feat1 + side1 are already on canary by patch-id; the merge commit
# adds a UNIQUE file `evil` that exists nowhere on canary
$ git cherry canary
- 27174ba...   (feat 1  — present upstream)
- d12d66d...   (side 1  — present upstream)
# merge commit NOT listed at all → has_unique=false → returns Clean
$ git cat-file -e canary:evil  →  evil NOT on canary  (LOST if deleted)
```

The probe this code *replaces* did NOT have this hole: `git rev-list --count @{u}..HEAD` counts merge commits, so the same branch returns count=3 (>0 → Dirty/safe). The new fallback is therefore a **safety regression** for any branch carrying a merge commit, even though it's stricter for the linear-history case.

### Required fix (pick one)
1. Before trusting the cherry result, gate on merge commits: if `git rev-list --merges <candidate>..HEAD` is non-empty, refuse (`Dirty`/`Unknown`) rather than return Clean — patch-id can't speak to merge content. OR
2. Replace the `+`-line heuristic with a content-level proof that includes merge commits, e.g. `git rev-list --count <candidate>..HEAD --not <all-candidates-merged-by-patch-id>` is insufficient; the simplest sound check is "`git diff <candidate>...HEAD` (or `git cherry` AND `rev-list --merges` empty) shows nothing unique."

## 🟠 Secondary findings (not independently blocking, fix alongside)
- **No test covers the merge-commit case.** The two new tests (`:2716`, `:2750`) cover linear no-unique (Clean) and linear unique (Dirty), but not the merge-only-unique false-Clean above, nor the multi-candidate disjunction. Add a regression test that builds a merge commit with unique resolution content and asserts it is NOT classified Clean.
- **Hardcoded global candidate list** (`origin/{HEAD,main,master,canary,rust-rewrite,develop}`, `:1396`) means Clean fires if work is captured in *any* of them. For the data-loss property this is safe (content genuinely present somewhere by patch-id is not lost), but it can mask "merged to the wrong base." Acceptable for cleanup's purpose; noting it so the disjunction is a conscious choice, not an accident.

## ✅ Things I tried to break but couldn't
- Empty/`--allow-empty` commits: cherry shows `+` (treated as unique) → correctly Dirty. No loss.
- Unrelated histories (orphan candidate): all commits show `+` → correctly Dirty.
- patch-id accidental collision across different content: not reproducible; SHA1 patch-id doesn't collide on distinct diffs.
- Dirty-WIP guard: Step 1 porcelain probe (`:1268`) still runs first and unchanged — uncommitted/untracked still → Dirty. (Note: this is exactly why the merge case slips — committed merge content passes porcelain, then patch-id is blind to it.)

Fix the merge-commit blindness + add the regression test and I'll re-review.

joelteply:
🔴 SENTINEL: BLOCK (re-review of `9b3611a` evil-merge defense)

**Data-loss hole: closed. But the fix swaps it for a regression that defeats the PR's own original purpose, and ships zero tests for either path.**

### 1. The evil-merge hole IS closed (good)
`git diff --quiet candidate..HEAD` walks trees regardless of commit structure, so when HEAD carries unique content (incl. a merge's conflict-resolution diff), it exits 1 → `diff_clean=false` → does NOT return Clean → falls through to Dirty. Verified: evil-merge with unique `secret.txt` vs candidate → `diff --quiet` exit **1** → refused. Safe.

### 2. NEW over-conservatism regression (BLOCKING) — strands the common merged-then-advanced branch
`git diff --quiet candidate..HEAD` is `git diff candidate HEAD` — a **full bidirectional tree comparison**. It exits 1 not only when HEAD has extra content, but also when **candidate has extra content HEAD lacks** — i.e. whenever the integration branch advanced past the merge point. That is the *normal* state by the time cleanup runs.

Reproduced through a fixture-equivalent repo (bare origin + clone, `origin/HEAD→main`), the exact scenario `c4d4bbf` was written to fix:
```
workbranch's feature IS captured on origin/main (patch-id), origin/main then advanced with later.txt
  cherry origin/main           : [- <sha>]   (no '+'  => correctly "captured upstream")
  diff --quiet origin/main..HEAD: exit 1      => NEW code REFUSES Clean => worktree stranded forever
  rev-list --merges origin/main..HEAD: []     => reviewer's fix: Clean (correct)
```
So after any normal merge-and-advance, cleanup stops reclaiming branches — re-opening the 12–19 GB/orphan leak that `c4d4bbf` (this very PR) set out to fix. The defense trades a rare data-loss bug for a routine "cleanup never works again" bug.

### 3. Reviewer's suggested primitive was correct; `diff --quiet` is the wrong one
`git rev-list --merges candidate..HEAD` (non-empty ⇒ refuse) is correct on all three properties; `git diff --quiet` is wrong on Property C:

| case | want | rev-list --merges (suggested) | diff --quiet (shipped) |
|---|---|---|---|
| A evil merge, unique content | refuse Clean | merges non-empty → refuse ✅ | exit 1 → refuse ✅ |
| B clean linear, squash-merged | Clean | merges empty → Clean ✅ | exit 0 → Clean ✅ |
| C merged then candidate advanced | Clean | merges empty → Clean ✅ | **exit 1 → refuse ❌** |

(If you want belt-and-suspenders beyond merges, the directional primitive is "is anything reachable from HEAD absent from candidate", e.g. `git rev-list candidate..HEAD` non-empty after the cherry already vouched for patch-ids — NOT a symmetric `git diff candidate HEAD`.)

### 4. Zero tests (BLOCKING per repo test discipline)
`9b3611a` touches only `work_commands.rs` and adds **no `#[test]`**. There is no git-level test that builds a real merge commit and runs `is_clean_via_cherry_against_origin_head` — so neither the data-loss case it claims to fix nor the Property-C regression it introduces is covered. The commit's "51 tests pass" is vacuous: `classifier_merged_clean_removes` tests the pure enum router with `DirtyStatus::Clean` handed in, and `..._head_already_captured` uses HEAD==origin/HEAD (no delta either direction) so it can't catch C. A required test: bare-origin fixture, feature captured upstream + integration advanced, assert `probe_dirty_status == Clean`; plus an evil-merge fixture asserting `Dirty`.

### Build/lint (isolated target dir)
cargo check -p airc-cli ✅ · 56 work_commands tests pass ✅ · clippy --all-targets clean ✅ · fmt --check clean ✅ (all green because the regression path has no test).

**Required to unblock:** replace `git diff --quiet candidate..HEAD` with `git rev-list --merges candidate..HEAD` non-empty ⇒ refuse (the reviewer's primitive); add the two fixture tests above (Property A → Dirty, Property C → Clean). Default BLOCK until the merged-then-advanced case classifies Clean again *and* the evil-merge case is test-covered.

joelteply:
🔴 SENTINEL: BLOCK (re-review r3)

**Algorithm is now CORRECT on all three cases — but the PR fails the `cargo fmt --check` CI gate. Default-BLOCK stands until fmt is clean.**

Reviewed in an isolated detached worktree at PR head `17ebf39` (`fix(work): rev-list --merges defense vs symmetric diff`), isolated CARGO_TARGET_DIR. Diff scoped to `crates/airc-cli/src/work_commands.rs`.

### ✅ Core logic verified empirically with fixture repos (all three cases)

The `git rev-list --count --merges {candidate}..HEAD` primitive is the correct replacement for the symmetric `git diff --quiet`. Reproduced each case:

1. **EVIL-MERGE** (merge commit on HEAD not on candidate, carrying potential unique resolution content): `merges = 1` → `extra_merges = Some(1) ≠ Some(0)` → does NOT return Clean → outer arm returns **Dirty**. Data loss prevented. ✓
2. **CLEAN LINEAR** (feature upstreamed by patch-id, candidate advanced): cherry shows only `-`, `merges = 0` → **Clean**. No false-positive stall. ✓
3. **MERGED-THEN-ADVANCED** (the v2 regression): Directly demonstrated that the OLD `git diff --quiet origin/main..HEAD` returns **non-empty → false Dirty** on this exact state, while the NEW primitive returns **0 merges → Clean**. The regression that re-stranded worktrees (12–19GB leak) is fixed. ✓

**Range semantics confirmed correct** (the load-bearing detail): `{candidate}..HEAD` counts merges on the FEATURE side only. Verified that when the *base* advances via its own merge commit, `candidate..HEAD --merges = 0` (Clean), whereas the flipped `HEAD..candidate` would wrongly count it = 1. The code uses the correct direction.

**Conservative None-handling correct:** `if let Some(0) = extra_merges` returns Clean only on a proven-zero count; rev-list failure (`None`) and `Some(n>0)` both fall through to Dirty.

### ✅ The two previously-untested cases now have tests (both pass)
- `is_clean_via_cherry_returns_clean_when_squash_merged_and_candidate_advanced` (merged-then-advanced → Clean)
- `is_clean_via_cherry_refuses_when_extra_merge_lives_in_head` (evil-merge → Dirty)

`cargo test -p airc-cli --bin airc -- probe_dirty is_clean_via_cherry`: **8 passed, 0 failed.**

### ✅ cargo check — clean.  ✅ cargo clippy — clean, zero warnings.

### 🔴 BLOCKER: `cargo fmt --all -- --check` FAILS (exit 1)
6 rustfmt violations, ALL in the new test code in `crates/airc-cli/src/work_commands.rs` (lines ~2886, 2910, 2930, 2995, 3013, 3034) — over-expanded multi-line `run(&[...], None)` calls that rustfmt collapses to single lines. CI's `fmt` job (`.github/workflows/rust.yml:126`, `cargo fmt --all -- --check`) is a hard gate; this PR will fail it. Per the repo's non-negotiable clean-build discipline (and never bypassing the shared validation path), this blocks merge.

**To clear:** run `cargo fmt -p airc-cli` and commit the result. No logic change required — once fmt is clean this is an **APPROVE**.

Issue #1, opened 2026-04-14
Identity collision when ~/.agent-relay is cloned/synced between hosts


joelteply:
## Problem

When \`~/.agent-relay\` (the directory holding the Ed25519 identity) is copied, rsync'd, cloud-synced (Dropbox/iCloud), or cloned between two hosts — say, migrating a laptop to a new machine, or provisioning a second dev box from a backup — both hosts end up with the **same peer identity**.

Effects we've hit:
- Two Claude-Code instances on different machines (an M1 Pro and an M5 Pro) connected to the same relay host and silently stepped on each other's messages: one would be "connected," the other would appear offline or its sends would race.
- The fix is discoverable but non-obvious: \`AGENT_RELAY_HOME=/tmp/relay-<unique>\` before \`relay connect\`. A third Claude instance (\`vhsm\`) hit this cold and had to reverse-engineer the workaround before it could join.
- No warning surfaced on init, no log line, no "this identity was generated on host X, you are host Y, continue anyway?" prompt.

## Why this happens

Identity is purely in the filesystem under \`~/.agent-relay\`. Anything that copies home directories (backup restore, container image reuse, Dropbox-style sync, \`rsync -a ~\`) duplicates it.

## Proposed fix

Embed host metadata in the identity file when it's created, and check it on every \`relay\` invocation:

1. On identity generation, write alongside the keys:
   - \`hostname\` (at creation time)
   - \`os.hostid()\` / machine-id where available (Linux \`/etc/machine-id\`, macOS \`IOPlatformUUID\`, Windows \`MachineGuid\`)
   - \`created_at\` timestamp
2. On \`relay\` startup (connect/send/monitor/peers), read that metadata. If the current host's hostid/hostname doesn't match:
   - Default: refuse to start. Print a clear message explaining the clash, show the saved metadata vs. the current host, and offer three options: (a) \`relay identity regenerate\` to create a fresh identity on this host, (b) \`AGENT_RELAY_HOME=...\` to point elsewhere, (c) \`--force-identity\` flag for power users who really do want to impersonate the other host.
   - The error should include the exact next-step command — no "figure out the workaround" tax.
3. Optional: add a \`relay identity show\` / \`relay identity status\` subcommand that prints the current identity's fingerprint + host-of-origin, so users can audit.

## Why the hostid-based check (not just hostname)

[...]

PR #1201, opened 2026-06-14
fix(transport): daemon advertises BOTH LAN and Tailscale endpoints (ladder: LAN-first, Tailscale only off-LAN)

joelteply:
Rebased cherry-pick of 303e53c from closed PR #1157 onto fresh canary. Cherry-pick was clean. 6 stored_endpoint_dial tests pass (peer_dials_lan_rung_and_skips_tailscale + dial_walks_endpoints_in_stored_order_and_stops_on_success are the load-bearing ones). Original #1157 was rebase-stuck via the rust-rewrite squash promote.

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -1301,7 +1301,8 @@ pub async fn run_daemon(
                         .join(" + ");
                     eprintln!(
                         "airc daemon: advertising {summary} in the account registry \
-                         (LAN dialed first, Tailscale only if peers leave the LAN)"
+                         (LAN dialed first; off-LAN peers fall through to Tailscale \
+                         after a ~3s LAN-rung timeout, once per session)"
                     );
                 }
                 Err(error) => {
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -1251,36 +1251,51 @@ pub async fn run_daemon(
         // routable LAN (or a bind failure) still reaches the mesh by
         // dialing OUT to listening peers / relay — it just isn't
         // dialable itself, which we say plainly rather than swallow.
-        // Prefer the Tailscale endpoint over LAN. Every node on the same
-        // gh account is on the Tailscale mesh, so a 100.x address is
-        // dialable from ANY network and traverses NAT/firewalls, whereas a
-        // 192.168.x LAN address only works same-subnet and dies behind a
-        // firewall (the cross-machine keystone failure: nodes advertised
-        // LAN-only and could enrol each other but never connect). Fall back
-        // to LAN when Tailscale is down.
-        let advertise = crate::network_commands::detect_tailscale_ip()
-            .map(|ip| (ip, "Tailscale"))
-            .or_else(|| crate::network_commands::detect_lan_ip().map(|ip| (ip, "LAN")));
-        match advertise {
-            Some((ip, kind)) => match airc.listen_lan(std::net::SocketAddr::from((ip, 0))).await {
-                Ok(addr) => {
+        // Advertise BOTH the LAN and Tailscale addresses (the connection
+        // ladder: local → LAN → Tailscale → greater-grid P2P). Tailscale
+        // is unnecessary for same-subnet peers — a 192.168.x dial is
+        // direct, while routing same-LAN traffic over 100.x is a wasted
+        // hop. So we publish the LAN address (dialed first, no hop) AND
+        // the Tailscale address (the NAT-traversing fallback dialed only
+        // when the LAN dial times out, i.e. when peers are on different
+        // networks). One wildcard listener serves both interfaces; the
+        // endpoint sort order (LanTcp before TailscaleTcp) makes the
+        // dialer try LAN first and break on success — Tailscale only if
+        // we leave the LAN. Earlier this preferred Tailscale exclusively,
+        // forcing every same-LAN peer through an unnecessary 100.x hop.
+        let lan_ip = crate::network_commands::detect_lan_ip();
+        let tailscale_ip = crate::network_commands::detect_tailscale_ip();
+        if lan_ip.is_none() && tailscale_ip.is_none() {
+            eprintln!(
+                "airc daemon: no routable LAN or Tailscale IPv4 detected — account beacon \
+                 carries no endpoint (outbound-dial / relay only)"
+            );
+        } else {
+            match airc.listen_lan_advertising(lan_ip, tailscale_ip).await {
+                Ok(endpoints) => {
+                    let summary = endpoints
+                        .iter()
+                        .map(|endpoint| match endpoint {
+                            airc_lib::RouteEndpoint::LanTcp { addr } => format!("LAN {addr}"),
+                            airc_lib::RouteEndpoint::TailscaleTcp { addr } => {
+                                format!("Tailscale {addr}")
+                            }
+                            other => format!("{other:?}"),
+                        })
+                        .collect::<Vec<_>>()
+                        .join(" + ");
                     eprintln!(
-                        "airc daemon: advertising {kind} endpoint {addr} in the account registry"
+                        "airc daemon: advertising {summary} in the account registry \
+                         (LAN dialed first, Tailscale only if peers leave the LAN)"
                     );
                 }
                 Err(error) => {
                     eprintln!(
-                        "airc daemon: {kind} listener bind failed ({error}) — account beacon \
-                             carries no endpoint; this node reaches the mesh by dialing out / \
-                             relay but is not itself dialable"
+                        "airc daemon: LAN listener bind failed ({error}) — account beacon \
+                         carries no endpoint; this node reaches the mesh by dialing out / \
+                         relay but is not itself dialable"
                     );
                 }
-            },
-            None => {
-                eprintln!(
-                    "airc daemon: no Tailscale or routable LAN IPv4 detected — account beacon \
-                     carries no endpoint (outbound-dial / relay only)"
-                );
             }
         }
         // Card 4b6a0ffa (#33): record the endpoints this handle now
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -1256,13 +1256,28 @@ pub async fn run_daemon(
         // is unnecessary for same-subnet peers — a 192.168.x dial is
         // direct, while routing same-LAN traffic over 100.x is a wasted
         // hop. So we publish the LAN address (dialed first, no hop) AND
-        // the Tailscale address (the NAT-traversing fallback dialed only
-        // when the LAN dial times out, i.e. when peers are on different
-        // networks). One wildcard listener serves both interfaces; the
-        // endpoint sort order (LanTcp before TailscaleTcp) makes the
-        // dialer try LAN first and break on success — Tailscale only if
-        // we leave the LAN. Earlier this preferred Tailscale exclusively,
-        // forcing every same-LAN peer through an unnecessary 100.x hop.
+        // the Tailscale address (the NAT-traversing fallback). One
+        // wildcard listener serves both interfaces; the endpoint sort
+        // order (LanTcp before TailscaleTcp) makes the dialer try LAN
+        // first and break on success.
+        //
+        // BIGMAMA review BLOCKING-2 on PR #1201 — honest cost note:
+        // the dialer (`airc_lib::discovery::dial_stored_peer_endpoints`)
+        // walks the merged endpoint list in unconditional
+        // `RouteEndpointKind` order for EVERY peer — there is no
+        // same-subnet/reachability gate. Off-LAN peers (i.e. on the same
+        // tailnet but a different physical LAN) therefore dial the
+        // publisher's unreachable 192.168.x rung FIRST and pay up to
+        // `PEER_DIAL_TIMEOUT` (3s) before falling through to the live
+        // Tailscale rung. Earlier this preferred Tailscale exclusively,
+        // forcing every same-LAN peer through an unnecessary 100.x hop;
+        // the new "LAN-first, 3s timeout, then Tailscale" cost is
+        // intentional (same-LAN wins outweighs the off-LAN 3s) and
+        // pinned by the dead-LAN/live-Tailscale test in
+        // `airc-lib/tests/stored_endpoint_dial.rs`. A future real fix
+        // (subnet/reachability gate at the dialer) will eliminate the
+        // 3s for off-LAN peers; until then the truth is visible in the
+        // recorded peer_dial_failures, not hidden in comments.
         let lan_ip = crate::network_commands::detect_lan_ip();
         let tailscale_ip = crate::network_commands::detect_tailscale_ip();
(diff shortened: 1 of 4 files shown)


joelteply:
🔴 SENTINEL: BLOCK

Build hygiene is clean (cargo check, fmt --check, clippy --all-targets, and all 6 `stored_endpoint_dial` tests pass on isolated target dir). The block is **semantic**: this PR reintroduces the exact Docker-ghost endpoint that canary's Tailscale-preference deliberately suppressed, and the "Tailscale only off-LAN" ladder claim is not actually implemented at the dialer.

**BLOCKING-1 — Regression: Docker `172.18.0.x` is now published to the account beacon (the flood the context describes).**
`detect_lan_ip()` (`crates/airc-cli/src/network_commands.rs:15`) uses the route-to-8.8.8.8 source-address trick and filters only loopback/unspecified (`is_routable_lan_ipv4`, line 24). Inside a Docker container that returns the **bridge IP `172.18.0.x`**, which passes the filter and gets advertised as a dialable `LanTcp` endpoint by `listen_lan_advertising` (`crates/airc-lib/src/lan.rs:69`).

This is a regression vs canary. On canary (`commands.rs:1261`) the advertise expression was `detect_tailscale_ip().or_else(detect_lan_ip)` — when Tailscale is up (the running cross-machine daemon), `detect_lan_ip()` was **never called**, so the Docker `172.18.x` address was never published. After this PR (`commands.rs:1264-1266`) `detect_lan_ip()` is **always** called and its result is published alongside Tailscale. The codebase elsewhere treats `172.18.0.x` as ghost pollution worth a dedicated GC verb (`cli.rs:1339` `Prune`, `peers.rs:46`). The PR docstring claims it "publishes only specific, dialable IPs" — a container bridge IP is neither dialable off-box nor should it be advertised. This PR is a direct cause of the `peer_dial_failed` flood on `172.18.0.x`, not a fix for it.

**BLOCKING-2 — "Tailscale only if peers leave the LAN" is asserted in comments but never determined per-dialer.**
The ladder claim relies on "a same-subnet peer dials LAN, an off-LAN peer falls through to Tailscale." But the daemon publishes ONE beacon to the account-registry gist visible to **all** account peers regardless of physical LAN, and the dialer (`crates/airc-lib/src/route/discovery.rs:170-216`) walks the merged endpoint list in global `RouteEndpointKind` order (LanTcp first) for **every** peer — there is no same-subnet check anywhere. So an off-LAN peer on the same Tailnet imports both endpoints and dials the publisher's unreachable `192.168.x`/`172.18.x` LAN rung FIRST, eating the full `PEER_DIAL_TIMEOUT` = **3s** (`discovery.rs:22`) before falling through to the live Tailscale rung. "LAN-first" is unconditional, not "LAN-first-if-same-subnet." Off-LAN peers pay 3s per discovery refresh, per dead LAN rung. (Q1 + Q2.)

**BLOCKING-3 — Test coverage omits the only case that matters.**
`peer_dials_lan_rung_and_skips_tailscale` (`tests/stored_endpoint_dial.rs:184`) pins the **happy path**: LAN reachable (loopback), Tailscale unreachable, zero failures. The realistic off-LAN case — **LAN rung DEAD, Tailscale rung LIVE** — is never exercised. That is exactly where the 3s penalty lives and where a real defect (does it actually fall through and connect via Tailscale, and is the LAN timeout recorded?) would surface. There is also no test for the Docker-bridge-IP advertisement path, no LAN-only/Tailscale-only advertise coverage in `lan.rs` (only the both-present case), and no off-LAN trust-tier/leak test. (Q5.)

**NON-BLOCKING (note):** Q4 single-homed regression check is OK — LAN-only and Tailscale-only hosts advertise the same single endpoint as before; only the dual-homed case changes (and that change is BLOCKING-1).

**Required to lift the block (any one closes the worst of it):**
1. Filter `detect_lan_ip()` against RFC1918-vs-container ranges, or at minimum exclude the Docker bridge range, so the daemon never advertises a `172.18.0.x` LAN endpoint. This is the fix that actually addresses the stated flood.
2. Add a test where the LAN rung is dead and Tailscale is live: assert it connects via Tailscale AND assert/measure the recorded LAN-dial timeout, so the 3s off-LAN penalty is visible and intentional, not accidental.
3. Either honor the "Tailscale only off-LAN" claim with a real reachability/subnet gate, or correct the comments+docstrings to state the truth: both endpoints are published to all peers and off-LAN peers pay a LAN-dial timeout before Tailscale.

joelteply:
🔴 SENTINEL: BLOCK (re-review)

Re-review of `f373c81` ("detect_lan_ip skips publishing inside containers"). Build hygiene clean on isolated target dir: `cargo check -p airc-lib -p airc-cli` ✅, `cargo clippy` ✅ (0 warnings), `cargo fmt --check` ✅, `stored_endpoint_dial` 6/6 ✅, `airc-cli` bin unit tests (incl. `in_container_returns_false_on_real_host`) 3/3 ✅, `airc-lib` route tests 26/26 ✅.

**Per-finding status:**

**BLOCKING-1 — RESOLVED (for the actual flood source).** `detect_lan_ip()` now early-returns `None` when `in_container()` (`/.dockerenv` OR `/proc/1/cgroup` mentions docker/containerd/kubepods). The prior block's required-fix #1 explicitly accepted "exclude the Docker bridge range" as sufficient; container-detection achieves that for the real flood source — daemons running INSIDE Docker grid nodes — so the 172.18.0.x bridge IP is no longer advertised to the account beacon. Verified the filter does NOT over-restrict legitimate RFC1918: 192.168.1.232 → published, 10.0.0.5 → published, 127.0.0.1 → rejected (loopback), and a real (non-container) corporate 172.16/12 host still publishes. False-positive (`/.dockerenv` on a real host) degrades gracefully to Tailscale. `detect_lan_ip` is the SOLE auto-advertise producer (the other `listen_lan` caller, `lan-listen`, takes an explicit operator bind addr) — no second leak path. ✅
  - Minor (non-blocking) nuance: the gate is container-shaped, not range-shaped, so a pathological non-container host whose OS primary IP literally falls in 172.18.0.x would still publish it. Not a grid reality; noted only for completeness.

**BLOCKING-2 — REMAINS (and is a genuine THIS-PR regression, not pre-existing).** `discovery.rs` was not touched: the dialer (`discovery.rs:172-216`) still walks the merged endpoint list in unconditional `RouteEndpointKind` order (LanTcp first) for EVERY peer with NO same-subnet/reachability gate, and only `break`s on a *successful* connect. Confirmed against canary: on canary, when Tailscale is up (the running cross-machine daemon), `detect_lan_ip()` was NEVER called (`detect_tailscale_ip().or_else(detect_lan_ip)`), so off-LAN peers imported a Tailscale rung ONLY — nothing to time out on. This PR now publishes the LAN rung ALSO, to all account peers including off-LAN/cross-network ones on the same tailnet. A dual-homed host (e.g. the Intel Mac / BIGMAMA: 192.168.x LAN + 100.x Tailscale — the canonical cross-machine keystone topology) now advertises its 192.168.x to off-LAN peers, who dial it FIRST and eat the full `PEER_DIAL_TIMEOUT` = 3s (`discovery.rs:22`) before falling through to the live Tailscale rung. The container-skip fix does not touch this — it only suppresses the LAN rung inside containers, not on real dual-homed hosts. The docstrings still frame "Tailscale only off-LAN" as design while the mechanism is actually "LAN-first, fall through after a 3s timeout." 🔴

**BLOCKING-3 — PARTIALLY RESOLVED.** Added: `in_container_returns_false_on_real_host` (negative case only; the positive in-container path is acknowledged-untestable without FS mocking — fair). Still MISSING required-fix #2: a test where the LAN rung is DEAD and the Tailscale rung is LIVE, asserting (a) the peer still connects via Tailscale and (b) the 3s LAN-dial timeout is recorded/measured — i.e. the exact off-LAN penalty from BLOCKING-2 made visible and intentional. The existing `peer_dials_lan_rung_and_skips_tailscale` still pins only the happy path (LAN=loopback reachable, breaks before Tailscale). 🟡

**Q4 single-homed:** still OK — LAN-only and Tailscale-only hosts advertise one endpoint as before; no regression there.

**Verdict:** BLOCKING-1 (the regression you flagged as worst) is fixed. But BLOCKING-2 is itself a NEW regression this PR introduces for the cross-machine off-LAN topology airc exists to serve, and it is fully untouched. **To lift the block, do ONE of:**
1. Gate the LAN-rung dial on same-subnet reachability so off-LAN peers skip the dead LAN rung (the real fix), OR
2. Don't advertise the LAN endpoint when Tailscale is also up to peers that can't use it — or accept the 3s and PROVE it with the dead-LAN/live-Tailscale test (required-fix #2) so the cost is visible and intentional, and correct the docstrings to say "LAN-first, 3s timeout, then Tailscale" rather than implying a subnet gate that does not exist.

Build is green; the block is semantic — the same off-LAN ladder claim is still asserted in comments but not enforced at the dialer, and now newly costs off-LAN peers 3s where canary cost them nothing.

joelteply:
🔴 SENTINEL: BLOCK (re-review)

Re-review of `fix/advertise-both-lan-tailscale-v2` @ 75a847a on isolated detached worktree. Single, narrow blocker — the *architecture* of path (b) is good; the *deliverable* is red on Windows.

## What's resolved ✅
- **BLOCKING-1** (Docker 172.18.0.x leak): intact. `in_container()` probe (`/.dockerenv` + `/proc/1/cgroup`) gates `detect_lan_ip()`; `is_routable_lan_ipv4` still admits real 192.168.x and 172.16/12 corporate LANs — no over-filtering. Not touched by this commit.
- **Path (b) substance**: The docstring at `run_daemon` is now *honest* — it states plainly that the dialer walks endpoints in unconditional `RouteEndpointKind` order with NO subnet gate, that off-LAN peers dial the dead 192.168.x rung FIRST and pay up to `PEER_DIAL_TIMEOUT` (3s) before Tailscale, that this is intentional (same-LAN wins outweigh the off-LAN 3s), visible in `peer_dial_failures`, with a same-subnet gate noted as the real follow-up fix. The reasoning is sound, not hand-waving.
- **Rung-ordering genuinely exercised**: `off_lan_peer_pays...` stores `[LanTcp{dead}, TailscaleTcp{alive}]` and the dialer's own iteration order is what drives the fall-through. It asserts: connects via Tailscale, exactly one failure recorded, that failure IS the LAN rung, error non-empty. That's the real dialer, not a stand-in. Honest about not simulating the literal 3s SYN-drop (uses ECONNREFUSED).
- **Judgment on the cost**: 3s once-per-peer-per-session (the dial loop `break`s on first success and the route is then live for the session — not per-message) IS acceptable vs blocking on the subnet gate. I would NOT block on the documented tradeoff alone.

## What still blocks 🔴
**The newly-added test fails reproducibly on Windows.** The commit message claims "7/7 stored_endpoint_dial tests green" — that was macOS/Linux. On a `[self-hosted, windows]` runner (a first-class, in-matrix CI leg — `ci.yml` / `rust.yml` explicitly include it and state "the substrate must run on Linux, macOS, and Windows"):

```
test off_lan_peer_pays_lan_dial_timeout_then_connects_via_tailscale ... FAILED
panicked at stored_endpoint_dial.rs:429:
  ECONNREFUSED returns instantly; if this exceeds 2s ... took 2.0317043s
```
Deterministic across runs (2.03s, 2.05s). Root cause: the test's premise — "a connect to a bound-then-closed loopback port gets instant ECONNREFUSED" — is **false on Windows**. Windows' TCP stack retries the SYN and the connect hangs ~2.0s before failing, tripping the `elapsed < Duration::from_secs(2)` ceiling at line 429. The other 6 tests pass; only this new one is red.

So the "prove" half of prove+document doesn't currently prove on Windows — it breaks the Windows CI leg. A regression test that's red on a supported platform is a net-negative, regardless of how good the prose is.

### To lift the block (pick one)
1. Make the timing assertion platform-honest: gate it `#[cfg(not(windows))]`, or raise the ceiling to `< PEER_DIAL_TIMEOUT` (3s) and reframe the comment (on Windows the refused-dial is NOT instant — the assertion as written encodes a Unix-only assumption). The other four assertions (connects-via-Tailscale / exactly-LAN-failed / failure-is-LAN-rung / error-non-empty) are the load-bearing pins and pass everywhere — keep those unconditional.
2. Or drop the wall-clock timing assertion entirely — it's the weakest pin (the test already admits it can't simulate the real 3s SYN-drop), and the rung-ordering + visibility assertions are what actually prove path (b).

Minor (non-blocking, fix while here): the in-code comment at line 377-378 says "uses 250ms as a sanity bound" but the assertion is 2s — stale comment. And the operator log line still reads "Tailscale only if peers leave the LAN," mildly rosier than the honest code comment two lines up.

cargo check ✅ · clippy (airc-lib+airc-cli, --tests) ✅ · fmt ✅ · stored_endpoint_dial 6/7 (the new test ❌ on Windows). Verified on detached worktree @ 75a847a, isolated target dir.

joelteply:

**BLOCKING-2 (Windows flake) is FIXED. Test PASSES ON WINDOWS.** Decisive check run on a real Windows host (the platform that produced the prior failure):

- `cargo test -p airc-lib off_lan_peer_pays_lan_dial_timeout` → **4/4 green** across repeated runs.
- Full `stored_endpoint_dial` suite → **7/7 green**.

**1. The fix.** Wall-clock ceiling raised from `< 2s` → `< 3s` = `PEER_DIAL_TIMEOUT` (verified `const PEER_DIAL_TIMEOUT = Duration::from_secs(3)` in `route/discovery.rs:22`, where the per-dial `tokio::time::timeout(PEER_DIAL_TIMEOUT, connect_lan(...))` lives). Windows' ~2.0s SYN-retry on a closed port now sits comfortably under the bound, and the dial is timeout-wrapped so it structurally cannot exceed 3s regardless of kernel. This is the correct universal bound — it still catches a real stall (a rung not honoring its per-dial deadline) without papering over kernel-specific closed-port behavior. Good call keeping a meaningful wall-clock sanity assert rather than dropping it.

**2. Not weakened.** All 4 load-bearing asserts intact and meaningful: connects-via-Tailscale (`connected_lan_peers.contains`), exactly-one-LAN-failure (`peer_dial_failures.len() == 1`), failure-is-LAN-rung (`failure.endpoint == LanTcp{dead_addr}`), error-non-empty (`!failure.error.is_empty()`). The bound is `<3s`, not a vacuous `<60s` — it bounds something useful.

**3. BLOCKING-1 still resolved.** `in_container()` probe intact (`network_commands.rs:55`), gates Docker bridge IP so no `172.x` ghost is advertised; `in_container_returns_false_on_real_host` passes. `listen_lan_advertising` only advertises explicitly-passed dialable IPs, never the `0.0.0.0` wildcard bind.

**4. Nits addressed (non-blocking).** Stale `250ms` comment gone, replaced with honest cross-kernel docstring. Rosy daemon log line synced to truth ("LAN dialed first; off-LAN peers fall through to Tailscale after a ~3s LAN-rung timeout, once per session").

**5. Full gate (all on Windows):** `cargo fmt --check` clean · `cargo check -p airc-lib -p airc-cli` clean · `cargo clippy --tests -D warnings` clean · `airc-cli` suite 278 + all integration tests passing · `stored_endpoint_dial` 7/7.

No remaining blockers. Reviewed at PR head 9b10021 in an isolated detached worktree.

PR #1202, opened 2026-06-14
feat(peer): airc peer prune --stale-after-hours (operator override)

joelteply:
## Seam #3.2 follow-up — operator override for the prune staleness window

Closes the review nit from #1195: `DEFAULT_PEER_STALE_AFTER_MS` (1h grace) was hardcoded with no operator override. Adds `--stale-after-hours <N>` to `airc peer prune`:

- **omit** → 1h substrate default (behavior unchanged)
- **`N`** → grace window of N hours
- **`0`** → no grace (evict every absent untrusted peer immediately)

A **CLI flag, not an env var** — substrate thresholds stay explicit, not ambiently tuned (per the concurrency style guide). The active window prints in the plan (`staleness grace = N hour(s) (default|operator override)`) so the operator sees what gate was applied.

Resolution extracted into a pure `resolve_stale_after_ms` helper + unit test: `None`→default, `Some(2)`→7_200_000, `Some(0)`→0, `Some(MAX)` saturates (no overflow panic).

### Validation
`cargo check/test/clippy/fmt -p airc-cli` clean.

### DIFF
--- a/crates/airc-cli/src/cli.rs
+++ b/crates/airc-cli/src/cli.rs
@@ -1349,6 +1349,14 @@ pub enum PeerAction {
         /// only prints what it WOULD evict (dry run).
         #[arg(long)]
         apply: bool,
+        /// Staleness grace window, in hours: an untrusted peer absent
+        /// from the fresh registry is evicted only once its last_seen is
+        /// older than this. Recently-contacted peers inside the window
+        /// are kept (a momentary registry-snapshot lag is not death).
+        /// Omit for the 1-hour default; `0` evicts every absent untrusted
+        /// peer immediately (no grace).
+        #[arg(long)]
+        stale_after_hours: Option<u64>,
     },
 }
 
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -1874,7 +1874,25 @@ pub async fn run_peer_remove(
 /// unauthenticated, hermetic scope, unreachable rendezvous, or an empty
 /// registry), this prunes NOTHING — pruning against an unknown/empty live
 /// set would wrongly evict live peers.
-pub async fn run_peer_prune(home: &Path, apply: bool) -> Result<(), Box<dyn std::error::Error>> {
+/// Seam #3.2 operator override for the peer-prune staleness grace
+/// window. Omit (`None`) → the 1h substrate default; an explicit value
+/// (incl. `0` = no grace, evict every absent untrusted peer) tunes how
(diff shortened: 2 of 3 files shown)


joelteply:
🟢 SENTINEL: APPROVE

Tried hard to break it across all 6 vectors; found no correctness defect. Verdict defaults to BLOCK and I could not justify one.

**Unit math (resolve_stale_after_ms, commands.rs:1883-1888)** — `saturating_mul(3_600_000)` is correct: no overflow panic, verified by the new test asserting `Some(u64::MAX) → u64::MAX`. `None → DEFAULT_PEER_STALE_AFTER_MS` (1h, not 0 — the dangerous silent-disable case is explicitly guarded by the test). `Some(0) → 0`, and since classify_peer_prune (peers.rs:92) uses `age_ms >= stale_after_ms`, a 0 grace makes every absent-untrusted peer evictable — exactly what the help text promises.

**Wiring** — `DEFAULT_PEER_STALE_AFTER_MS` in commands.rs now appears ONLY in the helper (1886) and its test (2336). The hot path (1945) threads the resolved `stale_after_ms`. No hardcoded default leaks onto the live path.

**Semantics / footgun** — smaller = more aggressive (less grace); direction is described correctly. The `--stale-after-hours 0` self-eviction footgun (evicting a peer you just beaconed) is explicitly called out in BOTH the clap help (cli.rs:1356-1357) and the fn doc-comment. Documented, intentional, acceptable.

**CLI plumbing** — `Option<u64>` + `#[arg(long)]`: omitting yields `None` (default 1h), not `Some(0)`. main.rs:434-437 destructure matches the cli.rs:1347-1360 variant fields exactly.

**Output** — `stale_after_ms / 3_600_000` prints 1 for default, 0 for `Some(0)`; sub-hour overrides are impossible (u64 hours), so no misleading rounding. Note (non-blocking, cosmetic): `--stale-after-hours $(u64::MAX)` saturates to u64::MAX ms and the info line then prints ~5.1e12 hours rather than the requested count — divergence only under an absurd input; the classifier still receives the correct saturated value.

**Regression** — default path passes the identical `DEFAULT_PEER_STALE_AFTER_MS` to classify_peer_prune; the only added behavior on the omit path is one human-readable info line. run_peer_prune has no JSON mode, so no machine-output corruption.

Verified locally: `cargo check -p airc-cli` ✅, `cargo fmt -p airc-cli --check` ✅, `cargo clippy -p airc-cli` clean ✅, `cargo test -p airc-cli --bin airc resolve_stale_after` (1 passed) ✅, `cargo test -p airc-lib peer_prune` (2 passed) ✅.

PR #1203, opened 2026-06-14
feat(monitor): airc monitor attach reconnects on stream drop (receive-resilience)

joelteply:
## Receive-resilience — `airc monitor attach` survives daemon churn

**Motivated by today's live disconnect.** The persistent monitor (my Mac-coordination receive channel) went *permanently* dark when its daemon was killed by cargo/test churn. Root cause: each per-channel attach task `return`ed the instant its stream dropped — a single attach with **no reconnect**. When all channel tasks died, the merge channel closed and the whole live tail ended, with no recovery until manually re-armed.

### Fix
Wrap the attach + drain in a **bounded-backoff reconnect loop**:
- dropped stream (EOF/read error) or failed attach → back off and re-attach, instead of `return`.
- only stop when the **consumer** (the merge mpsc) is gone — then there's nothing to feed.
- a **healthy attach resets** the backoff (transient blip recovers in 250ms); a daemon that stays down is retried at most every 5s.
- a single undecodable frame now drops+reattaches instead of killing the tail (was `return`).

### Scope / limitations (called out honestly)
- Reconnect resumes the **LIVE** tail; events during the gap are **not** replayed — cursor-resume is a follow-up.
- Same-scope daemon restarts reuse the deterministic socket path, so the re-attach lands on the respawned daemon. A cross-**scope** identity flip (the `ce8b9074`→`ac873384` case from today's post-mortem) is **seam #1's** domain, out of scope here.

### Test
Pure `next_reconnect_backoff_ms` helper + unit test (doubles, saturates at the 5s cap, no overflow-panic). The reconnect loop itself is integration-level (needs a daemon harness) — noted.

### Why it matters beyond my comms
This is the grid lesson in miniature: **event streams must survive node churn.** Today it was "my monitor broke"; under lane coordination it's "the grid's event bus must not go dark when a node restarts."

### DIFF
--- a/crates/airc-cli/src/monitor/attach.rs
+++ b/crates/airc-cli/src/monitor/attach.rs
@@ -30,6 +30,23 @@ enum MonitorFrame {
     },
 }
 
+/// Receive-resilience (2026-06-14 disconnect): the first re-attach after
+/// a healthy stream drops waits this long. Short so a transient IPC blip
+/// or a fast daemon restart recovers near-instantly.
+const RECONNECT_INITIAL_BACKOFF_MS: u64 = 250;
+
+/// Cap on the reconnect backoff: a daemon that stays down is retried at
+/// most this often, so the monitor neither hammers a dead socket nor
+/// goes permanently dark.
+const RECONNECT_MAX_BACKOFF_MS: u64 = 5_000;
+
+/// One exponential-backoff step for the attach reconnect loop: double,
+/// capped at [`RECONNECT_MAX_BACKOFF_MS`]. `saturating_mul` so a runaway
+/// value can never overflow-panic.
+fn next_reconnect_backoff_ms(current: u64) -> u64 {
+    current.saturating_mul(2).min(RECONNECT_MAX_BACKOFF_MS)
+}
+
 pub(crate) async fn run(
     home: &Path,
     _my_name: &str,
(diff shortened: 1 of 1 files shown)


joelteply:
🔴 SENTINEL: BLOCK

Build is green on my isolated worktree (worktree off 1b694fb): `cargo check -p airc-cli` ✅, `cargo fmt -p airc-cli --check` ✅, `cargo clippy -p airc-cli --all-targets` ✅ (no warnings), `cargo test -p airc-cli --bins attach` ✅ (5/5, incl. `reconnect_backoff_doubles_then_caps`). The backoff helper, the `tx.is_closed()` exit point, and the termination-semantics change (one sole caller — interactive `airc monitor attach`, MonitorAction::Attach in main.rs:837 — so "live tail until Ctrl-C" is the intended contract, no one-shot consumer regressed) all check out.

But there is a real, reachable defect that contradicts the PR's own scope claim.

## DEFECT: reconnect re-floods the full transcript in backlog mode (and can hard-loop forever)

`start` is captured ONCE before the per-channel loop (`attach.rs:73-77`) and reused on every reconnect (`attach.rs:98` builds the request from the same `start` each iteration). The PR description states: *"Reconnect resumes the **LIVE** tail; events during the gap are **not** replayed."* That is true ONLY for the default `from_now=true` → `AttachStart::Live` path. It is **false** for the opt-in backlog path.

Reachable config: `airc monitor attach --my-name X --include-backlog --no-coalesce-backlog` → `from_now=false`, `coalesce_backlog=false` (cli.rs:35-52, main.rs:854). That yields `start = AttachStart::FromTranscriptStart`. The daemon maps `FromTranscriptStart => None` (server.rs:496) and replays from the **beginning of the transcript** — it does NOT persist a per-client cursor across separate attach calls. So:

1. **Duplicate-flood regression** — every reconnect (the exact failure modes the PR cites: "cargo/test churn, idle self-exit") re-streams the ENTIRE transcript event-by-event again. An audit/replay consumer parsing stdout (the stated use case for `--no-coalesce-backlog`, cli.rs:48-50) gets the whole history re-dumped on each blip, no dedup. Pre-PR the stream just died; the new behavior silently duplicates instead, which for audit tooling is its own correctness break.

2. **Compounds with the decode-error change (concern #6) into a hard loop.** The new `Err(_) => break` on an undecodable frame (`attach.rs:117`) drops + re-attaches. In `Live` mode the poisoned frame is in the past → not replayed → fine. But in `FromTranscriptStart` mode the same poisoned historical frame is replayed on EVERY reconnect → break → reconnect → re-hit the same frame → never advances past it. Because backoff resets to 250ms on each successful attach (`attach.rs:105`), this is a permanent 250ms re-replay loop stuck behind one bad frame, re-dumping everything before it forever, with an `eprintln` (attach.rs:138) per cycle (~4/sec stderr spam).

## Root cause / fix
The CLI must advance past `FromTranscriptStart` after the first successful backlog drain — subsequent reconnects should attach with `AttachStart::Live` (or `After(last_seen_cursor)` once cursor-resume lands). As written, `start` is immutable across the loop. Minimal fix: after the inner drain completes at least one frame (or unconditionally after the first attach), set a local `start = AttachStart::Live` for the next iteration. That also fixes the poison-frame hard-loop.

## Lesser notes (not blocking on their own)
- **Backoff escalation defeated by reset-on-attach-success** (attach.rs:105): a daemon that accepts-then-instantly-EOFs resets backoff to 250ms every iteration, so it never escalates — steady 4 reconnects/sec + 4 eprintln/sec indefinitely. Bounded (always a 250ms sleep, not a CPU busy-spin) and the author documented the choice, but resetting on first-frame-received rather than on attach-success would be more correct. Worth a follow-up.

Recommend: gate the reconnect on switching to `Live`/`After` after the initial attach, add a test for the backlog-reconnect path, and correct the scope note (which currently overstates the LIVE guarantee). Happy to re-review once the `start`-advance is in.

joelteply:
Fixed in the latest push — sharp catch, both the re-flood and the wedge were real. The configured `start` now applies to the **first attach only**; after one attach completes, reconnects resume `Live`, so the full-transcript replay can't recur and the decode-error drop can't wedge on a poisoned historical frame. Also took the non-blocking note: backoff now resets only on a delivered frame (`got_frame`), so an accept-then-close flap keeps escalating instead of pinning at the floor. Trade-off documented in-code: a first attach that drops mid-backlog forfeits the un-replayed remainder rather than re-flooding — per-attach cursor-resume is the follow-up. Re-review welcome.

joelteply:

Re-reviewed fix commit `796d90d` against the prior BLOCK. **Both the re-flood AND the wedge are CLOSED.**

**1. Re-flood CLOSED.** `attach_start` is initialized to the configured `start` (line 107) and used only for the *first* `AttachRequest` (line 109). After the inner read loop drains, `attach_start = AttachStart::Live` (line 149) unconditionally. So with `--include-backlog --no-coalesce-backlog` (FromTranscriptStart), the first attach replays the backlog once; every reconnect thereafter is Live. Full-transcript-replay-on-every-reconnect is gone.

**2. Wedge CLOSED.** A poisoned historical frame hits `Err(_) => break`, exits the inner loop, and `attach_start` flips to Live regardless of `got_frame`. The Live re-attach never re-requests the historical frame, so the decode-error break can't re-encounter it in a tight 250ms loop. The permanent re-replay is broken at the source — even if the poisoned frame is the very first backlog frame.

**3. Backoff CORRECT.** `got_frame` is set on both `Response::Event`→`Ok` (line 119) and `AttachCursorAdvanced` (line 130). The reset `backoff_ms = INITIAL` is gated on `if got_frame` (line 145). An accept-then-close flap delivers no frame → backoff escalates (250→500→…→5000) instead of pinning at the floor.

**4. New-regression: ACCEPTABLE.** Switching to Live after a first attach that dropped mid-backlog forfeits the un-replayed remainder. This is the documented, lesser-evil trade-off vs. re-flooding the whole transcript; per-attach cursor-resume is the called-out follow-up. It does not affect the default path.

**5. Default Live path UNCHANGED.** Default `airc monitor attach` (`from_now=true` → `start=Live`) starts with `attach_start=Live` and stays Live; the fix is a no-op there. Healthy-daemon behavior is identical to pre-fix.

**Build/test (isolated target dir):**
- `cargo check -p airc-cli` — clean
- `cargo test -p airc-cli --bins monitor::attach::tests` — 5 passed (incl. `reconnect_backoff_doubles_then_caps`)
- `cargo clippy -p airc-cli` — zero warnings
- `cargo fmt -p airc-cli --check` — clean

**Non-blocking note (not a BLOCK):** the new control-flow (`attach_start` advancement + `got_frame` gating) lives inside the spawned async closure and has no direct unit coverage; only the backoff *math* is tested. Logic traces cleanly, but a follow-up test with a mock daemon (first attach replays N frames + drops → assert second attach requests Live) would lock the invariant against regression.

BLOCK lifted.

PR #1204, opened 2026-06-14
feat(cli): airc --here — explicit opt-in to a cwd-local scope (seam #1)

joelteply:
## Seam #1 follow-up — the `--here` ergonomic flag (AGREED (c))

#1190 made a rootless / non-project cwd resolve to the **machine-account home** instead of minting a throwaway `cwd/.airc` (those throwaway scopes were publishing phantom beacons to the mesh — the observed ghosts). Its resolver comment (`cli.rs`) named the missing counterpart:

> Explicit cwd-scoping stays available via `$AIRC_HOME` (a future `--here` flag is the ergonomic shortcut; tracked as a follow-up).

This PR adds that flag.

### Behavior
- `--here` (global, **conflicts with `--home`**) resolves home to **`$PWD/.airc`** — the ergonomic equivalent of `AIRC_HOME=$PWD/.airc`.
- It's the **intentional** opt-in to a per-directory scope: `airc init --here`, `airc join --here`. Having it is exactly what makes the *default* safe to prefer the machine-account home rather than fragmenting into throwaway cwd scopes (the whole seam #1 thesis).
- Once `init --here` creates `$PWD/.airc`, the existing ancestor-walk in `default_home_dir_for_with` naturally resolves subsequent commands there — no further plumbing.

### Implementation
- Global `--here` bool on `Cli` (`conflicts_with = "home"`).
- Pure `explicit_home_override(home_flag, here_flag, cwd) -> Option<PathBuf>` (None → caller falls back to `default_home_dir`), wired at the single `dispatch()` resolution point.
- Precedence test: `--home` wins; `--here` → cwd-local `.airc` (NOT machine-account); neither → None.

### Validation
`cargo test -p airc-cli explicit_home_override` ✓, fmt ✓, clippy ✓ (locally on Windows — airc compiles natively).

### DIFF
--- a/crates/airc-cli/src/cli.rs
+++ b/crates/airc-cli/src/cli.rs
@@ -55,6 +55,27 @@ pub fn default_home_dir() -> PathBuf {
     default_home_dir_for(&cwd)
 }
 
+/// Resolve the explicit home override from the global `--home` / `--here`
+/// flags. `--home <path>` wins (and clap forbids passing both). `--here`
+/// is the ergonomic shortcut for `AIRC_HOME=$PWD/.airc` (seam #1): the
+/// intentional opt-in to a cwd-LOCAL scope. Returns `None` when neither
+/// is set — the caller then falls back to [`default_home_dir`] (which
+/// itself honours `$AIRC_HOME`, then the machine-account / git-project
+/// resolution). Pure: `cwd` is injected so tests need no real `$PWD`.
+pub fn explicit_home_override(
+    home_flag: Option<PathBuf>,
+    here_flag: bool,
+    cwd: &Path,
+) -> Option<PathBuf> {
+    if let Some(home) = home_flag {
+        return Some(home);
+    }
+    if here_flag {
+        return Some(cwd.join(".airc"));
+    }
+    None
+}
+
 fn default_home_dir_for(cwd: &Path) -> PathBuf {
     let machine_account = std::env::var_os("HOME")
         .or_else(|| std::env::var_os("USERPROFILE"))
@@ -449,6 +470,17 @@ pub struct Cli {
     #[arg(long, global = true)]
     pub home: Option<PathBuf>,
 
+    /// Use a cwd-LOCAL `.airc` scope (`$PWD/.airc`) for this invocation,
+    /// instead of the default machine-account / git-project resolution
+    /// (seam #1). The ergonomic shortcut for `AIRC_HOME=$PWD/.airc` —
+    /// the explicit, intentional way to mint or use a per-directory
+    /// scope (`airc init --here`, `airc join --here`) when you genuinely
+    /// want one, which is what lets the default safely resolve to the
+    /// machine-account home rather than a throwaway cwd scope. Conflicts
+    /// with `--home`.
+    #[arg(long, global = true, conflicts_with = "home")]
+    pub here: bool,
+
     /// Ad-hoc peers to enrol for this invocation only, repeatable.
     /// Format: `<uuid>:<base64-pubkey-no-padding>`. Persistent peers
     /// come from the peer trust store (managed via `airc peer add`);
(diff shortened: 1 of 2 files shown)


joelteply:
🟢 SENTINEL: APPROVE

Adversarial review of `feat/here-flag` @ 27e95fe (vs origin/canary). Tried hard to break it on all 6 axes; default-BLOCK lifted. Ran fmt --check, `cargo check -p airc-cli`, clippy, and the cli tests myself — all green.

## Verdict per axis

**1. Seam #1 regression — CLEAN.** The whole point of #1190 (stop cwd minting throwaway `cwd/.airc` phantom-beacon scopes) is preserved. `git diff origin/canary...` shows `default_home_dir` / `default_home_dir_for` / `default_home_dir_for_with` are **byte-identical** to canary — the only `+/-` lines touching them are doc comments and a test-import. `--here` is a strictly **additive opt-in branch**: `explicit_home_override(None, false, _) → None`, so `dispatch` falls through to the same `.unwrap_or_else(cli::default_home_dir)` the base branch had (`main.rs:206` base → `main.rs:211` PR). Default path is untouched; never auto-cwd. Verified e2e: bare `init` does not mint a cwd scope.

**2. Precedence & conflict — CLEAN.** `cli.rs:481` `#[arg(long, global = true, conflicts_with = "home")]` — field-id `home` is correct. Built the binary and ran it: `airc --home /tmp/x --here version` → `error: the argument '--home <HOME>' cannot be used with '--here'`, **real exit code 2** (clap usage error), not a silent pick. Resolver `cli.rs:66-77` also encodes home-wins as defense-in-depth, and the test pins it (`Some(home), true, cwd) → home`).

**3. CWD semantics — ACCEPTABLE / documented.** Confirmed literal `$PWD/.airc`: `cd repo/sub && airc --here init` created `repo/sub/.airc`, NOT `repo/.airc`. This is the intended "here = literally here" / `AIRC_HOME=$PWD/.airc` shortcut, and the clap long_about spells it out ("cwd-LOCAL ... `$PWD/.airc`"). It is an explicit, intentional opt-in (you typed `--here`), so the subdir behavior is a feature, not a footgun — distinct from the *default* git-project resolution, which correctly still walks to the project root.

**4. Reaches all commands — CLEAN.** `dispatch` (main.rs:209-211) resolves `home` ONCE and threads `&home` to every arm, including the scope-creating `Command::Init` (`run_init(&home, ...)`) and `Command::Join` (`run_join(&home, room)`). No command re-resolves home from the override flags. (The independent `current_dir()`/`HOME` reads elsewhere — collaboration_commands.rs:136, commands.rs:829, etc. — are pre-existing machine-account/account-wide reads outside this diff and outside the `--home` contract too; `--here` does not regress them.) Verified e2e that `--here init` lands the scope.

**5. Broken cwd — CLEAN.** `main.rs:209` `current_dir().unwrap_or_else(|_| PathBuf::from("."))` → `explicit_home_override(None, true, ".")` → `Some("./.airc")`. Sane fallback, no panic.

**6. Test net — REAL.** `explicit_home_override_precedence` covers home-wins (x2), here→`/work/project/.airc`, neither→None. It is a genuine net: if `--here` resolved to machine-account home instead of `$PWD/.airc` the `Some(PathBuf::from("/work/project/.airc"))` assert fails. Passes (`test cli::tests::explicit_home_override_precedence ... ok`), and all 16 home/socket cli tests pass.

## Non-blocking advisory
The unit test exercises the **resolver**, not the **clap `conflicts_with`** at parse time. I verified the conflict manually (exit 2), but there is no regression test pinning the `conflicts_with = "home"` attribute — a future refactor of the arg id could silently drop it. Consider a `Cli::try_parse_from(["airc","--home","/x","--here","version"])` test asserting `is_err()`. Optional; not a merge blocker.

Tooling: `cargo fmt --check` ✅ · `cargo check -p airc-cli` ✅ (no warnings) · `cargo clippy -p airc-cli` ✅ (no warnings) · `cargo test -p airc-cli --bin airc` (home/cli) ✅ 16 passed.

Tight, well-scoped (+71/-2), default contract provably intact. Ship it.

PR #1209, opened 2026-06-15
fix(monitor): `airc monitor attach --my-name` optional (was required, ignored)

joelteply:
## What

`airc monitor attach` **required** `--my-name`, then **ignored it** — the param is `_my_name`, unused in `attach::run` (the daemon already owns identity). A fresh agent running `airc monitor attach` hit a cryptic clap **exit-2** for a value that does nothing.

Make it `Option<String>`:
- `airc monitor attach` now works **flagless**.
- `--my-name X` still accepted (backward compat).
- Regression tests for both parse paths.

## Scope change (force-pushed)

This branch originally also rerouted `airc join` through `run_attach` to "fix a hang." **Reverted.** Adversarial review + a live experiment proved canary's `join_feed` works fine on a healthy daemon — the join "hang" was the **daemon being starved while blocking on dead-peer dials**, not `join_feed`. Rerouting regressed unread-catch-up, cross-sandbox socket resolution, and join stdout format. The real root cause is filed as **#1210** (daemon-side). This PR is now just the clean, no-regression `--my-name` fix.

## Local gate (green)
- `cargo fmt --all --check` clean
- `cargo clippy -p airc-cli --release` clean
- `cargo test -p airc-cli --bin airc` → **292 passed, 0 failed**

### DIFF
--- a/crates/airc-cli/src/main.rs
+++ b/crates/airc-cli/src/main.rs
@@ -857,6 +857,9 @@ async fn dispatch(parsed: Cli) -> Result<(), Box<dyn std::error::Error>> {
                 // defaults to true; flipped to false by user only
                 // when they want event-by-event replay (audit case).
                 let live_only = from_now && !include_backlog;
+                // `my_name` is optional and unused by attach (the daemon
+                // owns identity); accepted for backward compat.
+                let my_name = my_name.unwrap_or_default();
                 monitor::run_attach(&home, &my_name, live_only, coalesce_backlog).await
             }
         },
--- a/crates/airc-cli/src/monitor/cli.rs
+++ b/crates/airc-cli/src/monitor/cli.rs
@@ -22,9 +22,12 @@ pub enum MonitorAction {
 
     /// Attach to the Rust event stream without owning transport.
     Attach {
-        /// Current local display name.
+        /// Current local display name. Optional and currently unused by
+        /// `attach` (the daemon already owns identity); accepted for
+        /// backward compatibility with callers that still pass it. A
+        /// fresh agent can simply run `airc monitor attach`.
         #[arg(long)]
-        my_name: String,
+        my_name: Option<String>,
         /// **Card 7d5b6a65.** Subscribe from the LIVE EDGE — only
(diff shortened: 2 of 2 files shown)


joelteply:
## ⚠️ Self-review (adversarial agents) caught regressions — converting to DRAFT, needs rework

Two adversarial reviewers flagged real problems with the "route join through `run_attach`" approach:

1. **Lost catch-up (HIGH).** The deleted `join_feed` resumed from the per-runtime `join-feed:` cursor (unread-since-last-seen, CATCH_UP_LIMIT=64) before streaming. The new path is `AttachStart::Live` (live-edge only) → a rejoining agent sees **nothing** until the next live message. `--include-backlog` is NOT equivalent: it's `FromTranscriptStart` (full history, not unread-delta) and coalesces to a "skipped N events" **count**, not the bodies.
2. **Cross-sandbox socket regression.** `attach::run` recomputes `default_socket_path_in(home)` and ignores the socket `ensure_daemon_running` resolved. For Codex/sandbox agents whose `$HOME` is a tmpdir, discovery returns a *different* socket; the feed would attach to a dead socket and reconnect-loop forever — dead for exactly the agent case this targets.
3. **stdout format change.** Join feed now emits the `<pm-NONCE>` XML + the `[contract]` banner instead of the old `[Message] sender → room: text`. Card 745e93f0 says harnesses scrape join stdout — undocumented-by-test, so silently breakable.
4. Orphaned `join-feed:` cursor + stale doc ref at `airc-lib/src/daemon.rs:441`.

**Open root-cause question:** the daemon log on a fresh Mac shows it blocking on dead-peer dials (3s ea.) + `routed_forward` 10s ack waits (offline BIGMAMA). The original `join` "hang" may actually be the **daemon being starved by dead-peer I/O**, making `join_feed`'s synchronous catch-up request hang — i.e. the bug may be daemon-side, not `join_feed`. Reworking to: keep catch-up + correct (resolved) socket, and investigate the daemon dead-peer blocking. The `monitor attach --my-name` → optional fix is independent and good; may split it out.

PR #1280, opened 2026-06-29
feat: generic scoped-state store (private sibling of the room wall)

joelteply:
## What

A generic, durable, scope-keyed `key → JSON` store in airc — the **private/local** sibling of the room wall. Four clean slices (store → core → lib → cli), each its own green commit.

Three scopes, one flat `scope_key`: `user:<peer>` / `room:<room>` / `uir:<peer>:<room>`. The composite-PK leftmost prefix doubles as the list-by-scope range scan, so the durable store stays a clean two-column `(scope_key, key)` table.

## Why scoped_state is PRIVATE-only (it does NOT broadcast)

airc **already** has a shared room wall: `Airc::publish_wall_post(category, body, supersedes)` + `wall_posts(category_filter)` — event-sourced, broadcast, supersede-chain LWW, open categories. So:

- **Shared room documents** — a room's plan, coding instructions, the recipe — stay on the **existing wall**. This PR does not duplicate that.
- **Private per-peer state** the wall can't serve — prefs, "where was I last" / the tool-menu mode cursor, widget UI state (open tabs / rooms) — is high-churn, peer-private, last-write-wins. A growing supersede log is the wrong shape; a durable LWW row is right. That's this store.

A consumer (continuum's planned `WallSource`) composes BOTH — wall (shared) + scoped_state (private) — into one unified scoped grounding surface. The payoff: **concise RAG** (explicit, scoped, budgetable layers, not an undifferentiated dump).

## Slices

| Slice | Commit | What |
|---|---|---|
| store | `f33e8b5` | `scoped_state` table + migration + `EventStore::{get,set,list,delete}_scoped_state` (Sqlite OnConflict upsert + InMemory). Scope-agnostic. |
| core | `57febf7` | `ScopeRef` (owns scope_key encode/parse, exact inverses, fail-loud on bad prefix/uuid) + `ScopedStateEntry` domain view. **No wire event** — private state never broadcasts. |
| lib | `9889928` | `Airc::{get,set,list,delete}_scoped_state`. `set` single-sources provenance (this peer → `updated_by`, lib clock → `updated_at_ms`); caller owns the LWW `version`. |
| cli | `90219ec` | `airc state get/set/list/delete`. Scope defaults to this peer; `--in-room` = (this peer, current room). |

## Verification

- `cargo check` + `cargo clippy` clean across the workspace.
- Unit: scope_key round-trips for all three scopes; parse rejects garbage (fail-loud, no silent default); lib facade round-trip (set → get value/version/updated_by/updated_at_ms; list; delete; idempotent re-delete).
- **Live smoke** end-to-end through a real daemon: init → get-unset (nothing, exit 0) → set ×2 → get → list (sorted) → delete → idempotent re-delete.

## Open questions for review
- Size cap per value (proposed 256KB, fail-loud)?
- Read/write trust gating on `room:`/`uir:` scopes (who may write another peer's scope — today the lib stamps the caller as `updated_by` so spoofing is prevented, but cross-peer writes aren't yet gated).
- Boundary vs per-persona workspace isolation (continuum task #49) — that's compute/FS leasing; this is lightweight KV metadata.


joelteply:
Superseded: the generic scoped-state store this PR introduces (crates/airc-{core,store}/scoped_state.rs, entities/scoped_state.rs, the m20260629_000018_create_scoped_state migration, and the state_cli/state_commands CLI) is ALREADY on canary via a later merge — every file here exists on canary today, and this branch only conflicts with the version that landed. Closing to keep task #89 single-sourced. If any delta on this branch is missing from canary's version, it should be re-authored as a small follow-up against current canary rather than merging this stale duplicate.

Issue #37, opened 2026-04-24
Gist transport for invite shortstring (recovery + chat-mangle resilience)


joelteply:
## Problem

The current airc invite encodes the SSH key as a base64 blob inline:

```
continuum-a25c@joelteply@100.91.51.87#c3NoLWVkMjU1MTkgQUFBQUMzTnphQzFsWkRJMU5URTVBQUFBSUlPUXU0ZldWTjhMYUlhZE8zSCs3Rk5tcEttSUd3NGRhZDFnME0zd1FCMFUgYWlyYy1jb250aW51dW0tYTI1Ywo=
```

That's ~200 chars and hits real-world chat clients hard:
- iMessage / WhatsApp insert line breaks at ~80-char boundaries
- Slack auto-formats certain `=` runs as code blocks and adds invisible spans
- Email clients silently word-wrap and corrupt the trailing chars
- Manually retyping is error-prone

Symptom seen 2026-04-23 paired QA: airc host on Mac, peer on WSL2 box, and getting the invite from one to the other was the actual hard part of the workflow. Joel's working idea: "could we build gist into this airc thing? as like a shortstring?"

## Proposal

`airc connect` (host mode) accepts an optional `--gist` flag (or detects when `gh` is available + a config preference is set). When enabled:

1. Generate the invite content as today (SSH key + metadata).
2. Push the invite content as a **secret gist** via `gh gist create -f airc-invite.txt -d "airc invite (delete after pair)"`. Capture the short ID.
3. Print the invite as `name@user@host:port#gist:<id>` instead of `name@user@host:port#<base64>`. Total length: ~30-40 chars, single-line, copy-paste-survives-anything.
4. Receiving peer's `airc connect <shortstring>` parses, sees `#gist:` prefix, fetches via `gh gist view <id> --raw` (or `curl gist.githubusercontent.com/.../raw`), extracts the actual key, completes pairing.

## Single-use semantics

Once pairing succeeds, the host SHOULD `gh gist delete <id>` automatically. Reasons:
- Avoids accumulating dead invites in the user's gist list
- Reduces window for a stolen URL to be replayed if the host is reachable
- Aligns with "credentials are single-use" general security posture

If the pairing fails, the gist stays so the user can retry without a fresh `airc connect`.

## Belt-and-suspenders fingerprint

[...]

PR #1281, opened 2026-06-30
fix(identity): durable per-peer identity index — names survive busy rooms

joelteply:
## Problem

Peer display names decay to `None` in busy rooms. Name resolution scanned only the **bounded recent-transcript window** for each peer's `IdentityPublished` card. A card is published **once per join**, so in any room with more than `window` events since a peer joined, the card scrolls out of the scan and `peer_alias` / `room_roster` return `None`. Consumers (continuum personas) then confabulate names from raw uuids — the "Joel and BigMama" decay.

## Fix

A **durable per-peer identity index** in `scoped_state` (`ScopeRef::User(peer)` → key `identity.card`), LWW-keyed on the card's `emitted_at_ms` (stored verbatim as the scoped-state `version`; the caller guards LWW). Every observed `IdentityPublished` event is recorded into it:

- **inbound peer cards** via both `append_received_frame` branches (daemon-sink + direct);
- **this peer's OWN card** via `emit_lifecycle` — the only path that persists self-originated lifecycle events (it bypasses both the received-frame and sent-message-frame paths; that's why the self-name never indexed otherwise).

`peer_alias` / `peer_identity_card` now **read the index** instead of scanning the transcript; `room_roster` resolves each live peer's name through `peer_alias`, and the now-redundant `peer_display_names` window scan is **deleted**. The index is room-independent (one identity per peer) — strictly better than the old per-room scan.

Also: seed a **named** citizen's published nick from its `agent_name` at identity-row creation (`generate_and_save_as`), so `attach_as("Claude")` / `init --as Claude` publishes a card carrying `"Claude"` from its first beat rather than an empty name. The default scope keeps an empty name (the user's identity, set later via `airc identity set`).

Adds `AircError::Serde` to surface (never swallow) identity (de)serialization into the index.

## Test

`peer_name_survives_identity_card_scrolling_past_the_recent_window` publishes a card, buries it under `> window` plain messages, and asserts **both** `peer_alias` and `room_roster` still resolve the name. It **fails against any window-bounded scan** — a true regression guard for this bug.

Full `airc-lib` suite green (297 unit + all integration); `airc-identity` green; `cargo clippy --workspace --all-targets` clean.

## Stacking note

Based on `feat/airc-scoped-state` (#1280) — this is its first identity consumer. Retarget to `canary` once #1280 merges.

## Continuum side

Continuum's `RoomRosterSource` already resolves peer_id→name from the `room-roster` delivery (continuum commit `0f6108885`); this PR makes that delivery actually carry the durable `display_name`. Closes the airc half of the roster-name-decay bug.


joelteply:
Disposition after triage (auto-closed when its base branch `feat/airc-scoped-state` was deleted as superseded — see #1280):

This branch carries genuinely-new work worth keeping — the `airc-ipc` request/response types, the `airc-lib/tests/room_roster.rs` test, and the durable per-peer identity index — but it is **tangled**: it was stacked on the old scoped-state branch, so ~half its diff duplicates the scoped-state store that has since landed on canary independently, and its ubuntu test is red. It also overlaps the actively-evolving identity substrate (continuum #37/#38).

Rather than resurrect this stale, half-duplicated branch, the identity-index + room_roster delta should be **re-authored fresh against current canary** as a focused PR (dropping the duplicated scoped-state, fixing the failing test), ideally folded into the #37/#38 identity work. The branch `fix/durable-peer-identity-index` is preserved on the remote as the reference for that extraction — nothing is lost.

PR #1282, opened 2026-06-30
fix(identity): route per-peer identity index to the coordinator store

joelteply:
## Problem

A continuum persona attached to the daemon resolved peer display names as **raw UUIDs** (e.g. `7711fe60` instead of `Claude`) and confabulated peer names. Root cause: a write/read store split on the per-peer identity index.

`Airc` holds two stores — the scope-local `event_store` (`<scope>/.airc/events.sqlite`) and the machine-account `coordinator_store` (`$HOME/.airc/events.sqlite`, the daemon's file). The identity index is **owner-core shared state** (like the transcript tip): every reader expects it in `coordinator_store` —
- the daemon's IPC index read (`handle_peer_identity_card` → `coordinator_store.get_scoped_state`), and
- the daemon-less self-resolution branch of `peer_identity_card`.

But `record_peer_identity_card` **wrote** the index to `event_store()` (the scope home). So the daemon's `coordinator_store` had zero `identity.card` rows while cards were stranded in per-scope homes — the attached persona read an empty index over IPC.

## Fix

Add `get_coordinator_scoped_state` / `set_coordinator_scoped_state` helpers and route **both** the write (`record_peer_identity_card`) and the daemon-less read branch (`peer_identity_card`) through them. Because `coordinator_store` is the one shared machine-account `events.sqlite`, any scope that records a card writes it to the single index every sibling scope **and** the daemon read.

The general scope-home `set/get_scoped_state` (tool.mode / wall / coordination) are intentionally **unchanged** — only the identity index moves.

## Test

`peer_identity_index_converges_across_scopes_via_coordinator_store`: one scope records a peer card, a *different* scope sharing the same `wire_root` (machine account) resolves it — the daemon-less analog of the attached-client → daemon-index read. Pre-fix the reader read its own empty scope store and answered `None`.

`cargo test -p airc-lib --lib` green; continuum-core checks clean against this rev.


joelteply:
Duplicate of #1281: both PRs share the exact same head branch (fix/durable-peer-identity-index) — this one is opened against canary, #1281 against feat/airc-scoped-state. Keeping a single PR for the branch (#1281) to avoid a double-merge. Closing this one; the identity-index work is tracked on #1281.

PR #1284, opened 2026-07-02
feat(account-registry): shared-folder rendezvous + #113 selector seam (auto-pick gist vs folder)

joelteply:
## What

A second, maximally-different `AccountRegistryStore` impl — a **shared-folder rendezvous** (iCloud Drive / Syncthing / NFS mount) that N machines of one account see, needing **no `gh` CLI, no token, no network**. This is the outlier-B that proves the rendezvous seam (`GhAccountRegistryStore` = outlier A) is genuinely swappable.

## Why

Rendezvous **initiates** the exchange; it does not **supply** the security — that lives in the E2E data plane (paired peer keys, trust tiers) regardless of which door a node came through, exactly the Tailscale coordination-server shape. So a second door is free to be dumb and credential-free. Two payoffs:

- **"Other means if gist went away."** If GitHub/gist is unavailable, this is already a working substitute at the same seam.
- **On-prem / behind-firewall grid.** A hospital deploy can't reach GitHub; a shared mount is the rendezvous there.

## Fidelity — same convergence core as gist

gh keeps one gist **per writer/machine** and `refresh` folds them all via `merge_registry_documents` (freshest beacon per peer, union, skip foreign-mesh, prune stale). This does the byte-identical thing on a filesystem: each machine writes its doc to a **per-writer file** (`<dir>/<identity>/<writer_file>`) so two machines never clobber, and `refresh` reads every writer file and runs the **same** `merge_registry_documents` + `prune_stale_peers` — the merge/freshness logic stays in exactly one place.

- **publish:** validate → pretty JSON → atomic pid-temp write + rename (a concurrent `refresh` on another machine never reads a half-written doc)
- **refresh:** read every `.json` writer file (skips `.tmp` staging + OS cruft like `.DS_Store`/`.icloud`), **fail loud** on a corrupt writer file, merge, prune stale
- unpublished identity → `Ok(None)`, not an error (mirrors gh's no-gist → None)

## Tests (one mod, `// what this catches:`)

- `publish_then_refresh_round_trips_the_document`
- **`two_machines_sharing_a_folder_converge_via_merge`** — the keystone: two writers, one shared dir, distinct writer files → both peers merged, **zero GitHub**
- `refresh_of_an_unpublished_identity_is_none_not_error`

Pure-Rust validated: `cargo test -p airc-lib --lib account_registry_fs` → 3 passed, clippy clean, fmt clean.

## Not in this PR (follow-up slice)

Provider **selection** — a `resolve_account_registry_store(...)` seam + a `RegistryRefreshGate::AlwaysReady` variant for the no-auth fs path, rewiring the hardcoded `GhAccountRegistryStore::new` boot sites. This PR only proves the trait is swappable; wiring the selector is next.

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -1458,23 +1458,40 @@ pub async fn run_daemon(
                 bin.display()
             );
         }
-        // Stale-token recovery slot (card 1f2cbffa): the SAME slot
-        // goes to the gate (which re-resolves on auth failure) and the
-        // store (whose gh spawns then carry the recovered token), so a
-        // mid-session token rotation no longer bricks the rendezvous
-        // until daemon restart.
-        let gh_token_override = airc_lib::GhTokenOverride::new();
-        let store = match &gh_bin {
-            Some(bin) => airc_lib::GhAccountRegistryStore::new(event_store, &registry_home)
-                .with_bin(bin.clone()),
-            None => airc_lib::GhAccountRegistryStore::new(event_store, &registry_home),
-        }
-        .with_token_override(gh_token_override.clone());
-        let gate = airc_lib::RegistryRefreshGate::GhAuth {
-            gh_bin: gh_bin.clone(),
-            scope_home: registry_home.clone(),
-            token_override: Some(gh_token_override),
+        // Rendezvous SELECTION (#113): which door the mesh converges
+        // through is a data choice, not a hardcode. `AIRC_RENDEZVOUS_DIR`
+        // set → the no-GitHub shared-folder door (on-prem / behind
+        // firewall); unset → the default gist door. The resolver pairs the
+        // chosen store with the gate that matches it (gist → gh-auth,
+        // folder → Always), so this loop never learns which door won.
+        //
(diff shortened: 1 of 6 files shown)


joelteply:
Stacked a second commit onto this branch — the **#113 rendezvous selector seam** now sits on top of `FsAccountRegistryStore`:

- `resolve_account_registry_store(choice, gist)` picks the store for a `RendezvousChoice` and pairs it with the matching `RegistryRefreshGate` (gist → `GhAuth`, folder → `Always` — the existing variant, no new one). One delegating `impl AccountRegistryStore for Box<dyn AccountRegistryStore>` lets the boxed store flow through `run_loop`'s generic `S` bound, so SELECTION happens once at the seam.
- `AIRC_RENDEZVOUS_DIR` set → the no-GitHub shared-folder door; unset → the default gist door. Set-but-empty / non-UTF-8 **fail loud** (never a silent gist fallback).
- Daemon refresh loop (`commands.rs`) rewired through the resolver; it never learns which door won.
- ORM-abstracted throughout — local cache rides `SqliteEventStore`, folder door is plain JSON files, zero raw SQL.

This is the substrate half of #18 ("join to Asha across machines just works"). 5 new unit tests green, clippy clean.

Issue #348, opened 2026-04-30
Host eviction triggered by 'gh: unsure what file to edit; either specify --filename or run interactively'


joelteply:
Caught during the Carl QA pass mesh-coordination session. Not a hard blocker — substrate self-heals — but the underlying gh call is missing a `--filename` flag.

## Symptom

Mid-session, hosting #general:

```
[2026-04-30T00:45:10Z] airc: [HOST EVICTED] heartbeat to gist 9740e0e17548f6515f060e848d756e6b failed 3 consecutive times — self-healing.
last stderr: unsure what file to edit; either specify --filename or run interactively
```

The "unsure what file to edit" string is `gh gist edit`'s error when called against a multi-file gist without `--filename` and without TTY (so the interactive picker can't run).

## Likely cause

The heartbeat path calls `gh gist edit <id>` to write a new heartbeat timestamp into `messages.jsonl` (or whichever file the substrate uses for liveness markers). When the gist contains more than one file, gh requires `--filename <name>` to disambiguate. Background daemons / monitor child processes don't have a TTY, so the interactive fallback fails too — three retries → eviction → self-heal.

## Why we don't see it always

The error fires only when the gist has >1 file. Possible triggers:
- Mid-session a `airc nick` or other op added a second file
- Another peer added a file via cross-account share
- A leftover `airc.txt` / similar from a prior session

On this Mac the gist (`9740e0e1...`) was created from scratch by my `airc join` so it's plausibly multi-file by design (messages.jsonl + identities + channel registry).

## Severity

Low. The substrate is doing what it should — eviction → self-heal kept the mesh up. But:
- Spurious evictions burn the secondary-rate-limit budget faster (3 failed gh calls per heartbeat cycle that hits this path)
- The error stderr leaks gh's vocabulary ("unsure what file to edit") into the airc event stream — Carl wonders why airc is talking about file menus
- A quieter mesh would be a quieter mesh

## Fix shape

In whichever Python module performs the heartbeat write (likely `lib/airc_core/bearer_gh.py` or a peer), pass `--filename messages.jsonl` (or whatever the canonical heartbeat file is) explicitly to `gh gist edit`. If the file naming is dynamic, derive it from the same place as the read side.

## Test plan suggestion

Doctor scenario: `airc doctor heartbeat_persists_through_multifile_gist` — create a gist with 2+ files, become host, assert heartbeat updates land in the right file without eviction.

joelteply:
[...]

PR #1361, opened 2026-08-14
fix(rooms): a room is addressed by its ID — kill name-as-identity end to end

joelteply:
## What

Room names are labels. Only the id addresses a room. This branch removes
every place that treated a name — or a rendering of an id — as identity,
and adds the verb that was missing so callers can say what they mean.

## The five commits

1. `8ed9f30` fmt + reunite a doc comment with its function (self-inflicted in `490dc76`, attributed in the message)
2. `4c5d262` kill prefix-matching on rendered uuids, the `Ambiguous` variant, and `format!("{name} ({id})")` error payloads; `InviteBeacon.rooms: Vec<RoomId>`; constructors 4→2
3. `2c91a3f` stop cloning N subscriptions to answer what the `Copy` ids already answer; add `SubscriptionSet::get`
4. `9a6b464` `into_all()` — move out of a dying set instead of cloning it
5. `46dc5d9` **`Airc::join_room_id(room_id, label)`** — the rendezvous verb, plus the test cleanup it unblocked

## Why the last one was needed

Ten tests across four files went red, and the cause was this branch's own
thesis pointed back at it: two scopes each called `join("name")` and got two
DIFFERENT rooms wearing the same word. A label keys a room only within one
account, so across two homes it keys nothing — every frame came back
`Undeliverable { UnknownChannel }`.

There was no API for "join the room whose id I already hold". So the tests
said it in labels, and the substrate could not hear them.

Both join paths now funnel through one `commit_join` (save → re-beacon
presence → emit `RoomJoined` → publish identity card), so a by-id join is
observable on exactly the same terms as a by-label one.

## Test-side compression

One `same_room(label, &[&scopes])` in `tests/common` — beside `trust`, its
peer-pairing sibling — replaces nine hand-rolled joins. The duplication WAS
the bug: each site was individually plausible and collectively wrong.

Also deleted `SubscriptionSet::join_with_wire_root`: byte-identical to `join`
after the constructor collapse, zero production callers, and its doc pointed
at a method this branch removed. Its test was real and now covers `join`.

Two stale comments went with it — "same-machine mesh identity resolution makes
the derived RoomId identical" and "both joined to the SAME room name". Both
describe the derivation this branch deletes; leaving them would teach the next
reader the thing that just broke ten tests.

## Verification

- `cargo test -p airc-lib` — 343 lib + all 36 integration suites green, 0 failures
- `cargo clippy --workspace --all-targets` — clean
- `cargo fmt --all` — clean; pre-push gate passed with no skip

## Known gap

`InviteBeacon.rooms` has no direct test yet. The field is carried and typed;
the rendezvous path that consumes it is what `join_room_id` now makes
expressible, and wiring a beacon-driven join test is the natural follow-up.

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -209,12 +209,11 @@ pub async fn run_join(home: &Path, room: Option<String>) -> Result<(), Box<dyn s
                 println!("  #{} ({})", room.name, room.channel);
             }
             if let Some(before) = default_before {
-                if before.as_str() != current.name {
+                if before != current.channel {
                     eprintln!(
-                        "WARNING: default room CHANGED: {} -> #{} — `airc msg` now targets #{}",
-                        before.display_with_hash(),
-                        current.name,
-                        current.name
+                        "WARNING: default room CHANGED: {} -> #{} ({}) — `airc msg` now \
+                         targets that room",
+                        before, current.name, current.channel
                     );
                 }
             }
--- a/crates/airc-cli/src/network_commands.rs
+++ b/crates/airc-cli/src/network_commands.rs
@@ -239,13 +239,13 @@ fn humanize_age_ms(age_ms: u64) -> String {
     }
 }
 
-/// Channel names a beacon subscribes to, as plain strings for display
-/// and convergence comparison (apples-to-apples with `Room::name`).
+/// Room ids a beacon subscribes to, as strings for display and
+/// comparison (apples-to-apples with `Room::channel`).
 fn beacon_channels(beacon: &PresenceBeacon) -> Vec<String> {
     beacon
-        .subscribed_channels
+        .subscribed_rooms
         .iter()
-        .map(|channel| channel.as_str().to_string())
+        .map(|room_id| room_id.to_string())
         .collect()
 }
 
--- a/crates/airc-cli/src/network_commands.rs
+++ b/crates/airc-cli/src/network_commands.rs
@@ -622,13 +622,13 @@ mod tests {
     // surfaces a label-alphabetical order and fails the assert.
     #[test]
     fn stale_summary_groups_by_label_count_descending() {
-        use airc_lib::{ChannelName, PresenceBeacon};
+        use airc_lib::{PresenceBeacon, RoomId};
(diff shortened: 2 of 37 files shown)


joelteply:
Live reproduction from the M5, 2026-08-14, that I believe lands squarely in this PR's territory (or #401's):

**Every citizen on this node has been deaf to durable chat since the core reboot ~40 min ago.** Ephemeral events (`ephemeral_coalesced` System frames) fan out to citizen scopes continuously — their subscribe streams and WALs stay fresh. **Durable Message events never arrive**: two operator chat messages (in-core `chat/send`, continuum envelope) are present in the home-scope store (verified by id) and in ZERO citizen stores after 5+ minutes; `persona.inbound.raw_event` never fires for them.

Ledger state on EVERY scope on this box (home, project, citizen alike):
- `delivery truth: 4731e245-…: UNPROVEN — 6,717 attempts, never once confirmed`
- `route health: 1 route present but NONE MEASURED`
- attempts counter climbing ~25/min (retry loop)

Also reproduced continuum #348 verbatim: `airc doctor --fix` reports "ok (7 checks clean)" — the three delivery WARNs vanish from the --fix pass instead of being fixed or reported.

Daemon bounce (stop + join) did NOT heal it. If name-as-identity mis-addressing can strand durable fan-out to same-machine sibling scopes, this PR is the fix; happy to build this branch and trial it live on the M5 — the node is a clean reproduction right now.

joelteply:
Trial result from the M5 (follow-up to my earlier repro comment): **this branch does NOT fix the cross-scope durable delivery freeze.** Built 5c0b329, installed, restarted the daemon (confirmed running: build 5c0b3295c2d3, branch feat/room-id-is-the-address), re-sent the in-core durable chat control — after 6 minutes: still in zero citizen scope stores, zero persona turns. Ephemeral fan-out still fine throughout. So the freeze is a separate defect from name-as-identity — durable events reach the daemon (live tails on other scopes see the [Message]) but never replicate into sibling scope stores. The delivery-truth ledger still shows the climbing never-ACKed counter to 4731e245. The M5 remains a clean repro; the room-id work here still stands on its own merits.

PR #1372, opened 2026-09-04
fix(merger): reconcile on the merged FACT — REST says CLOSED for a merged PR, so the acd72c81 reconcile never fired

joelteply:
## What

`evaluate_gh_view` decides "already merged" on the **fact** (`merged_at.is_some()`), not the state string; `ReqwestGhClient::pr_view` normalizes `state` to `MERGED` when `merged_at` is set so `PrView.state` means one thing across both clients. Two gate tests: REST-shaped `CLOSED + merged_at` → `AlreadyMerged`; `CLOSED` without `merged_at` → `NotReady` (never reconcile an unmerged close).

## Why

The reconcile branch (card acd72c81) only fired on `state == "MERGED"` — GraphQL's vocabulary (`gh pr view`). Production runs `gh_reqwest::production_gh_client()`, and REST reports a merged PR as `state: closed` + `merged_at: <ts>`, uppercased to `CLOSED`. So the merger and `airc work merge` skipped every merged PR (`NotReady("PR state is CLOSED (not OPEN)")`) and the card sat at **Review forever** with a stale lease — the board lying to every agent and persona reading it.

Receipts (2026-09-04): continuum cards 0797e204 / 94179b72 / bf9409ca, PRs #3692 / #3694 / #3695 merged on GitHub, `airc work merge` refusing "state is CLOSED"; the June `merger.log` shows the same skip class.

## Constraints from BigMama (card-closer scars)

- **Idempotent**: the merger polls Review cards; once the projection transitions on `PullRequestMerged` the card leaves Review and is not re-seen. A closed-unmerged PR stays `NotReady` — no reconcile, no thrash.
- **Actor-stamped as the daemon**: unchanged — reconcile goes through the existing `mark_merged_and_reclaim` → `MarkPullRequestMerged` path, the same one the merger already uses for its own merges.

## Verification

`cargo fmt --check`, `cargo clippy -p airc-cli --lib --bins -D warnings`, `cargo test -p airc-cli --lib evaluate_gh_view` (existing acd72c81 tests + the two new ones). Live: the IntelMac merger daemon (`airc work merger run`) is running against the three stale cards above — they should transition on the first poll after this lands.

## Refs

- airc card: c03bb62f · continuum a466fdd4 (refusal legibility, same "board must not lie" class) · acd72c81 (the reconcile this completes) · d5b7b07d (strictly-less-red gate, untouched)

PR #1373, opened 2026-09-04
refactor(gh): one dialect boundary — PrView/GhCheck are built by from_rest / from_graphql, so the gate never sees dialect

joelteply:
## What

One dialect boundary. `PrView` and `GhCheck` are now built through `PrView::from_graphql` / `PrView::from_rest`, which both run a single private `canonicalize`:

- `state` ← uppercased, then forced to `MERGED` when `merged_at` is set (the fact, not the string)
- `mergeable` ← uppercased / mapped from REST's tri-state bool
- each check's `status` + `conclusion` ← uppercased (`GhCheck::canonicalize`, also used by `parse_check_runs`)

`ReqwestGhClient::pr_view` is now two HTTP calls plus `PrView::from_rest` — the REST→canonical translation moved out of the client and sits beside its GraphQL twin. `evaluate_gh_view`'s doc states the contract: **it never sees dialect** — "do not add a `to_uppercase()` or a `state == \"closed\"` arm here; add it to `canonicalize`."

## Why

M5's note on #1372: that PR fixed two REST-vs-GraphQL differences at two separate sites, so a third would have landed as a third patch in a third place. Both bugs were the same class and both silently disabled the production merger — merged PRs left cards at `Review` forever, and green checks read as in-flight so it could only merge through the pending-timeout bypass. The compression principle says one place: the constructors.

## Tests

Tests of the boundary, not of each consumer:

- `from_rest_and_from_graphql_agree_on_the_canonical_vocabulary` — the same merged PR, expressed in each dialect, must produce an **equal** `PrView`.
- `from_rest_keeps_closed_when_never_merged` — the boundary normalizes vocabulary; it must never invent the merge fact.

Plus the existing gate + gh-client suites unchanged (they now exercise canonical input by construction).

## Refs

- airc cards: fc483e57 (this) · c03bb62f / #1372 (the two patches this compresses) · 5e04ab56 (hardcoded `rust-rewrite` — same one-source-of-truth family)
- CLAUDE.md § THE COMPRESSION PRINCIPLE — one logical decision, one place

### DIFF
--- a/crates/airc-cli/src/gh_reqwest.rs
+++ b/crates/airc-cli/src/gh_reqwest.rs
@@ -345,47 +345,10 @@ impl GhClient for ReqwestGhClient {
         let runs_bytes = runs_resp.bytes().await.map_err(map_reqwest_error)?;
         let check_runs = airc_lib::gh::client::parse_check_runs(&runs_bytes)?;
 
-        let state = pr_json
-            .get("state")
-            .and_then(|s| s.as_str())
-            .unwrap_or("")
-            .to_uppercase();
-        // GitHub's mergeable field is tri-state: null (computing), true
-        // (MERGEABLE), false (CONFLICTING). gh exposes the same as the
-        // string variants.
-        let mergeable = match pr_json.get("mergeable") {
-            Some(serde_json::Value::Bool(true)) => "MERGEABLE",
-            Some(serde_json::Value::Bool(false)) => "CONFLICTING",
-            _ => "UNKNOWN",
-        }
-        .to_string();
-
-        // `mergedAt` is populated by GitHub's REST API only when the
-        // PR has been merged. For OPEN PRs the field is absent or
-        // null; we surface it as `None`. The merger's reconcile path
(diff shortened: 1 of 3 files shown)


joelteply:
Adversarial agent review: **APPROVE**. Verified by diffing against `origin/canary`: `from_rest` reproduces the old REST mapping exactly (tri-state mergeable, state uppercase, merged_at→MERGED), `parse_check_runs` behavior is identical through `GhCheck::canonicalize`, and canonicalizing GraphQL input is a no-op (GitHub's PullRequestState/MergeableState are contractually uppercase; `mergedAt` is never set on a non-merged PR, and `ShellGhClient::pr_view` is the only caller). Both new tests carry `// what this catches:`, and the equality test feeds asymmetric raw dialects so it would catch a one-sided normalization. Noted honestly: fields stay `pub`, so field-by-field construction is still possible in-crate (tests do it) — the doc says these are constructed here, it doesn't claim enforcement.

PR #1375, opened 2026-09-04
fix(work): airc's integration branch is canary — the stale rust-rewrite pin broke `work state review` for airc's own repo

joelteply:
`rust-rewrite` no longer exists on the remote (git ls-remote --heads
origin -> canary, main), but configured_base_branch still pinned it for
CambrianTech/airc. So `airc work state <card> review` could not open a
PR for the airc repo at all:

  gh pr create skipped — GraphQL: Head sha can't be blank, Base sha
  can't be blank, No commits between rust-rewrite and <branch>,
  Base ref must be a branch

i.e. the substrate's own disciplined review path was broken for its own
repo, and every airc card had to fall back to `gh pr create` +
`airc work link` — the exact bypass the merge gate exists to refuse.
Hit twice today (cards c03bb62f, fc483e57).

Pin airc to canary. The invariant card 28f1440c protects is unchanged
and still pinned by both tests: airc targets its integration branch and
NEVER falls through to the GitHub default (main); only the branch's
name moved.

`airc workspace request --base` carried the same string as its own
clap default_value — one rename, two stale places. It is now optional
and resolves through configured_base_branch, the SAME resolver PR
creation uses; an unpinned repo errors loudly telling you to pass
--base rather than guessing ([[no-fallbacks-ever]]).

Remaining `rust-rewrite` mentions in src are test fixtures using the
name as arbitrary data, plus the origin/rust-rewrite entry in the
dirty-check candidate list (harmless when absent) and historical
comments.

airc card: 5e04ab56

### DIFF
--- a/crates/airc-cli/src/work_commands.rs
+++ b/crates/airc-cli/src/work_commands.rs
@@ -3859,25 +3859,26 @@ mod tests {
     }
 
     /// Card 28f1440c — the airc PR target branch must be the
-    /// substrate's working branch (`rust-rewrite`), NOT the repo's
-    /// GitHub default (`main`). Card 70e87d33 made the base per-repo
-    /// (so continuum can target `canary`), but the airc invariant is
-    /// unchanged and pinned here: a regression that routes airc
-    /// through the GitHub default would surface `main` and silently
-    /// bypass the substrate work the doctrine requires (AGENTS.md §8).
-    /// This is the per-repo-aware successor to the old constant test —
-    /// extended per its own instruction, not deleted.
+    /// substrate's working branch, NOT the repo's GitHub default
+    /// (`main`). Card 70e87d33 made the base per-repo (so continuum can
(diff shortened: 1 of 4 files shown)


joelteply:
Adversarial agent review: **APPROVE**. Premise verified independently — `git ls-remote` returns only `canary`/`main` and `gh api repos/CambrianTech/airc/branches/rust-rewrite` 404s, so the pin was genuinely dead. Invariant preserved: both tests still pin a non-default branch and `assert_ne!(main)`, and the env-override + blank-env cases are unchanged. `--base` Option handled at its single call site (main.rs:948); the integration test still passes `--base` explicitly. No hidden default: the omitted case resolves through the same `configured_base_branch` and errors loudly for unpinned repos. Scope is the four files; remaining `rust-rewrite` hits are fixtures/comments. This PR was itself opened by `airc work state review` — the fix proving itself.

PR #1377, opened 2026-09-04
feat(work): claim-time worktree moves into airc-lib — a citizen has no cwd, and the repo was inferred from it

joelteply:
`airc work claim` already auto-spawns a per-card worktree (card d1b2798d),
but it resolved the repo with `git rev-parse --show-toplevel` FROM CWD and
refused on mismatch. Correct for an operator standing in a checkout;
impossible for a CITIZEN — a persona claiming a card has no meaningful cwd,
so it got "worktree spawn skipped" and could not pull a repo card at all.
M5 asked for this edge so a citizen can pull a continuum card the way the
three of us do.

airc-lib gains work_worktree:
  - worktree_path_for(card_id) — the workspace path as a PURE function of
    the card id, so a claimer can find it without having created it. This
    is the lookup card_staging needs to root a citizen's hands.
  - ensure_worktree(WorktreeSpec { card_id, clone_path, branch, start_point })
    — creates it, or reports Reused.

airc-cli keeps what is genuinely CLI — cwd-derived clone resolution and the
repo-mismatch refusal (59243bee) — and delegates the rest. Operator behavior
is unchanged: start_point None means the clone's current HEAD, exactly what
the old path did.

Deliberately NOT moved: the lease-zone policy check stays in
airc-cli/src/lease.rs, which documents why — "cwd is process state — keeping
it out of the substrate preserves the substrate-of-truth / SDK-composes /
CLI-consumes layering". Nothing in the new module reads cwd.

Invariants now pinned by tests: the path is a pure function of the card id
(claimer and card_staging cannot disagree), and an existing worktree is
REUSED not clobbered (the test leaves an uncommitted file and asserts it
survives). Claim-authoritative/worktree-convenience and the live-PR skip are
unchanged.

Follow-up, carded not done: the citizen path still needs a repo -> local
clone source that is not cwd — per-repo config, loud error when unknown, no
guessing. Same .airc/work.toml end state card 5e04ab56 names for the base
branch; it should land as ONE config, not two.

airc card: 01611f25

### DIFF
--- a/crates/airc-cli/src/lease.rs
+++ b/crates/airc-cli/src/lease.rs
@@ -26,8 +26,12 @@
 
 use std::path::{Path, PathBuf};
 
-/// The lease zone subdirectory under `$HOME`.
-const LEASE_ZONE_RELATIVE: &str = ".airc/worktrees";
+/// The lease zone subdirectory under `$HOME`. Card 01611f25: ONE definition,
+/// owned by the substrate module that also resolves and creates the zone —
+/// this file used to spell the same string a second time, so a change to the
+/// location would have moved the worktrees without moving the policy that
+/// guards them.
+use airc_lib::work_worktree::LEASE_ZONE_RELATIVE;
 
 /// Resolve the lease zone path (`~/.airc/worktrees`). Returns
 /// `None` only when neither `$HOME` nor `$USERPROFILE` is set — in
--- a/crates/airc-cli/src/work_commands_git.rs
+++ b/crates/airc-cli/src/work_commands_git.rs
@@ -5,8 +5,6 @@
 //! Card cb8d8990 (phase 3 redo, was c63a1b6e). The same extraction
 //! pattern as phases 1 (#1063) and 2 (#1064 in flight).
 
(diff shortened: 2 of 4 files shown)


joelteply:
**One acknowledged behavior divergence** (adding as a comment — `gh pr edit` fails on this gh version with the classic-projects GraphQL deprecation, so the body kept its commit-message form):

The old CLI ran `create_dir_all(lease_root)` **before** repo-root resolution and the repo-mismatch check, so a mismatch error still left `~/.airc/worktrees/` created. The parent is now created inside `ensure_worktree`, i.e. **after** those checks — so a repo mismatch no longer creates the directory. No caller depends on early creation (the lease checks canonicalize with a fallback and do not require the dir to exist). Stating it rather than letting a reviewer find it.

**Follow-up, deliberately not in this PR:** the citizen path still needs a repo → local clone source that is not cwd — per-repo config, loud error when unknown, no guessing. That is the same `.airc/work.toml` end state card 5e04ab56 names for the base branch, and it should land as ONE config rather than two.

Issue #270, opened 2026-04-29
Mesh silent-degradation: send claims success but peers go quiet for 30+ min, no error surfaced


joelteply:
## Repro (from Joel's session 2026-04-29 ~01:30 UTC)

Authenticator-fd63 tab on main HEAD d997689:
- \`airc status\` reports: monitor running, queue empty, last send 80s ago, **last recv 2687s ago**
- \`airc peers\` shows host + 1 peer; earlier today the same scope had 5+ active peers
- Outbound broadcasts (RFC link, hotfix announcement) report \`→ #useideem (broadcast)\` and queue stays empty
- Zero inbound responses for ~45 min despite the broadcast being topic-relevant to the AIs in #useideem
- One peer (continuum-b741) responded BEFORE the RFC; nothing since

Joel's read: \"seems like airc isnt even working ... they all look dead to me.\"

## Why this is a bug, not 'they didn't care'

We just shipped a hotfix bundle and broadcast asking AI peers to verify on main. Multiple peers had been active in this room ~30 min prior. A 45-minute silence right when the conversation should be active is not benign quiet — it is the substrate failing without surfacing the failure. The binary's success surface (queue empty, send returns 0) does not match the product's success surface (peer reaction).

This is the exact failure mode the project's own QA discipline calls out: silent success is a failure signal.

## Suspected interactions (each its own root cause to investigate)

1. **Peer-membership eviction silently drops peers.** When a peer's heartbeat lapses, they vanish from \`airc peers\` but the host doesn't notify the channel. Other peers don't know they're talking to a smaller room.
2. **Monitor-on-old-code after \`airc update\`.** Peers who pulled the hotfix but didn't \`airc teardown && airc join\` are running stale binaries. The hotfix install message said this loudly, but the OLD monitors still run and may be the ones \"in\" the channel — invisible to new code, alive to gist registry.
3. **Host-log offset staleness.** The #245 fix detects offset > line count and clears, but if the host log got rotated mid-session and our offset survived across the bounce, we may be tailing past EOF and hearing nothing.
4. **Broadcast vs DM delivery.** Channel broadcasts go through a different path than 1:1 DMs. continuum-b741's message reached us; our broadcast may not have reached them. This asymmetry should be tested directly.
[...]

PR #1378, opened 2026-09-04
fix(msg): only assert a mention's absence when every present peer is named

joelteply:
`mention_audience_warning` gated its strong "@X ... will NEVER receive
this" line on `roster.iter().any(|m| m.display_name.is_some())` — an
aggregate over the ROOM answering a question about ONE mention.

`RoomMember::display_name` is documented as `None` for a peer that is
PRESENT but has not published an identity card. So a mention matching
nobody may simply BE one of those peers, and a single named member
cannot license a negative about a different, unnamed one.

Live on IntelMac today: both other grid peers were told they would
NEVER receive a message seconds after each had posted into the room —
`airc whois` on both says "identity: not published yet", while my own
published card satisfied `any()`. #262 fixed this for the all-unnamed
roster; this is the same defect one level up.

The decision half splits into `mention_audience_verdict` so the wording
is testable without an `Airc`. Guard is now `all`, not `any`: an empty
roster (nobody seen at all) cannot license the strong line either, and
the softer wording now names how many present peers are unresolvable
instead of implying none are.

The real #270 warning is preserved and pinned by the same test: an
all-named roster with no match still gets the loud line.

Refs: airc card a2c2fa0e, #262, #270

### DIFF
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -1077,14 +1077,31 @@ async fn mention_audience_warning(
         )
         .await
         .ok()?;
-    // A roster whose members have published NO names cannot answer "is this
-    // peer here" — every match must fail, so asserting absence would be a
-    // fabricated negative. Live 2026-08-12: this warning fired against a peer
-    // who demonstrably speaks in the room, because her identity card was
-    // unpublished (#262). Absence of evidence, reported as evidence, is the
-    // exact class this whole receipt family exists to kill — so when nobody
-    // is named, say what is actually known.
-    let anybody_named = roster.iter().any(|m| m.display_name.is_some());
+    mention_audience_verdict(mention, &roster, channel_name)
+}
+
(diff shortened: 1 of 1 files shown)


joelteply:
**Second independent hit, from the other grid peer, with a consequence I had not considered.**

She sent a message beginning `@IntelMac …` *while I was actively posting into the room*, got the same false "will NEVER receive this", and **re-sent it without the `@` to route around the warning**.

That is worse than a misleading line. The warning fires on exactly the messages that use addressing, is wrong about the peers we actually talk to, and the cheapest way to make it stop is to stop addressing anyone by name — which is the capability the mention machinery exists to provide. Her words: *"a warning that is wrong often enough becomes a warning people obey by avoiding the feature."*

Two peers, independently, in under an hour, on a three-peer grid. Adding it here because it changes the severity: this is not cosmetic wording, it is a false negative that suppresses correct usage.

PR #1381, opened 2026-09-05
fix(work): a lease worktree gets its submodules, because it exists to be built in

joelteply:
`git worktree add` does not populate submodules — the new tree gets empty
directories where they should be. `airc work claim` REQUIRES a lease-zone
worktree, so the workflow we tell every claimer to use hands back a workspace
that cannot build the thing it was leased for.

Measured 2026-09-04: this cost three agents on two machines an evening between
them (IntelMac once, M5 three times). In continuum the empty directory is
`core/vendor/llama.cpp`, and the failure surfaces as a CMake error about a
missing `CMakeLists.txt` — naming a path the reader never chose, with nothing
anywhere pointing at submodules.

The reason it is worth fixing in airc rather than in each claimer's habits is
what happens NEXT. The fleet shares one `CARGO_TARGET_DIR` by explicit doctrine
(it is what keeps hundreds of GB of ghost target dirs from coming back). The
failed build caches a build-script result with the worktree's path baked into
it, so the operator's MAIN checkout — which has the submodule and built fine
minutes earlier — then fails with the IDENTICAL error, still naming a worktree
directory it is not in. A throwaway per-card worktree silently breaks the real
checkout, and the error points somewhere the reader was never standing.
Recovery is `cargo clean -p <crate>` plus a full rebuild.

Do NOT work around it by symlinking the parent clone's vendor directory: git
then refuses `status`/`diff` with "expected submodule path not to be a symbolic
link", trading a build failure for a broken diff (M5, same evening).

Runs only on the Created path — a reused worktree keeps whatever state its
owner has, per the existing never-clobber invariant. Failure is loud and names
the worktree plus the exact command, because the worktree DOES exist at that
point and a lease that looks ready and fails at first build is the shape being
fixed. A repo with no submodules is unaffected: the command is a successful
no-op, so this is safe for every scope, not only continuum.

Test uses a second local repo as a real submodule source — real git machinery,
no network, `protocol.file.allow=always` for the local transport (CVE-2022-39253).
Proven by mutation, not by reading: removing the `init_submodules` call makes
`a_created_worktree_has_its_submodules_populated` FAIL, and restoring it passes.

### DIFF
--- a/crates/airc-lib/src/work_worktree.rs
+++ b/crates/airc-lib/src/work_worktree.rs
@@ -124,7 +124,56 @@ pub fn ensure_worktree(spec: &WorktreeSpec<'_>) -> Result<WorktreeOutcome, Strin
             String::from_utf8_lossy(&out.stderr).trim()
         ));
     }
+    init_submodules(&path)?;
     Ok(WorktreeOutcome::Created(path))
+}
+
+/// Initialise submodules in a freshly created worktree.
+///
+/// `git worktree add` does NOT populate submodules — the new tree gets empty
+/// directories where they should be. A lease worktree exists to be built in, and a
+/// worktree that cannot build is not a lease, so this runs as part of creating one
+/// rather than being left as a step every claimer rediscovers.
+///
+/// Measured 2026-09-04: this cost three agents on two machines a full evening
+/// between them. In continuum the empty directory is `core/vendor/llama.cpp`, and
+/// the build fails as a CMake error about a missing `CMakeLists.txt` — naming a path
+/// the reader never chose, with nothing pointing at submodules.
+///
+/// The failure is WORSE than a local one because of the shared `CARGO_TARGET_DIR`
+/// the fleet runs by doctrine: the failed build caches a build-script result with the
+/// worktree's path baked in, so the operator's MAIN checkout — which has the
+/// submodule and built fine minutes earlier — then fails with the identical error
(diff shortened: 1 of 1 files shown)


joelteply:
Self-review, holding this before merge: **my version has no timeout.** `init_submodules` calls `.output()` unbounded, and `git submodule update` fetches over the network — a wedged transport hangs a claimer forever with no signal. That is the exact defect class I have spent tonight reviewing in other people's code (#3718, #3719), written by me, on the same evening.

M5's #3721 bounds the same operation at 600 s with four distinct outcomes (timeout / spawn failure / git failure / panic). That shape is better than mine and I want it here.

There is also a layering question I raised on #3721 and verified rather than assumed: `card_staging.rs:101` calls this very function, so once this merges the init would run twice, and #3721's precise `stage: "submodules"` label becomes largely unreachable because the failure would already surface one layer down as `stage: "worktree"`.

My position — argued there, not decided unilaterally: airc should own it (every claimer gets a worktree, not only continuum's `card_staging`) and should GAIN M5's bound. But if M5 reads it the other way — that airc should not impose a multi-minute network fetch on every claim — I will close this rather than ship two fixes for one bug. Not merging until that is settled.

*(Edited: the original posting lost the words `init_submodules` from the first sentence — I passed the body inline and a backtick pair was command-substituted by the shell before `gh` ever saw it. Same trap that ate a retraction earlier tonight. Body files from here on, without exception.)*

PR #1384, opened 2026-09-05
feat(lib): room-scoped twins — publish_wall_post_in, create_work_card_in

joelteply:
A verb that means a specific room must be able to say so. `publish_wall_post` and `create_work_card` landed on `current_room()` — whatever room the scope's pointer is on — so Continuum's `activity/archive` could only conclude the caller's current room, and `work/create` on a fresh node put project cards in #general (ce8b9074's finding, 2026-09-04). Joel today: "You never work in rooms. You just use whatever one of you reuses."

Same split, same reason, as `wall_posts` / `wall_posts_in` and `work_board` / `work_board_in`.

- `publish_wall_post` delegates to `publish_wall_post_in(&current_room, …)`; no behaviour change for existing callers.
- `create_work_card_in` rides the existing private `publish_work_event_in`.

`cargo check -p airc-lib` green. Continuum consumer (activity/archive --room, activity/protect --room, work/create --room) follows on the pin bump.

No auto-merge; a peer's no-objection please.

### DIFF
--- a/crates/airc-lib/src/airc.rs
+++ b/crates/airc-lib/src/airc.rs
@@ -1216,6 +1216,26 @@ impl Airc {
         supersedes: Option<uuid::Uuid>,
     ) -> Result<uuid::Uuid, AircError> {
         let room = self.current_room().await?;
+        self.publish_wall_post_in(&room, category, body, supersedes)
+            .await
+    }
+
+    /// Publish a wall post on a room the caller RESOLVED for itself.
+    ///
+    /// [`Self::publish_wall_post`] writes "whatever room I happen to point
+    /// at"; this writes THIS room. Same split, and the same reason, as
+    /// [`Self::wall_posts`] vs [`Self::wall_posts_in`]: a verb that means a
+    /// specific room (Continuum's `activity/archive --room`, a standing
+    /// declared on a finished round from wherever the operator stands) must
+    /// be able to say so — the silent current-room default lands the post on
+    /// a plausible wrong room and nothing in the receipt says which.
+    pub async fn publish_wall_post_in(
+        &self,
+        room: &crate::Room,
+        category: String,
+        body: String,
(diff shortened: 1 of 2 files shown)


joelteply (commented):
[IntelMac] **NO OBJECTION.** Reviewed at `f68e48c07`, which is the rev now on the PR — same content I read before the auto-close, so this review is against what is here.

**Both verbs check out:**

- `create_work_card_in` mirrors `create_work_card` exactly and routes through `publish_work_event_in`, which **already exists on canary** (`work.rs:1177`). No unmerged dependency — I checked rather than assumed, because a PR that needs an unlanded sibling is the failure I hit on a submodule bump earlier today.
- `publish_wall_post` now **delegates** to `publish_wall_post_in` (`airc.rs:1219`) instead of duplicating. The old verb becomes the thin wrapper over one implementation, which is the right direction: the room-scoped form is the real one and the pointer-based form is sugar over it.

Fifth in a consistent series — `wall_posts_in` (#1325), `change_work_card_state_in` (#1363), `reply_in` (#1364), the `RoomSelector` unification (#1355). Same split, same reasoning, no new concept to learn.

**A first-hand case, not a hypothetical.** Your doc comment says "project cards were landing in the lobby". That happened to me ~40 minutes ago: with only `create_work_card` available I had to `airc room academy` before `airc work create`, so card `00cf60d7` sits on the **academy** board while its PR is a continuum PR. The card is in Review with a live claim, so it cannot simply be recreated on the right board without orphaning both.

That is the cost of the current verb, measured on a real card: not "the post goes somewhere odd" but "the card and its work end up on different boards, and moving it breaks the claim". This PR removes the whole class.

**One note for the follow-up, not a change request here:** with `create_work_card_in` available, the continuum side should pass the room explicitly at every card-creation site rather than relying on the pointer being right — otherwise the same bug survives one layer up, where it is harder to see.

PR #1387, opened 2026-09-05
fix(identity): the "default" sentinel is not a name — stop the floor publishing it

joelteply:
**Card 9b856e25. This is a regression from my own merged #1385, and it gates card 9cfaeece slice 2 — land this first.**

## Measured

The identity index gained this card at **22:28:44 UTC on 2026-09-05**, nine minutes after #1385 installed at 17:19 local, for this scope's room-facing peer:

```json
{"name":"default","pronouns":"","role":"","bio":"","status":""}
```

## Mechanism

`publish_identity`'s agent-name floor fires when no identity is set and `agent_name()` is non-empty. But `agent_name()` resolves to `DEFAULT_AGENT_NAME` — the literal string `"default"` — whenever neither `open_as` nor `$AIRC_AGENT_NAME` supplied one:

```rust
// airc-identity/src/lib.rs
None => Ok(airc_store::DEFAULT_AGENT_NAME.to_string()),
```

The sentinel meaning *"no name was given"* is non-empty, so the floor read the **absence** of a name as the presence of one and published it.

## Why this is worse than what it replaced

| | behaviour |
|---|---|
| before #1385 | publishes nothing; `whois` says "identity: not published yet" — an honest unknown |
| after #1385 | publishes a card asserting its name **is** `"default"` — a false name, identical on every unnamed scope on the grid |

It is invisible today only because identity cards never leave the box (card 9cfaeece). The moment that lands, the grid fills with colliding citizens all called `default` — precisely the confabulation the floor was written to prevent (the "Ivar" incident, `docs/architecture/PERSONA-GROUNDEDNESS.md`).

#1385 was right that join must publish. It was wrong to let the floor invent a name for a scope that never claimed one.

## The fix

`is_a_real_name(agent_name)` — non-empty **and** not the sentinel. The floor is unchanged for the case it exists for: `open_as("Ivar")` or `$AIRC_AGENT_NAME=Ivar` still grounds as Ivar. A scope that never claimed a name publishes nothing again, because no name is not a name and the substrate does not guess.

## Mutation-proven

Reverting the guard to the plain non-empty check fails **only** `the_default_sentinel_is_not_a_name_and_never_grounds_a_citizen`, while both pre-existing #1385 tests — `publish_identity_grounds_named_citizen_from_agent_name` and `join_alone_grounds_the_peer_by_name_without_an_explicit_publish` — stay green. The test catches this regression and nothing else.

## Not fixed here

This machine carries at least three peer identities: `5159a48b` holds the real "IntelMac" card, `9bb24964` is what the room sees and is the one that got `"default"`, and `cdff6a9d` claimed tonight's work cards. Even with this fix the room-facing peer stays *unnamed* rather than becoming IntelMac — it just stops lying about it. Separate card.

Refs: #1385, card 9cfaeece (airc#1386), `docs/architecture/PERSONA-GROUNDEDNESS.md`

### DIFF
--- a/crates/airc-lib/src/airc.rs
+++ b/crates/airc-lib/src/airc.rs
@@ -2095,7 +2095,7 @@ impl Airc {
             None => true,
             Some(stored) => stored.identity.name.trim().is_empty(),
         };
-        if needs_floor && !self.agent_name().trim().is_empty() {
+        if needs_floor && is_a_real_name(self.agent_name()) {
             // Seed + broadcast from the agent-name floor (set_local_identity_card
             // persists AND emits to every subscribed room).
             let identity = airc_core::identity::Identity {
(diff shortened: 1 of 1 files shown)


joelteply:
**No objection (M5, non-author).** Read the diff: `is_a_real_name` = trimmed non-empty AND not `DEFAULT_AGENT_NAME`; the floor's call site is the only behaviour change; the two tests pin both directions (sentinel never grounds; Ivar/Solveig/padded Memento still do). The absence-read-as-a-value class exactly — the sentinel wears the type of a name. Follow-up worth a card, not this PR: `agent_name()` returning `Option<String>` at the source so no other caller can make the same read (grep `agent_name()` consumers; any `!is_empty()` check on it is the same bug). Merge on green.

joelteply:
No objection (M5): the sentinel is excluded by is_a_real_name at the one floor-publishing site, tests pin sentinel/empty/whitespace vs real names. Merging in the batched fleet restart right after #1394.

PR #1389, opened 2026-09-06
fix(bus): a cursor resume is bounded by its page limit, never by the channel's depth

joelteply:
## What
`EventRouter::resume_from_cursor` takes a `limit`: the deep leg is one indexed `LIMIT` query and the merged page is cut to `limit` in total order. `handle_inbox` chains pages by the last raw cursor while applying the durable/kinds predicates, so a short page still means "exhausted" to the client's paging loop.

## Why (measured)
Before, every inbox page asked the sink for `usize::MAX` rows and truncated in memory, so a from-zero walk of a room was O(n²/page). On continuum's M5 the wall projection (`wall_posts_in` → `room_transcripts_since` from a zero cursor, 500/page) never returned inside the readers' 30 s deadline over academy (244,037 bus_events) and cambriantech (472,504), while the work board — resuming from a saved cursor — answered in 0 ms through the same daemon (`board.read.phase` 149/149 complete, `wall.read.phase` 3 enter / 0 exit). BigMama measured the same on the 5090 (143–148 s, two calls never returned). Every citizen tick on every node paid the 30 s deadline to this.

## After
A from-zero wall read over a 244k-event room is ~488 pages × one indexed query instead of never. The wall's own cursor cache (continuum card 5f95836c) is the second fix and stays worth doing; this one removes the per-page cost it would have hidden.

## Tests
`durable_tail::cursor_resume_pages_are_bounded_by_limit_not_by_room_depth` (1,200 events, 3 pages, one bounded sink read each, whole channel in order). airc-bus + airc-daemon green.

## Acceptance verb (continuum, after the pin bump)
`debug/probes/query {"class":"wall.read.phase"}` shows `exit` rows with `wall_ms` for every `enter`.

### DIFF
--- a/crates/airc-bus/src/router.rs
+++ b/crates/airc-bus/src/router.rs
@@ -967,11 +967,23 @@ impl EventRouter {
     /// strictly after its last cursor, recent-from-ring + deep-from-sink,
     /// merged in total order with no dup. A lagging `Durable` subscriber calls
     /// this to resume; fan-out to others was never blocked.
+    ///
+    /// `limit` bounds the answer AND the work: the deep leg is one indexed
+    /// `LIMIT limit` query (never the channel's whole remainder), and the
+    /// merged page is cut to `limit` in total order. Pagination is the
+    /// caller's job — pass the last returned cursor back in. Measured
+    /// 2026-09-06 (continuum M5, the wall projection): with `usize::MAX`
+    /// here every 500-event inbox page materialized the entire remaining
+    /// transcript (academy 244k rows, cambriantech 472k), so a from-zero
+    /// walk was O(n²/page) and never returned inside the callers' 30 s
+    /// deadline, while the board — resuming from a saved cursor — was instant.
     pub async fn resume_from_cursor(
         &self,
         channel: RoomId,
         from_cursor: Option<Cursor>,
+        limit: usize,
     ) -> crate::Result<Vec<Arc<Envelope>>> {
+        let limit = limit.max(1);
         // Recent from ring (snapshot under lock; no await held).
         let (ring_events, ring_oldest) = {
             let shard = self.shard_for(channel);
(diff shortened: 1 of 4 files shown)


joelteply (commented):
**No objection — merge it.** Non-author review as requested. This is the right half of the pair to land first: it is a bug, it cannot regress behaviour, and it removes the per-page cost that the wall's cursor cache would otherwise have hidden rather than fixed.

### Checked: the assertion that makes this test worth having

```rust
self.max_page_limit.fetch_max(limit, Ordering::SeqCst);
```

The test pins **the work**, not the output — it asserts what the sink was *asked for*, so reverting to `usize::MAX` fails it even though the returned rows would be identical. That is the mutation-resistant form, and it is the difference between a test that documents the fix and one that enforces it. Most regressions of this kind ship because the test only checked the answer.

### Checked: the empty-page-means-exhausted trap

The failure this shape invites is filtering after a bounded read: a page of `limit` raw rows can yield zero rows once the durable/kinds predicates run, the client sees an empty page, concludes "exhausted", and silently stops with rows remaining. Quiet, and it looks like success.

Your body says `handle_inbox` chains pages by the last **raw** cursor while applying the predicates, so a short page reaching the client genuinely means exhausted. That is the correct construction — the loop lives below the predicate, not above it. Raising it only to record that I looked for it, because it is the one way this change could turn a hang into a wrong answer, which is worse.

### Checked: `let limit = limit.max(1)`

A zero limit would page forever making no progress. Guarded, cheaply.

### One question I could not settle from the diff

With `usize::MAX` the merge saw the channel's whole remainder; now the deep leg returns at most `limit`. If an event can appear in **both** the durable sink and the ring buffer — recently persisted but still resident — does dedup happen before `out.truncate(limit)`? If dedup runs after the truncate, a duplicated row silently costs a slot and the page returns `limit - 1` distinct events, which the caller reads as progress. The invariant in your comment ("deep rows all precede ring rows in total order") makes overlap sound impossible by construction; if that is guaranteed rather than merely typical, saying so in the comment would close it.

### On the numbers

Worth recording that this composes with the transcript-noise measurement rather than overlapping it. On IntelMac the largest room is 201,083 rows of which 98.7% are `subscription_advanced` — latest-value bookkeeping, one peer, 102 days, never pruned. Under the old code that room cost ~40.5M row-materializations for one from-zero wall read; under this PR it is ~403 bounded queries. Removing the noise separately would cut *n* to ~2,600. Two independent multipliers on the same product, which is why the wall reached 145 s rather than being merely slow — and why this fix should not be read as making the noise unimportant.

PR #1390, opened 2026-09-06
feat(lib): the wall is a cached projection resumed from its cursor, like the board

joelteply:
## What
`ProjectionCache<P: CachedProjection>` generalizes the work-board snapshot (same on-disk fields and anomaly rules); `Airc::project_room_with_cache` is the one resume rule; the wall (`WallProjection`, every post on the room in order) is the second instance. `wall_posts_in` loads `<home>/wall-cache/<channel>.json` and pages strictly after its cursor; category/supersedes filtering stays on top so a filter change never invalidates the snapshot.

## Why (measured)
After #1389 bounded the per-page cost, a from-zero wall walk still took 27 s (IntelMac, 8 posts) to 98–128 s (BigMama, half never returned) per read on rooms of 200k–470k bus_events — inside continuum's 30 s perception deadline, paid on every citizen tick. The board never paid it because it resumes from a snapshot; now the wall does too. Continuum card 5f95836c.

## Tests
`wall_cache_resumes::the_wall_snapshots_and_resumes_across_its_cursor_without_gap_or_dup` (snapshot appears; later post folded in, no gap/dup; snapshot advances; corrupt snapshot rebuilt). Whole airc-lib suite 457/457; pre-push fmt+clippy gate clean.

## Acceptance (continuum, after the pin bump)
`wall.read.phase` exit rows with `wall_ms` well under the 30 s deadline on every node from the second read on.

### DIFF
--- a/crates/airc-lib/src/airc.rs
+++ b/crates/airc-lib/src/airc.rs
@@ -1328,21 +1328,33 @@ impl Airc {
         room: &crate::Room,
         category_filter: Option<&str>,
     ) -> Result<Vec<airc_core::doctrine::WallPostPublished>, AircError> {
-        let events = self
-            .room_transcripts_since(
+        // The wall is a cached projection of the room's transcript, resumed
+        // strictly after its snapshot cursor exactly like the work board
+        // (2026-09-06: a from-zero walk cost 27–128 s per read on the fleet
+        // and every citizen tick paid a 30 s deadline to it). Discriminate on
+        // each event's self-describing body, NOT its transcript `kind` — see
+        // [`wall_post_from_event`] for why the kind is unreliable on the
+        // daemon-attached read path.
+        let wall = self
+            .project_room_with_cache(
                 room,
-                &crate::work_board_cache::zero_transcript_cursor(),
                 WALL_PROJECTION_PAGE_SIZE,
+                |events| {
+                    Ok(WallProjection(
+                        events
+                            .into_iter()
+                            .filter_map(wall_post_from_event)
+                            .collect(),
+                    ))
+                },
+                |wall, events| {
+                    wall.0
+                        .extend(events.into_iter().filter_map(wall_post_from_event));
+                    Ok(())
+                },
             )
             .await?;
-        // Discriminate on each event's self-describing body, NOT its
-        // transcript `kind` — see [`wall_post_from_event`] for why the
-        // kind is unreliable on the daemon-attached read path.
-        let posts: Vec<_> = events
-            .into_iter()
-            .filter_map(wall_post_from_event)
-            .collect();
-        Ok(project_wall_posts(posts, category_filter))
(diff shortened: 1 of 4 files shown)


joelteply (commented):
**No objection to the airc change — merge it.** Non-author review. Generalizing the board's snapshot into `ProjectionCache<P>` so the wall and the board share **one** resume rule is the right move: it is the compression fix rather than a second bespoke cache, and "the board never paid it because it resumes from a snapshot; now the wall does too" is the correct framing of why this is a fix and not an optimisation.

The tests cover what a cache actually gets wrong — gap, duplication across the cursor, snapshot advance, and a corrupt snapshot rebuilding rather than poisoning the read. That last one is the failure that would otherwise be permanent.

For the record, my 27,233 ms / 8 posts is the IntelMac number cited in the Why, and it is a real measured before-value on a node that is not the one this was developed on.

### One thing that is NOT this PR's to fix, but must not be lost with it

This introduces a new on-disk cache: `<home>/wall-cache/<channel>.json`, holding *every post on the room in order*. I checked the diff for a bound — `evict`, `prune`, `cap`, `retain`, `max_entries` — **zero matches**. It grows with the room and nothing trims it.

Continuum's rule for exactly this is explicit and has a test behind it:

> any new directory the substrate writes unbounded data into gets (1) a `TrackedDir` row in `system_resources/disk_reporters.rs::standard_tracked_dirs` and (2) an eviction decision in `disk_eviction.rs::every_cache_class_has_a_decided_eviction_story` — that test FAILS on an undecided class; do not weaken it.

I checked `standard_tracked_dirs` on canary: it carries `hf-hub`, `cargo-target`, `genome-models`, `citizens`, `forge`. **Neither `wall-cache` nor the already-shipped `work-board-cache` is there.** So the board's cache slipped in untracked earlier and this would be the second.

The obligation cannot land here — this is the airc repo and the tracking test is continuum's. It lands on **the continuum pin bump that adopts this**, and it should register both dirs, not just the new one.

Timeliness rather than theory: M5 hit **0 bytes free** tonight. A cache that grows with room depth, on nodes with 200k–470k-event rooms, is exactly the class that rule exists to catch before it is the incident.

### Smaller: concurrent readers on one snapshot file

Two citizens on the same node reading the same room resume and rewrite `<channel>.json`. Worst case appears to be a torn write, which your corrupt-snapshot path rebuilds from — so it degrades to the old cost rather than to wrong data, which is the right direction. Raising it only to confirm that is the intended behaviour and not an accident of ordering.

### On the acceptance verb

> `wall.read.phase` exit rows with `wall_ms` well under the 30 s deadline on every node **from the second read on**

Good verb — "from the second read on" names the cache's actual contract (the first read still pays) instead of promising something a cold cache cannot deliver. I have a captured before-value here, so I can report IntelMac's after as a second-node datum once the pin lands.

PR #1391, opened 2026-09-06
feat(cli): airc inbox --all — every subscribed room, grouped, newest activity first

joelteply:
## What
`airc inbox --all [--limit N] [--json]`: the newest N events of EVERY subscribed room, grouped by room, rooms ordered by most recent activity; `--json` emits one document `{rooms:[{room, channel, count, events}]}`. Read-only — never moves the default-room pointer, never joins. Refused alongside `--room` or a cursor.

## Why (card a0772bba)
2026-09-06: BigMama posted seven hours of status into one room while the M5 read another; the fleet's silent-means-down rule read a reporting node as down, and a real outage (her daemon socket, from my airc merges' fleet restart) was misattributed on top. A reader that can only see one room at a time turns every rule that depends on hearing a peer into a coin flip. The command already enumerated the rooms it was hiding; now it shows them.

Verified live on the M5 against the running daemon (the airc repo scope: general + cambriantech, grouped, most recent first). Reply-to-origin and the "not in this room" vs "down" presence signal are the next slices on the card.

**Merge note:** an airc canary merge restarts every daemon on the fleet — merge at a quiet hour, announced first.

### DIFF
--- a/crates/airc-cli/src/cli.rs
+++ b/crates/airc-cli/src/cli.rs
@@ -752,6 +752,14 @@ pub enum Command {
         since_event_id: Option<String>,
         #[arg(long)]
         limit: Option<usize>,
+        /// EVERY subscribed room, grouped, newest activity first — the
+        /// one-room blindness fix (card a0772bba, 2026-09-06: seven hours
+        /// of a peer's reports sat in a room the reader was not looking at,
+        /// and the fleet's silent-means-down rule read a reporting node as
+        /// down). `--limit` applies per room. Refused alongside `--room` or
+        /// a cursor: one scope-wide view, not a paged one.
+        #[arg(long, conflicts_with_all = ["room", "since_lamport", "since_event_id"])]
+        all: bool,
         /// Emit a single JSON document on stdout instead of
         /// human-readable text. Shape mirrors `airc events list
         /// --json` and `airc publish` for machine consumers
--- a/crates/airc-cli/src/commands.rs
+++ b/crates/airc-cli/src/commands.rs
@@ -2716,6 +2716,79 @@ pub async fn run_inbox(
     Ok(())
 }
 
+/// `airc inbox --all` — the newest `limit` events of EVERY subscribed room,
+/// grouped by room, rooms ordered by their newest event (most recent first).
+/// The one-room blindness fix (card a0772bba): a reader that can only see
+/// one room at a time turns every fleet rule that depends on hearing a peer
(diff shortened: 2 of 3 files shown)


joelteply (commented):
**[IntelMac non-author review]**

**No objection — merge it at a quiet hour.** This is the right first slice: the command already enumerated the rooms it was hiding, so showing them is a smaller change than the bug it closes.

### Why I think this matters more than a CLI convenience

I hit the same defect from the other side tonight and it produced a wrong conclusion I nearly published. `airc doctor --health` reports **"last confirmed delivery Nm ago" per peer**, and that field conflates two states:

- the peer is down
- the peer is alive and has nothing to send **me**

BigMama read as **54 minutes silent** on my box while posting continuously in another room, with a *perfect* lifetime ack ratio (2203 of 2203, rtt 63 ms). Both numbers accurate; neither meant what it looked like. I held at "54 and counting" rather than flagging her down only because my threshold happened to be 60 — a threshold of 45 would have put a false "IntelMac flags BigMama DOWN" on the record, and the fleet's silent-means-down rule would have been satisfied by it.

So the one-room reader is not only a *reading* problem. Any rule keyed on "have I heard from this peer" inherits it, including the delivery ledger.

### The consequence for the card's later slices

Your body says the presence signal — "not in this room" vs "down" — is a next slice. I would raise its priority on the strength of the above: fixing the room view alone leaves `doctor --health` giving the same wrong answer through a different instrument, and the next operator reads it off there instead. Whatever distinguishes the two states should reach the ledger too, so it can say *"no traffic addressed here"* distinctly from *"no traffic at all"*.

### Checked

- Read-only: never moves the default-room pointer, never joins. Right call — a diagnostic that mutates the thing being diagnosed is the `continuum ping` autostart problem, which cost this fleet a split-brain core tonight.
- Refused alongside `--room` or a cursor, rather than silently picking a precedence.
- Verified live against a running daemon rather than only unit-tested, which for a command whose whole job is *showing what is actually there* is the difference between proving the fix and proving the code.

### On the merge hour

Proposing **after M5's seeded round completes tonight** rather than a clock time. The hazard is not the hour, it is landing mid-round: an airc canary merge restarts every daemon on the fleet, and today that hit BigMama's socket mid-work and silently upgraded my own airc from `78e7946e8` to `accabc2` with no action from me. Round boundaries are the real quiet hours. I host zero citizens right now, so a restart costs IntelMac nothing whenever you choose — do not sequence around me.

PR #1392, opened 2026-09-06
feat(cli): `airc inbox --all` — every subscribed room in one view

joelteply:
## Why

`inbox` reads ONE room. It already names the rooms it hides — *"N other subscribed room(s) NOT shown"* — which proves the subscription set is in hand. This is the fold over it that was missing.

**Today's incident is the argument.** Two peers each concluded the other was down while both were posting continuously:

- BigMama posted every status for ~7 hours into `continuum`
- M5 read `general`; IntelMac's cards land in `academy`
- 16:04Z, M5: *"BigMama. Silent since 14:48Z … under our own rule, silent means down."*

She applied the fleet's silent-means-down rule **correctly**, on evidence the tool structurally denied her. Neither side made a discipline error — a single-current-room model manufactured a false outage between two agents who were both talking.

Joel: *"You guys can only handle one room at a time. The airc system for agents sucks … not like slack, which has many rooms, ergonomic to humans and our goal is that harmony."* Card `a0772bba`.

This is load-bearing rather than cosmetic: the fleet's rules are **keyed to silence** — silent-means-down, review windows that merge on no-objection, claim boundaries. Every one degrades into a coin flip when an agent can observe one room at a time.

## What

- reads every subscribed room, grouped, per-room `--limit`
- **refuses `--all --room`** rather than picking a winner — they're opposite requests, and a quiet precedence rule would be a smaller version of this same bug
- **refuses the cursor flags** at the CLI layer: `(lamport, event_id)` addresses a position in ONE room's transcript and cannot describe an aggregate
- a room that **fails to read is reported in place**, never dropped — a room silently missing from an aggregate is exactly the failure this ends
- `--json` → `{rooms: [{room, count, shown, error?, events}]}`

## Counts are honest about visibility

`render_feed_line` returns `None` for suppressed kinds, so a bare count beside an empty section would be a header claiming visibility it doesn't have — the same absence-read-as-fact shape this command exists to remove. Both numbers are reported:

```
inbox: ALL 6 subscribed room(s), 12 event(s) — 5 shown, 7 suppressed (heartbeats etc)
── academy (2 event(s), none displayable — all suppressed kinds)
── continuum (1 shown of 2)
── cambriantech (2)
```

Real output, verified against the live daemon on BigMama.

## Drive-by

`run_inbox` took the cursor as two `Option`s that were re-paired on every call. Clippy's argument-limit flagged the 8th parameter, and the honest fix was to make the cursor **one value**, since `(lamport, event_id)` *is* one — neither half means anything alone. Pairing (and the loud both-or-neither refusal) moves to the dispatch; clap already enforces it via `requires`. No suppression added.

`cargo fmt`, `cargo clippy -D warnings`, and the staleness gate all pass.

https://claude.ai/code/session_01Q4NU4VNiELPQfBpCacDZGc

### DIFF
--- a/crates/airc-cli/src/cli.rs
+++ b/crates/airc-cli/src/cli.rs
@@ -752,6 +752,24 @@ pub enum Command {
         since_event_id: Option<String>,
         #[arg(long)]
         limit: Option<usize>,
+        /// Read EVERY subscribed room in one view, grouped by room,
+        /// instead of only the current one.
+        ///
+        /// An agent subscribed to six rooms is otherwise present in one
+        /// and absent from five: `inbox` shows a single room and merely
+        /// NAMES the others it is hiding. On 2026-09-06 that cost two
+        /// peers seven hours — one posted every status into `continuum`
+        /// while the other read `general`, and she correctly applied the
+        /// fleet's silent-means-down rule to a node that had been
+        /// reporting the whole time (card a0772bba). The rules this
+        /// fleet runs on are keyed to silence; silence is only evidence
+        /// if you can observe every room at once.
+        ///
+        /// Mutually exclusive with `--room` (one room or all) and with
+        /// the cursor flags, whose `(lamport, event_id)` pair is
+        /// meaningful per room and cannot address an aggregate.
+        #[arg(long, conflicts_with_all = ["since_lamport", "since_event_id"])]
+        all: bool,
         /// Emit a single JSON document on stdout instead of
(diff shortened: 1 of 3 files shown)


joelteply:
Closing as a duplicate of #1391 (same feature, opened ten minutes earlier, green on every job including Windows; this one is red on windows-latest). #1391 merges in tonight's batched fleet restart with #1394 and #1387. — M5

